PR #43, opened 2026-02-15
Implemented routes for asynq and initial ui

rmorlok:


sdks/js/src/taskMonitoring.ts
```diff
@@ -0,0 +1,267 @@
...
+  days?: number;
+}
+
+export type MonitoringTaskState = 'pending' | 'active' | 'scheduled' | 'retry' | 'archived' | 'completed';
+
+/**
+ * List all queues with their info
+ */
+export const listQueues = () => {
+  return client.get<QueueInfo[]>('/api/v1/task-monitoring/queues');
```
rmorlok: - [ ] I don't think this is correct under the new api contract

Issue #159, opened 2026-04-27
OAuth Integration Test Requirements


rmorlok:
## Sub-issues

Tracked under label [`oauth-test-suite`](https://github.com/rmorlok/authproxy/labels/oauth-test-suite).

### Phase 0 — Foundation (blocks everything else)

- [ ] #162 — Stand up go-oauth2-server test infrastructure
- [ ] #163 — Gap analysis & required-changes doc for go-oauth2-server

### Phase 1 — P0 scenarios

- [ ] #164 — Standard successful authorization-code flow (spec #1)
- [ ] #165 — User denial flow (spec #2)
- [ ] #166 — Scope mismatch handling (spec #3)
- [ ] #167 — Callback state security: CSRF, replay, cross-tenant (spec #4)
- [ ] #168 — Authorization-code exchange failures (spec #5)
- [ ] #169 — Access token expiry, refresh, and no-refresh-token flow (spec #6, #13)
- [ ] #170 — Refresh token failure modes + transient retry policy (spec #7, #8)
- [ ] #171 — Refresh token rotation + concurrent refresh safety (spec #9, #10)
- [ ] #172 — Third-party revocation detection (spec #11)
- [ ] #173 — Sensitive values are not logged (spec #12)

### Phase 2 — P1 scenarios

- [ ] #174 — PKCE validation (spec #14)
- [ ] #175 — Malformed callbacks: code+error, missing code (spec #15, #16)
- [ ] #176 — Malformed token responses (spec #17)
- [ ] #177 — Provider timeouts (spec #18)
- [ ] #178 — Provider auth method compatibility (spec #19)
- [ ] #179 — Proxy API call response handling: 200/401/403/429/5xx (spec #20–#24)

### Phase 3 — P2 scenarios

- [ ] #180 — Incremental authorization (spec #25)
- [ ] #181 — Proxy-initiated disconnect / revocation (spec #26)
- [ ] #182 — Provider account identity changes (spec #27)
- [ ] #183 — Multiple connections to same provider (spec #28)
- [ ] #184 — Open redirect protection (spec #29)
- [ ] #185 — Clock skew (spec #30)

---

# OAuth Integration Test Requirements

Use `rmorlok/go-oauth2-server` as a controllable fake third-party OAuth provider for integration tests.

[...]

PR #193, opened 2026-05-01
Add OAuth2 standard authorization-code flow integration test

### LINKED ISSUE #164: OAuth tests P0: standard successful authorization-code flow
Parent: #159
Phase: 1 (P0)
Spec: #159 scenario 1
Blocked by: #162

## Verify

- Proxy creates an authorization URL with required query params:
  - \`client_id\`, \`redirect_uri\`, \`response_type=code\`, \`scope\`, \`state\`
  - \`code_challenge\` and \`code_challenge_method\` if PKCE is enabled
- User approves requested permissions on the fake provider.
- Provider redirects back with \`code\` and \`state\`.
- Proxy validates \`state\`.
- Proxy exchanges authorization code for tokens.
- Proxy stores: access token, refresh token (if returned), expiry, granted scopes, provider account identity (if available).
- Subsequent proxied API call succeeds using the stored token.

## Observability

- \`oauth flow started\`, \`oauth flow succeeded\` events emitted with connection/tenant/provider/correlation IDs.

rmorlok:
## Summary

- Adds `integration_tests/oauth2/standard_flow_test.go` covering spec #1 from #159: a successful authorization-code flow end-to-end (authorize-URL params → user approve → callback → token exchange → ready connection → proxied API call with Bearer).
- Extends `helpers.Setup` with an `IncludePublic` option so a single `DependencyManager` boots both the API and public services in-process — needed because `/oauth2/redirect` and `/oauth2/callback` only live on the public service.
- Adds in-process OAuth flow helpers on `IntegrationTestEnv`: `PublicCallbackURL`, `InitiateOAuth2Connection`, `FollowOAuth2Redirect`, `DeliverOAuth2Callback`, `GetOAuth2Token`, `GetConnection`.

Closes #164.

## Test plan

- [x] `go test -tags integration ./oauth2/...` green locally (postgres + oauth-server via docker compose)
- [x] Full integration suite passes
- [x] `./scripts/preflight.sh` passes

### DIFF
--- a/integration_tests/helpers/setup.go
+++ b/integration_tests/helpers/setup.go
@@ -74,9 +74,14 @@ type IntegrationTestEnv struct {
 	// Logger is the test logger.
 	Logger *slog.Logger
 
-	// ServerURL is the base URL of the real HTTP server (e.g. "http://127.0.0.1:54321").
-	// Only set when StartHTTPServer is true.
+	// ServerURL is the base URL of the real HTTP server hosting the primary
+	// service (api or admin-api per SetupOptions.Service). Only set when
+	// StartHTTPServer is true.
 	ServerURL string
+
+	// PublicURL is the base URL of the real HTTP server hosting the public
+	// service. Only set when StartHTTPServer && IncludePublic.
+	PublicURL string
 
 	// BearerToken is a pre-signed admin JWT for making HTTP requests to ServerURL.
 	BearerToken string
@@ -105,11 +110,20 @@ type SetupOptions struct {
 	// When false, use ApiGin with httptest.ResponseRecorder for in-process testing.
 	StartHTTPServer bool
 
-	// IncludePublic also wires up the public service's gin engine so tests can
-	// drive `/oauth2/redirect` and `/oauth2/callback`. The public engine shares
-	// the same DependencyManager (DB, Redis, core, encryption) as the primary
-	// service. Only meaningful for in-process testing (StartHTTPServer=false).
+	// IncludePublic also wires up the public service alongside the primary
+	// service so tests can drive `/oauth2/redirect`, `/oauth2/callback`, and
+	// (when ServeMarketplaceUI is true) the marketplace SPA. The public service
+	// shares the same DependencyManager (DB, Redis, core, encryption) as the
+	// primary service. With StartHTTPServer=true, public is started on its
+	// own listener and exposed via env.PublicURL; otherwise it's served
+	// in-process via env.PublicGin.
 	IncludePublic bool
+
+	// ServeMarketplaceUI configures the public service to serve the built
+	// marketplace SPA at "/" (in addition to its API routes). Implies
+	// IncludePublic. The vite build runs once per `go test` process if
+	// `ui/marketplace/dist/index.html` is missing.
+	ServeMarketplaceUI bool
 }
 
 // repoRoot returns the absolute path to the repository root.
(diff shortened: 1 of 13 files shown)


integration_tests/helpers/setup.go
```diff
@@ -45,12 +47,21 @@ type IntegrationTestEnv struct {
 	// Gin is the HTTP router for in-process request testing (when StartHTTPServer is false).
 	Gin *gin.Engine
 
+	// PublicGin is the public service router. Only populated when SetupOptions.IncludePublic
+	// is true. Tests that drive an OAuth flow need this to deliver `/oauth2/redirect` and
+	// `/oauth2/callback` requests, which only the public service hosts.
+	PublicGin *gin.Engine
```
rmorlok, line 55: - [ ] You've started to use the `Public` prefix here for the public endpoints. If we are going to do that, use the `Api` prefix on those referring to the API service, such as with `Gin` above (becomes `ApiGin`).

integration_tests/helpers/setup.go
```diff
@@ -45,12 +47,21 @@ type IntegrationTestEnv struct {
...
+
 	// Cfg is the loaded configuration.
 	Cfg config.C
 
 	// AuthUtil provides JWT signing helpers for test requests.
 	AuthUtil *auth2.AuthTestUtil
 
+	// PublicAuthUtil signs requests for the public service (audience=public). Only
+	// populated when SetupOptions.IncludePublic is true.
+	PublicAuthUtil *auth2.AuthTestUtil
```
rmorlok, line 66: Same note as above regarding consistent prefixing.

integration_tests/helpers/setup.go
```diff
@@ -372,3 +407,128 @@ func (env *IntegrationTestEnv) CreateConnection(t *testing.T, connectorID apid.I
 
 	return id.String()
 }
+
+// PublicCallbackURL returns the URL the proxy emits as the OAuth `redirect_uri`.
+// The OAuth provider must be configured with this exact URL so authorize matches
+// (helper exists because public.GetBaseUrl() depends on resolved config — port 0
+// in integration.yaml means the URL is "http://localhost:0/oauth2/callback").
+func (env *IntegrationTestEnv) PublicCallbackURL() string {
```
rmorlok: Let's call this `PublicOAuthCallbackURL`

integration_tests/oauth2/standard_flow_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	"github.com/rmorlok/authproxy/integration_tests/helpers"
+	"github.com/rmorlok/authproxy/internal/apid"
+	"github.com/rmorlok/authproxy/internal/database"
+	sconfig "github.com/rmorlok/authproxy/internal/schema/config"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestStandardAuthorizationCodeFlow exercises spec #1 from issue #159:
```
rmorlok: Don't cit the ticket; instead, create a standard_flow_test.md companion file to this test file that contains the definition of the scenario from the spec. Please be sure to include mermaid diagrams, especially sequence ones, that shows the flow being executed. Cite which components are being run in process, in docker, etc.

rmorlok (commented):
Please look at these issues. I'd also like you to use a browser-driven approach for interactions where the user is going to be taken through a redirect flow. So something like selenium that will actually load the UI both from the Authproxy side (marketplace) and on the go-oauth2-server side. For tests where the primary focus is about token refresh, etc that isn't required, but when the goal is to make sure that the OAuth flow works, I want to make sure it works as the user would see it.

integration_tests/helpers/setup.go
rmorlok: Renamed to `ApiGin` in 8ecba35.

integration_tests/helpers/setup.go
rmorlok: Renamed to `ApiAuthUtil` in 8ecba35.

integration_tests/helpers/setup.go
rmorlok: Renamed to `PublicOAuthCallbackURL` in 8ecba35.

integration_tests/oauth2/standard_flow_test.go
rmorlok: Done in 8ecba35 — added `integration_tests/oauth2/standard_flow_test.md` with a mermaid sequence diagram and an in-process / docker component breakdown, and dropped the ticket citation from the test docstring.

Issue #202, opened 2026-05-01
Carry-forward labels — Step 2c: triggers (upgrade, namespace move, parent label propagation)


rmorlok:
Sub-task of #194. Was part of #196; split out for reviewability. Depends on the sub-issues covering 2a and 2b.

## Goals
Keep materialized \`apxy/\` labels in sync when the parent resource changes after a child has already been created. This covers four mutation paths:

1. **Connection upgrade to a new connector version** — when a connection's pointed-at connector version changes, recompute the connection's \`apxy/cxr/*\` labels from the new version. Preserve user labels and \`apxy/cxn/*\` self labels.
2. **Namespace move on any resource** — when a connection / actor / encryption-key / connector-version is moved to a new namespace, update its \`apxy/<self>/-/ns\` and recompute its \`apxy/ns/*\` from the new namespace.
3. **Namespace user-label change** — when a namespace's user labels change (Update / Put / Delete), recompute \`apxy/ns/*\` on:
   - all child resources within that namespace, AND
   - all descendant namespaces (recursively, since they materialized the chain).
4. **Connector version user-label change** — when a connector version's user labels change, recompute \`apxy/cxr/*\` on every connection currently pointing at that connector version.

(Note: \`ConnectorVersion\` is documented as \"immutable\" in the existing data model in the sense that a new version is created for substantive changes. However, label-only updates on a head version may exist; if not, this trigger is a no-op and can be omitted with a clear comment.)

## Implementation notes
- Add helpers like \`refreshConnectionApxyLabels(ctx, tx, connId)\` and \`refreshNamespaceApxyChain(ctx, tx, nsPath)\` that re-derive the \`apxy/\` portion and write it back via \`replaceUserLabelsInTableTx\`-equivalent (preserving the user portion).
- Namespace label change propagation will be the largest pass — likely paginated to avoid huge transactions. Consider running it asynchronously via the existing asynq queue if more than N descendant rows; otherwise inline transaction. Document the chosen threshold.
- All triggers must hold the row-level invariants from 2a (no key stomping across user/system portions).

## Out of scope
- Daily consistency checker (#198) — that catches drift if any of these triggers misfire.
- Backfill migration for resources created before 2a / 2b landed — separate concern; can be done in #198's PR or as a one-shot script.

## Acceptance criteria
- [ ] Connection-upgrade test: create connection v1, change its connector_version_id (or version pointer), verify \`apxy/cxr/*\` reflects the new version's user labels and the user labels are preserved.
[...]

Issue #201, opened 2026-05-01
Carry-forward labels — Step 2b: parent carry-forward on resource create


rmorlok:
Sub-task of #194. Was part of #196; split out for reviewability. Depends on the sub-issue covering 2a (endpoint preservation + self-implicit on create).

## Goals
On resource create, materialize the *parent's* carry-forward labels into the new resource's \`labels\` column, in addition to the self-implicit labels from 2a.

## Carry-forward graph (per #194 + nested-namespace clarification on #196)
- **Namespace**: parent = parent namespace (walked via \`SplitNamespacePathToPrefixes\`). The immediate parent's labels are loaded; their user portion is re-keyed as \`apxy/ns/<k>\`, their \`apxy/\` portion is forwarded as-is. Because each namespace already stored its parent's chain at create time, this naturally yields the full ancestor chain with deeper-overrides-shallower.
- **ConnectorVersion**: parent = its namespace. Same rule as above.
- **Connection**: parents = its connector version + its namespace. Take connector version's full \`labels\` → re-key user portion as \`apxy/cxr/<k>\`, forward \`apxy/\` portion as-is. Take namespace's labels → re-key user portion as \`apxy/ns/<k>\`, forward \`apxy/\` portion as-is. Self-implicit \`apxy/cxn/-/id\`, \`apxy/cxn/-/ns\` overrides parent pass-throughs on key collisions.
- **Actor**: parent = its namespace. Same rule.
- **EncryptionKey**: parent = its namespace. Same rule.

## Implementation notes
- Add a service-level helper \`computeParentApxyLabelsForX(ctx, tx, parents...) (Labels, error)\` per resource type, or one generic helper that takes a list of (parentRt, parentLabels) pairs. The latter is cleaner — the resource-specific create path just loads its parent rows and feeds them in.
- Hook into create paths inside the transaction so the parent read and the child insert are atomic.
- Existing helper \`BuildCarriedLabels(parentRt, parentLabels)\` (#195) does the re-keying; just merge the result into the child's \`labels\` before insert.
- Order of merge for child's stored labels: \`parentCarriedNamespace ∪ parentCarriedConnectorVersion ∪ selfImplicit ∪ userLabels\`, with self-implicit and user labels winning over pass-throughs of the same key.

## Acceptance criteria
[...]

Issue #198, opened 2026-05-01
Carry-forward labels — Step 4: daily consistency checker


rmorlok:
Sub-task of #194. Depends on #196.

## Goals
- Daily background task that walks every labelled resource and reconciles drift in the materialized `apxy/` labels.
- Defense-in-depth against bugs or skipped triggers in #196.

## Scope
- New task in `internal/database` that, for each labelled resource type (`ConnectorVersion`, `Connection`, `Namespace`, `Actor`, `EncryptionKey`):
  1. Pages through rows in batches.
  2. Recomputes expected `apxy/` labels from the resource's current `id`, `namespace`, and parent rows.
  3. If actual ≠ expected, writes the corrected set and emits a structured log entry.
- Rate-limit so the checker doesn't hammer the database — implementation can use a simple per-batch sleep/throttle. Default cadence: daily.
- Surface metrics: rows scanned, drift count by resource type and by label key class (`-/id`, `-/ns`, carried user label).

## Out of scope
- Backfilling labels for resources created before #196 lands — handled by a one-shot migration in #196's PR (or a follow-up PR).
- Repairing user label drift — only `apxy/`-prefixed keys are reconciled; user labels are authoritative.

## Acceptance criteria
- [ ] Task registered with a daily schedule.
- [ ] Rate-limiting in place (configurable batch size and inter-batch delay).
- [ ] Test: seed a row with stale `apxy/<self>/-/ns`, run one pass, confirm it is corrected.
- [ ] Test: seed a connection whose connector's user labels have diverged from its `apxy/cxr/<k>` materialization, confirm it is corrected.
- [ ] Logs / metrics emitted as specified.

Issue #196, opened 2026-05-01
Carry-forward labels — Step 2: materialize carry-forward on resources


rmorlok:
Sub-task of #194. Depends on #195.

## Goals
- Materialize implicit `-/id` and `-/ns` labels and parent carry-forward labels on each resource's `labels` column.
- Update `GetLabels()` to return the merged set without breaking existing user-label semantics in storage.
- Ensure user-supplied label writes never stomp on `apxy/` keys; internal writes never stomp on user keys.

## Affected resources (have `labels` columns)
`ConnectorVersion`, `Connection`, `Namespace`, `Actor`, `EncryptionKey`. `OAuth2Token` is intentionally excluded (no labels column, do not add one).

## Carry-forward graph
- **Self-implicit** on every labelled resource: `apxy/<rt>/-/id`, `apxy/<rt>/-/ns`.
- **ConnectorVersion**: parent is `Namespace` → `apxy/ns/<k>` for each user label on the namespace + namespace's `-/id`/`-/ns` re-keyed under `apxy/ns/...`.
- **Connection**: parents are `ConnectorVersion` + `Namespace` → `apxy/cxr/<k>` and `apxy/cxr/-/...` from connector version (which already includes `apxy/ns/...` due to its own carry-forward) + own `apxy/ns/...` from connection's namespace.
- **Actor**: parent is `Namespace` → `apxy/ns/...`.
- **EncryptionKey**: parent is `Namespace` → `apxy/ns/...`.

## Materialization triggers
- Resource create — compute parent + self labels, persist alongside user labels.
- Connection upgrade to a new connector version — recompute `apxy/cxr/*`, replacing prior entries; keep user labels and own `apxy/cxn/...`.
- Namespace move on any resource — update own `apxy/<self>/-/ns`, recompute `apxy/ns/...` from the new namespace, recompute downstream resources accordingly.
- Namespace user-label change — recompute `apxy/ns/...` on every resource within (and below — see open question below) that namespace.
- Connector-version user-label change (only the head version a connection points at) — recompute `apxy/cxr/...` on every connection of that connector. Note: connector versions are immutable, so the only change vector is *which* version a connection points at.

## API guard
- `PutXLabels` / `UpdateXLabels` / `DeleteXLabels` mutate only the user-key portion. They preserve all `apxy/`-prefixed keys.
- User-input validation (from #195) rejects any `apxy/`-prefixed key in the request payload.

## `GetLabels()` semantics
[...]

PR #203, opened 2026-05-01
Carry-forward labels — Step 2a: self-implicit labels on create + endpoint preservation (#200)

### LINKED ISSUE #200: Carry-forward labels — Step 2a: endpoint preservation + self-implicit labels on create
Sub-task of #194. Was part of #196; split out for reviewability.

## Goals
1. **Defend \`apxy/\` keys at user-input boundaries.** \`UpdateXLabels\` / \`PutXLabels\` / \`DeleteXLabels\` must never insert, modify, or remove \`apxy/\`-prefixed keys. The \`apxy/\` portion of the stored set is always preserved verbatim across user-driven label writes.
2. **Materialize self-implicit labels on every resource create.** Each labelled resource (\`ConnectorVersion\`, \`Connection\`, \`Namespace\`, \`Actor\`, \`EncryptionKey\`) gets \`apxy/<rt>/-/id\` and \`apxy/<rt>/-/ns\` written into its \`labels\` column at create time, where \`<rt>\` is the resource's apid prefix without the trailing underscore.
3. **Relax label-value max length for \`apxy/.../-/ns\` system keys.** Namespace paths can exceed 63 chars; the user-facing \`ValidateLabelValue\` cap stays at 63, but the system-mode validator (used for stored \`apxy/\` values) accepts longer values up to a larger system-only cap (proposed: 253, matching the prefix cap).

## Out of scope
- Parent carry-forward on create (sub-issue 2b — to be filed).
- Triggers: connection upgrade, namespace move, parent label-change propagation (sub-issue 2c — to be filed).

## Implementation notes
- Add helpers in \`labels.go\`: \`SplitUserAndApxyLabels(Labels) (user, apxy Labels)\` and \`MergeApxyAndUserLabels(user, apxy Labels) Labels\`. Apxy and user keys are disjoint by prefix so merge never collides.
- Add a tx helper \`replaceUserLabelsInTableTx\` (or inline equivalent) that reads the current row's labels, extracts the \`apxy/\` portion, and writes back \`apxy/\` ∪ new-user-labels. Used by \`UpdateXLabels\`.
- For \`DeleteXLabels\`, key validation already rejects \`apxy/\` (#195). The actual deletion in \`deleteLabelsInTableTx\` then naturally only touches user keys.
- For create paths (\`CreateConnection\`, \`createNamespace\`/\`CreateNamespace\`, \`CreateConnectorVersion\`, \`CreateActor\`, \`CreateEncryptionKey\`): before insert, call \`BuildImplicitIdentifierLabels(c.Id, c.Namespace)\` and merge into \`c.Labels\` (or equivalent path field for namespaces).
- \`GetLabels()\` keeps returning the stored map verbatim (already includes any \`apxy/\` keys after this change).
- For the value-length relaxation: introduce a parallel \`validateApxyLabelValue\` (or a system-mode flag on \`ValidateLabelValue\`) — applied only when validating values whose key starts with \`apxy/\` AND the call site is the system-mode validator (\`ValidateLabels\`, \`Labels.Validate()\`). User-input validators continue to use the 63-char cap regardless.

## Acceptance criteria
- [ ] \`UpdateConnectionLabels\` / \`PutConnectionLabels\` / \`DeleteConnectionLabels\` test: row pre-populated with \`{apxy/cxr/-/id, user_key}\`, user calls \`Update\` with \`{user_key2}\` — final state contains both \`apxy/cxr/-/id\` and \`user_key2\`, and no \`user_key\`.
- [ ] Same test for the actor / namespace / encryption-key equivalents.
- [ ] Create test for each resource type: after \`CreateX\`, the row's \`labels\` contains \`apxy/<rt>/-/id == <id>\` and \`apxy/<rt>/-/ns == <ns_path>\`.
- [ ] Validation test: \`ValidateLabels\` accepts \`apxy/cxn/-/ns: root.foo.bar.baz.qux.quux.corge.grault.garply.waldo.fred.plugh\` (>63 chars) when the value is keyed under \`apxy/\`, but \`ValidateLabelValue\` (user mode) still rejects the same string.
- [ ] \`GetLabels()\` on a resource returns the merged set including any \`apxy/\` keys.
- [ ] \`./scripts/preflight.sh\` passes; \`go test ./...\` passes.

rmorlok:
Closes #200. First slice of #196's split (sub-issues #200 / #201 / #202).

## Summary
After this change, every newly-created labelled resource persists \`apxy/<rt>/-/id\` and \`apxy/<rt>/-/ns\` into its \`labels\` column at create time, and user-driven label writes can no longer overwrite or drop those system-managed keys.

### Scope
- **Self-implicit on create.** \`CreateConnection\`, \`CreateNamespace\`, \`CreateEncryptionKey\`, \`CreateActor\` (insert), \`UpsertActor\` (insert and update), and \`UpsertConnectorVersion\` (insert and update) now inject \`apxy/<rt>/-/id\` and \`apxy/<rt>/-/ns\` before write. Update paths preserve any existing \`apxy/\` entries on the row.
- **Endpoint preservation.** \`UpdateConnectionLabels\` / \`UpdateNamespaceLabels\` / \`UpdateEncryptionKeyLabels\` now use a new \`replaceUserLabelsInTableTx\` that rewrites only the user portion. \`Put\`/\`Delete\` already touch only their supplied keys (after #195's user-input validators reject \`apxy/\` keys).
- **Value-length relaxation for \`apxy/.../-/ns\`.** Added \`ApxyLabelValueMaxLength = 253\` so namespace paths fit. \`ValidateLabels\` / \`Labels.Validate\` dispatch on key prefix; user-input \`ValidateLabelValue\` still enforces the standard 63-char cap.
- **\`Actor.sameAsData\` user-only comparison.** Compares only the user portion of stored labels, so the presence of system-managed \`apxy/\` entries doesn't trigger spurious upsert updates from sources (JWT claims, sync_actors) that don't carry them.

### New helpers
- \`ApxyLabelValueMaxLength\`, \`apxyLabelValueRegex\`, \`ValidateApxyLabelValue\`.
- \`SplitUserAndApxyLabels\`, \`MergeApxyAndUserLabels\`.
- \`InjectSelfImplicitLabels\` (apid-keyed resources).
- \`NamespaceLabelToken\`, \`BuildImplicitIdentifierLabelsForToken\`, \`BuildNamespaceImplicitIdentifierLabels\`, \`InjectNamespaceSelfImplicitLabels\` (path-keyed namespace resources).
- \`replaceUserLabelsInTableTx\` (tx-level; preserves \`apxy/\` portion).

### Tests
- New unit tests covering the helpers, value-length relaxation, and self-implicit injection.
- Existing tests updated to split user vs apxy portions when asserting label content (otherwise the new system keys break assertions).

## Out of scope (will land in #201 / #202)
- Parent carry-forward on resource create (#201).
- Triggers: connection upgrade, namespace move, parent label-change propagation (#202).

## Notes for review
- Picked **option 1** for the open namespace-path-length question — relax the value cap for system-managed \`apxy/.../-/ns\` keys (253 chars), keep user-input cap at 63. Locked in via \`apxyLabelValueRegex\` and the prefix-based dispatch in \`ValidateLabels\`.
- Backfill for resources created before this PR is intentionally not done here — the daily consistency checker (#198) will handle drift across the fleet, including any pre-existing rows.

## Test plan
- [x] \`go test ./internal/database/\` — new unit tests pass; updated integration tests pass.
- [x] \`go test ./...\` — full repo passes.
- [x] \`./scripts/preflight.sh\` — passes.

### DIFF
--- a/internal/database/connector_version.go
+++ b/internal/database/connector_version.go
@@ -379,9 +379,25 @@ func (s *service) UpsertConnectorVersion(ctx context.Context, cv *ConnectorVersi
 				return errors.New("cannot modify non-draft connector")
 			}
 
+			// Preserve existing apxy/ system labels — the caller passes only
+			// the user-portion. Re-inject self-implicit labels in case the
+			// stored row predates Step 2a or namespace was somehow changed.
+			var existingLabels Labels
+			if scanErr := sqb.
+				Select("labels").
+				From(ConnectorVersionsTable).
+				Where(sq.Eq{"id": cv.Id, "version": cv.Version, "deleted_at": nil}).
+				QueryRowContext(ctx).
+				Scan(&existingLabels); scanErr != nil {
+				return scanErr
+			}
+			_, existingApxy := SplitUserAndApxyLabels(existingLabels)
+			mergedLabels := MergeApxyAndUserLabels(cv.Labels, existingApxy)
+			mergedLabels = InjectSelfImplicitLabels(cv.Id, cv.Namespace, mergedLabels)
+
 			result, err := sqb.Update(ConnectorVersionsTable).
 				Set("state", cv.State).
-				Set("labels", cv.Labels).
+				Set("labels", mergedLabels).
 				Set("annotations", cv.Annotations).
 				Set("encrypted_definition", cv.EncryptedDefinition).
 				Set("updated_at", apctx.GetClock(ctx).Now()).
@@ -420,6 +436,7 @@ func (s *service) UpsertConnectorVersion(ctx context.Context, cv *ConnectorVersi
 			}
 
 			cpy := *cv
+			cpy.Labels = InjectSelfImplicitLabels(cpy.Id, cpy.Namespace, cpy.Labels)
 			now := apctx.GetClock(ctx).Now()
 			cpy.CreatedAt = now
 			cpy.UpdatedAt = now
--- a/internal/database/connector_version.go
+++ b/internal/database/connector_version.go
@@ -379,9 +379,8 @@ func (s *service) UpsertConnectorVersion(ctx context.Context, cv *ConnectorVersi
 				return errors.New("cannot modify non-draft connector")
 			}
 
-			// Preserve existing apxy/ system labels — the caller passes only
-			// the user-portion. Re-inject self-implicit labels in case the
-			// stored row predates Step 2a or namespace was somehow changed.
+			// Preserve apxy/ system labels — the caller passes only the
+			// user portion.
 			var existingLabels Labels
 			if scanErr := sqb.
 				Select("labels").
@@ -393,7 +392,6 @@ func (s *service) UpsertConnectorVersion(ctx context.Context, cv *ConnectorVersi
 			}
 			_, existingApxy := SplitUserAndApxyLabels(existingLabels)
 			mergedLabels := MergeApxyAndUserLabels(cv.Labels, existingApxy)
-			mergedLabels = InjectSelfImplicitLabels(cv.Id, cv.Namespace, mergedLabels)
 
 			result, err := sqb.Update(ConnectorVersionsTable).
 				Set("state", cv.State).
--- a/internal/database/actor.go
+++ b/internal/database/actor.go
@@ -188,8 +188,7 @@ func (a *Actor) setFromData(d IActorData) {
 
 func (a *Actor) sameAsData(d IActorData) bool {
 	// Compare labels — only the user portion. apxy/ labels are system-managed
-	// and never appear in IActorData, so excluding them avoids spurious
-	// updates after Step 2 carry-forward materialization.
+	// and never appear in IActorData, so excluding them avoids spurious updates.
 	aUserLabels, _ := SplitUserAndApxyLabels(a.Labels)
 	dLabels := d.GetLabels()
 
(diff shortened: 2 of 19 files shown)


internal/database/connector_version.go
```diff
@@ -379,9 +379,25 @@ func (s *service) UpsertConnectorVersion(ctx context.Context, cv *ConnectorVersi
 				return errors.New("cannot modify non-draft connector")
 			}
 
+			// Preserve existing apxy/ system labels — the caller passes only
+			// the user-portion. Re-inject self-implicit labels in case the
+			// stored row predates Step 2a or namespace was somehow changed.
```
rmorlok: Step 2a refers to an implementation step. Don't put that information into the code. If you add temporary comments for WIP, cite a ticket number instead.

In this particular case, you don't need to worry about existing data. There is no live system yet so you can assume all data will be fresh.

rmorlok: Done in 09ba9d7:
- Dropped the "Step 2a" reference here and the "Step 2 carry-forward materialization" reference in `actor.go`'s `sameAsData` comment.
- Removed the self-implicit re-injection on this update path entirely. ConnectorVersion namespace is immutable (the guard at the top of `UpsertConnectorVersion`), so `apxy/cxr/-/ns` can't drift on update — only the `apxy/` portion needs to survive the user-portion replacement.
- Applied the same simplification to `UpsertActor`'s update path for consistency. If a namespace-changing upsert leaves `apxy/act/-/ns` momentarily stale, that's the namespace-move trigger's job (#202), and #198 catches anything that slips.

PR #205, opened 2026-05-01
Carry-forward labels — Step 2c: propagate label changes through carry-forward chain (#202)

### LINKED ISSUE #202: Carry-forward labels — Step 2c: triggers (upgrade, namespace move, parent label propagation)

rmorlok:
Closes #202. Final slice of #196's split.

## Summary
When a parent's user labels change, descendants need their materialized `apxy/` portion re-derived. To avoid blocking the originating API request — a deeply nested namespace can fan out to thousands of descendants — propagation runs off the request path via asynq background tasks. The synchronous DB write only updates the parent's own row; the task handler walks descendants in per-row short transactions.

### Database layer
- New file `internal/database/labels_propagate.go`:
  - `recompute<Resource>LabelsTx` helpers (one per resource type) — open a short transaction, re-run the create-time materialization against the current state of a row's parents, persist the result, commit. Used as the building block for every refresh.
  - `RefreshNamespaceLabelsCarryForward(ctx, nsPath)` — walks every Connection / Actor / EncryptionKey / ConnectorVersion in `nsPath`, then walks each direct child namespace, recomputes the child's own labels, and recurses. Each row's update is its own short transaction; no long-running tx is held.
  - `RefreshConnectionsForConnectorVersion(ctx, id, version)` — refreshes every connection pointing at a given (cv id, version).
- Both `Refresh*` methods are exposed on the `database.DB` interface so the asynq task handlers can drive them.

### Tasks layer
- `internal/database/tasks/task_propagate_namespace_labels.go` — task type `database:propagate_namespace_labels` with payload `{namespace_path}`. Handler delegates to `db.RefreshNamespaceLabelsCarryForward`. Malformed / empty payloads return `asynq.SkipRetry`.
- `internal/database/tasks/task_propagate_connector_version_labels.go` — task type `database:propagate_connector_version_labels` with payload `{connector_version_id, version}`. Handler delegates to `db.RefreshConnectionsForConnectorVersion`. Same malformed-payload behavior.

### Core layer (request path)
- `core.UpdateNamespaceLabels` / `PutNamespaceLabels` / `DeleteNamespaceLabels` enqueue `propagate_namespace_labels` after a successful DB write.
- `core.UpdateDraftConnectorVersion` (the only path that mutates a cv's user labels — primary and active are immutable) enqueues `propagate_connector_version_labels`.
- Failures to enqueue are logged but never fail the originating request; #198 (daily consistency checker) is the safety net if a task is dropped.

### Triggers explicitly NOT wired
- **Connection upgrade.** No API mutates `Connection.ConnectorVersion` post-create today. `recomputeConnectionLabelsTx` is in place so #198 can repair drift if such an API is added later.
- **Namespace move on resources.** Same — no API mutates a resource's `Namespace` column post-create. The per-resource recompute helpers cover this defensively.

## Tests
- DB-level: `TestNamespaceLabelChangePropagation` and `TestConnectorVersionLabelChangePropagation` exercise `Refresh*` directly so the propagation primitive itself stays covered.
- Task handlers: `TestPropagateNamespaceLabelsTask` and `TestPropagateConnectorVersionLabelsTask` use the DB mock to verify payload parsing, delegation, and `SkipRetry` on bad input.
- Core: `TestNamespaceLabelEndpointsEnqueuePropagation` asserts that Update / Put / Delete on namespace labels each enqueue exactly one `database:propagate_namespace_labels` task.
- Existing `TestUpdateDraftConnectorVersion` updated to expect the asynq enqueue. Route tests (`connectors_test.go`, `namespaces_test.go`) configure the asynq mock with `AnyTimes()` so they don't have to care about enqueue calls.

## Test plan
- [x] `go test ./internal/database/` and `./internal/database/tasks/` — propagation primitive + task handler tests pass.
- [x] `go test ./internal/core/` — enqueue test passes.
- [x] `go test ./...` — full repo passes; no cross-package regressions.
- [x] `./scripts/preflight.sh` — passes.

### DIFF
--- a/internal/database/namespace.go
+++ b/internal/database/namespace.go
@@ -744,10 +744,7 @@ func (s *service) UpdateNamespaceLabels(ctx context.Context, path string, labels
 		ns.Labels = merged
 		ns.UpdatedAt = now
 		result = ns
-
-		// Propagate the user-label change to every descendant resource and
-		// child namespace so the materialized apxy/ns/* portion stays in sync.
-		return s.propagateNamespaceLabelsChange(ctx, tx, path)
+		return nil
 	})
 
 	if err != nil {
@@ -788,8 +785,7 @@ func (s *service) PutNamespaceLabels(ctx context.Context, path string, labels ma
 		ns.Labels = mergedLabels
 		ns.UpdatedAt = now
 		result = ns
-
-		return s.propagateNamespaceLabelsChange(ctx, tx, path)
+		return nil
 	})
 
 	if err != nil {
@@ -830,8 +826,7 @@ func (s *service) DeleteNamespaceLabels(ctx context.Context, path string, keys [
 		ns.Labels = remainingLabels
 		ns.UpdatedAt = now
 		result = ns
-
-		return s.propagateNamespaceLabelsChange(ctx, tx, path)
+		return nil
 	})
 
 	if err != nil {
--- a/internal/database/namespace.go
+++ b/internal/database/namespace.go
@@ -744,7 +744,10 @@ func (s *service) UpdateNamespaceLabels(ctx context.Context, path string, labels
 		ns.Labels = merged
 		ns.UpdatedAt = now
 		result = ns
-		return nil
+
+		// Propagate the user-label change to every descendant resource and
+		// child namespace so the materialized apxy/ns/* portion stays in sync.
+		return s.propagateNamespaceLabelsChange(ctx, tx, path)
 	})
 
 	if err != nil {
@@ -785,7 +788,8 @@ func (s *service) PutNamespaceLabels(ctx context.Context, path string, labels ma
 		ns.Labels = mergedLabels
 		ns.UpdatedAt = now
 		result = ns
-		return nil
+
+		return s.propagateNamespaceLabelsChange(ctx, tx, path)
 	})
 
 	if err != nil {
@@ -826,7 +830,8 @@ func (s *service) DeleteNamespaceLabels(ctx context.Context, path string, keys [
 		ns.Labels = remainingLabels
 		ns.UpdatedAt = now
 		result = ns
-		return nil
+
+		return s.propagateNamespaceLabelsChange(ctx, tx, path)
 	})
 
 	if err != nil {
--- a/internal/core/service_connector_version_update.go
+++ b/internal/core/service_connector_version_update.go
@@ -8,9 +8,26 @@ import (
 	"github.com/rmorlok/authproxy/internal/apid"
 	"github.com/rmorlok/authproxy/internal/core/iface"
 	"github.com/rmorlok/authproxy/internal/database"
+	dbtasks "github.com/rmorlok/authproxy/internal/database/tasks"
 	cschema "github.com/rmorlok/authproxy/internal/schema/connectors"
 	"github.com/rmorlok/authproxy/internal/util"
 )
+
+// enqueueConnectorVersionLabelPropagation enqueues an asynq task that
+// refreshes the materialized apxy/cxr/* portion of every connection
+// pointing at (id, version). Failures to enqueue are logged but do not
+// fail the originating request — the daily consistency checker (#198)
+// covers any drift if the task is dropped.
+func (s *service) enqueueConnectorVersionLabelPropagation(ctx context.Context, id apid.ID, version uint64) {
+	task, err := dbtasks.NewPropagateConnectorVersionLabelsTask(id, version)
+	if err != nil {
+		s.logger.Error("failed to build connector version label propagation task", "id", id, "version", version, "error", err)
+		return
+	}
+	if _, err := s.ac.EnqueueContext(ctx, task); err != nil {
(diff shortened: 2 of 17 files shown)


internal/database/namespace.go
```diff
@@ -744,7 +744,10 @@ func (s *service) UpdateNamespaceLabels(ctx context.Context, path string, labels
 		ns.Labels = merged
 		ns.UpdatedAt = now
 		result = ns
-		return nil
+
+		// Propagate the user-label change to every descendant resource and
+		// child namespace so the materialized apxy/ns/* portion stays in sync.
+		return s.propagateNamespaceLabelsChange(ctx, tx, path)
```
rmorlok: I don't think it's a good idea to do this synchronously with this call, which would be in a single api call. There could be thousands of resources that would change, and the changes of the child may need to propagate the changes as well. Let's change directions here. Let's create a background task that will handle label changes from a resource and propagates it to appropriate children. Each of the individual child updates in the background task do not need to be within a transaction.

rmorlok: Done in c56e5b2. Propagation now runs off the request path:

- Removed the inline propagation calls from `db.UpdateNamespaceLabels` / `PutNamespaceLabels` / `DeleteNamespaceLabels` and from the `UpsertConnectorVersion` update path. Those endpoints just write the user's labels and return.
- Added two asynq task types under `internal/database/tasks/`:
  - `database:propagate_namespace_labels` — walks every resource and descendant namespace under the given path; each row's update runs in its own short transaction.
  - `database:propagate_connector_version_labels` — refreshes every connection pointing at a (cv id, version).
- Refactored `RefreshNamespaceLabelsCarryForward` / `RefreshConnectionsForConnectorVersion` to use per-row short transactions and exposed them on the `database.DB` interface so the task handlers can drive them.
- Wired core (`UpdateNamespaceLabels` / `PutNamespaceLabels` / `DeleteNamespaceLabels` and `UpdateDraftConnectorVersion`) to enqueue the appropriate task after a successful DB write. Enqueue failures are logged but never fail the request; #198 (daily checker) is the safety net if a task is dropped.
- Added unit tests for both task handlers + a core test asserting that each namespace label endpoint enqueues exactly one propagation task.

Issue #197, opened 2026-05-01
Carry-forward labels — Step 3: per-request label snapshot


rmorlok:
Sub-task of #194. Depends on #196.

## Goals
- At request time, snapshot the connection's effective labels into the per-request label set so `LogRecord.Labels` reflects a point-in-time view.
- Reject `apxy/`-prefixed keys submitted via `ProxyRequest.Labels`.

## Scope
- When building the per-request label set:
  1. Start from the connection's `labels` (already materialized in #196 — includes `apxy/cxr/...`, `apxy/ns/...`, `apxy/cxn/-/id`, `apxy/cxn/-/ns`).
  2. Merge in user-supplied `ProxyRequest.Labels` (validated against #195's user-input rules — `apxy/` rejected).
  3. Persist into `LogRecord.Labels` and any downstream consumers.
- `ProxyRequest.Validate()` calls the user-input validator from #195.

## Acceptance criteria
- [ ] Request labels written to `LogRecord` contain connection's full materialized label set + user-supplied keys.
- [ ] Per-request label submission with an `apxy/`-prefixed key returns a clear validation error and no record is persisted.
- [ ] Updating a connection's `apxy/...` labels after a request was logged does not retroactively change the historical record.
- [ ] Tests covering the merge order (user labels do NOT override `apxy/` keys, since they're rejected at validation; document the precedence explicitly).

Issue #195, opened 2026-05-01
Carry-forward labels — Step 1: grammar, validation, helpers


rmorlok:
Sub-task of #194. Foundation work — no user-visible behavior change beyond the new validation rule.

## Goals
- Establish the `apxy/` reserved label-key namespace.
- Allow multi-segment prefixes (e.g. `apxy/cxr/-`) only when the prefix begins with `apxy/`.
- Provide internal helpers used by Steps 2–4.

### Grammar change in `internal/database/labels.go`
- `ValidateLabelKey` accepts keys of the form `apxy/<seg>(/<seg>)*/<name>` where each `<seg>` matches the existing prefix-segment rule (DNS label) and `<name>` matches the existing name rule.
- For all other keys, the existing single-slash `[prefix/]name` grammar is unchanged.
- Length caps unchanged.

### Validation modes
- `ValidateLabels` (current) becomes the *user-input* mode — rejects any key starting with `apxy/`.
- New `ValidateLabelsSystem` (or equivalent) — allows `apxy/` keys; used by internal carry-forward writes.
- API endpoints that accept user-supplied labels (`PutXLabels`, `UpdateXLabels`, `DeleteXLabels` for connection / connector version / namespace / actor / encryption key, plus per-request `ProxyRequest.Labels`) call the user-input validator.
- Internal core code performing carry-forward writes calls the system validator.

### Helpers
- `apidPrefixToLabelToken(apid.Prefix) string` — strips trailing underscore, e.g. `cxr_` → `cxr`, `cxn_` → `cxn`.
- `BuildImplicitIdentifierLabels(id apid.ID, namespacePath string) Labels` — returns `{apxy/<rt>/-/id: <id>, apxy/<rt>/-/ns: <namespacePath>}`.
- `BuildCarriedLabels(parentRt string, parentLabels Labels) Labels` — for each key in `parentLabels`, emits `apxy/<parentRt>/<key>` if the key has no prefix, and forwards `apxy/<grandparent>/...` keys as-is.

## Out of scope
- Any actual carry-forward at create / upgrade / namespace move (Step 2).
- Per-request label snapshotting (Step 3).
- Daily consistency checker (Step 4).

## Acceptance criteria
- [ ] Grammar tests cover: valid `apxy/cxr/-/id`, valid `apxy/cxr/<userkey>`, valid `apxy/ns/<userkey>`, invalid `apxy/` (empty name), invalid double-slash, invalid trailing slash, invalid segment chars.
[...]

Issue #194, opened 2026-05-01
Carry-forward Labels


rmorlok:
I want to extend the concept of labels that we have today for resources to take on labels from the element that defined them.

For example, suppose connector con_123 has the following labels as defined by the user in the config or api:

```json
{
    "id": "my_id"
	"type": "google_drive",
	"variant": "shared"
}
```

I would like connection cxn_321 to implicitly have the following labels:

```json
{
    "apxy/con/-/id": "con_123",
    "apxy/con/id": "my_id",
	"apxy/con/type": "google_drive",
	"apxy/con/variant": "shared",
	"subscription_level": "pro"
}
```

Note that subscription_level above is a user defined label applied to that connection (not inherited data from the connector). Note that the authproxy/connector/-/id is the actual id from the connector, where as authproxy/connector/id is the id user defined label on the connector. the -/ prefix indicates implicit labels from other identifiers.

When a request is made for that connection, the request would have the following labels:

```json
{
	"apxy/con/-/id": "con_123",
    "apxy/con/id": "my_id",
	"apxy/con/type": "google_drive",
	"apxy/con/variant": "shared",
	"apxy/cxn/-/id": "cxn_321",
	"apxy/cxn/subscription_level": "pro",
	"purpose": "some_feature"
}
```

Note here that purpose is a per-request label, apxy/cxn/-/id has been applied implicitly from the connection id, and e.g. apxy/con/type has been carried forward from the connection, which was ultimately defined for the connector.

apxy will be a restricted prefix which user defined labels are not allowed to start with. Note also that the con / cxn comes from the id prefix for those resource types -- that is how the value should be determined.

In terms of implicitly defined labels, I would like to have -/id from the resource id and -/ns for the namespace of the resource (not shown).

rmorlok:
Broken into 4 sub-tasks per discussion:

- #195 — Step 1: grammar, validation, helpers (foundation)
- #196 — Step 2: materialize carry-forward on resources
- #197 — Step 3: per-request label snapshot
- #198 — Step 4: daily consistency checker

Sequencing: #195 → #196 → (#197 ‖ #198).

Open question still to resolve in #196's thread: nested-namespace label inheritance — only the immediate namespace, or walk ancestors with deeper-overrides-shallower.

rmorlok:
Step 2 (#196) split into three smaller sub-issues for reviewability:

- #200 — Step 2a
- #201 — Step 2b
- #202 — Step 2c

[...]

PR #208, opened 2026-05-02
Carry-forward labels — Step 4: daily consistency checker (#198)

### LINKED ISSUE #198: Carry-forward labels — Step 4: daily consistency checker

rmorlok:
Closes #198. Final piece of the carry-forward labels work in #194.

## Summary
Drift in the materialized \`apxy/\` portion of any labelled resource — from a missed propagation task, a deploy that wrote labels via a path that predates carry-forward materialization, or an unforeseen bug — is now caught and corrected by a daily asynq cron task. The reconciliation helpers also let the propagation tasks shipped earlier in #202 stop doing no-op writes when nothing actually changed.

### Database layer
- \`writeRecomputedLabels\` now compares the recomputed labels against the row's current labels and skips the UPDATE when they're equal. Returns \`(corrected bool, err error)\`.
- Each \`recompute<Resource>LabelsTx\` helper threads that bool back to its caller. All five resource types (Connection, Actor, EncryptionKey, ConnectorVersion, Namespace) are covered.
- Add \`labelsEqual\` helper.
- Add \`ReconcileCarryForwardLabels(ctx, batchSize, interBatchDelay) (int64, error)\` on \`database.DB\`. Walks every labelled resource type (namespaces top-down, then connector versions, connections, actors, encryption keys), pages by \`batchSize\` via the existing \`Enumerate\` builders, and recomputes each row in its own short transaction. Sleeps \`interBatchDelay\` between resource types to throttle DB load. Logs a structured entry per drift fix and a summary at the end.

### Tasks layer
- \`internal/database/tasks/task_reconcile_carry_forward_labels.go\`: task type \`database:reconcile_carry_forward_labels\` with default \`batchSize=100\` and \`interBatchDelay=0\`. Handler delegates to \`db.ReconcileCarryForwardLabels\`.
- Registered as a \`@daily\` periodic task alongside the existing purge and cleanup periodics.

## Tests
- \`TestReconcileCarryForwardLabels\` in the database package:
  - **Drift-correction**: seed a connection, corrupt its labels via raw SQL to bypass the carry-forward materialization, run the reconciler, verify the count is 1 and the connection's \`apxy/cxr/type\` is restored.
  - **Idempotence**: a clean run after the first stable pass reports 0 corrected.
- \`TestReconcileCarryForwardLabelsTask\` in the tasks package: handler delegates to db with the right defaults and propagates errors so asynq retries.
- \`TestNewReconcileCarryForwardLabelsTask\`: the constructor returns a task with the expected type.

## What this closes for #194
With this PR, all five sub-issues are merged: foundation primitives (#195), self-implicit + endpoint preservation (#200), parent carry-forward on create (#201), background propagation on parent label change (#202), per-request snapshot (#197), and the daily safety-net reconciler here (#198).

## Test plan
- [x] \`go test ./internal/database/\` and \`./internal/database/tasks/\` — new and existing tests pass.
- [x] \`go test ./...\` — full repo passes; no cross-package regressions.
- [x] \`./scripts/preflight.sh\` — passes.

### DIFF
--- a/internal/database/labels_propagate.go
+++ b/internal/database/labels_propagate.go
@@ -5,10 +5,12 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"time"
 
 	sq "github.com/Masterminds/squirrel"
 	"github.com/rmorlok/authproxy/internal/apctx"
 	"github.com/rmorlok/authproxy/internal/apid"
+	"github.com/rmorlok/authproxy/internal/util/pagination"
 	aschema "github.com/rmorlok/authproxy/internal/schema/auth"
 )
 
@@ -57,7 +59,7 @@ func (s *service) RefreshNamespaceLabelsCarryForward(ctx context.Context, nsPath
 		return err
 	}
 	for _, childPath := range childPaths {
-		if err := s.recomputeNamespaceLabelsTx(ctx, childPath); err != nil {
+		if _, err := s.recomputeNamespaceLabelsTx(ctx, childPath); err != nil {
 			return err
 		}
 		if err := s.RefreshNamespaceLabelsCarryForward(ctx, childPath); err != nil {
@@ -99,7 +101,7 @@ func (s *service) RefreshConnectionsForConnectorVersion(ctx context.Context, id 
 	}
 	rows.Close()
 	for _, connID := range ids {
-		if err := s.recomputeConnectionLabelsTx(ctx, connID); err != nil {
+		if _, err := s.recomputeConnectionLabelsTx(ctx, connID); err != nil {
 			return err
 		}
 	}
@@ -164,7 +166,7 @@ func (s *service) refreshConnectionsInNamespace(ctx context.Context, nsPath stri
 		return err
 	}
 	for _, id := range ids {
-		if err := s.recomputeConnectionLabelsTx(ctx, id); err != nil {
+		if _, err := s.recomputeConnectionLabelsTx(ctx, id); err != nil {
 			return err
 		}
 	}
@@ -177,7 +179,7 @@ func (s *service) refreshActorsInNamespace(ctx context.Context, nsPath string) e
 		return err
 	}
 	for _, id := range ids {
-		if err := s.recomputeActorLabelsTx(ctx, id); err != nil {
+		if _, err := s.recomputeActorLabelsTx(ctx, id); err != nil {
 			return err
 		}
 	}
@@ -190,7 +192,7 @@ func (s *service) refreshEncryptionKeysInNamespace(ctx context.Context, nsPath s
 		return err
 	}
 	for _, id := range ids {
-		if err := s.recomputeEncryptionKeyLabelsTx(ctx, id); err != nil {
+		if _, err := s.recomputeEncryptionKeyLabelsTx(ctx, id); err != nil {
 			return err
 		}
 	}
@@ -226,7 +228,7 @@ func (s *service) refreshConnectorVersionsInNamespace(ctx context.Context, nsPat
 	}
 	rows.Close()
 	for _, ref := range refs {
-		if err := s.recomputeConnectorVersionLabelsTx(ctx, ref.id, ref.version); err != nil {
+		if _, err := s.recomputeConnectorVersionLabelsTx(ctx, ref.id, ref.version); err != nil {
 			return err
 		}
 	}
@@ -235,9 +237,11 @@ func (s *service) refreshConnectorVersionsInNamespace(ctx context.Context, nsPat
 
 // recomputeConnectionLabelsTx opens a short transaction, re-derives the
 // connection's full labels from its current connector version + namespace +
-// own user labels, and persists the result.
-func (s *service) recomputeConnectionLabelsTx(ctx context.Context, id apid.ID) error {
-	return s.transaction(func(tx *sql.Tx) error {
+// own user labels, and persists the result if it differs. Returns true if
+// drift was detected and corrected.
+func (s *service) recomputeConnectionLabelsTx(ctx context.Context, id apid.ID) (bool, error) {
+	var corrected bool
+	err := s.transaction(func(tx *sql.Tx) error {
 		var conn Connection
 		err := s.sq.
 			Select(conn.cols()...).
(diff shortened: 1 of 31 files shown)


internal/database/tasks/task_reconcile_carry_forward_labels.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+const taskTypeReconcileCarryForwardLabels = "database:reconcile_carry_forward_labels"
+
+// Defaults for the reconciliation pass. Drift is rare under normal
+// operation, so a small per-resource page and a short pause between
+// resource types is enough throttling to keep DB load negligible.
+const (
+	reconcileBatchSize       int32         = 100
+	reconcileInterBatchDelay time.Duration = 0
```
rmorlok: Instead of expressing this as a delay between batches, let's express it as records / sec processed. Use some sort of rate limiter library to accomplish this. So the limiter would be checked before an individual row is processed. This limiter would need to be passed to ReconcileCarryForwardLabels. That method could possibly take the limiter as a optional parameter if there are other situations where it's not needed.

internal/database/labels_propagate.go
```diff
@@ -440,22 +463,197 @@ func (s *service) recomputeNamespaceLabelsTx(ctx context.Context, path string) e
...
+
+	s.logger.Info("carry-forward labels reconciliation complete", "total_corrected", totalCorrected)
+	return totalCorrected, nil
+}
+
+func (s *service) reconcileNamespaces(ctx context.Context, batchSize int32) (int64, error) {
+	var corrected int64
+	err := s.ListNamespacesBuilder().
+		Limit(batchSize).
+		Enumerate(ctx, func(pr pagination.PageResult[Namespace]) (pagination.KeepGoing, error) {
```
rmorlok: To accomplish my ask about rate limiting, I think we'll need to add an Enumerate variant that takes a rate limiter.

internal/database/tasks/task_reconcile_carry_forward_labels.go
rmorlok: Done in 1cb1853.

- New \`pagination.EnumerateThrottled[T]\` helper wraps any builder's \`Enumerate\` (passed as a method value) and calls \`limiter.Wait(ctx)\` before invoking the per-item callback. Nil limiter means unlimited.
- \`ReconcileCarryForwardLabels\` now takes \`*rate.Limiter\` instead of an \`interBatchDelay\`. The same limiter is shared across all resource-type walks so the records/sec budget is fleet-wide, not per type.
- Daily asynq task builds a default \`rate.NewLimiter(100, 100)\` so a fleet of millions tops out under the limit while a small fleet finishes in seconds.
- New \`golang.org/x/time/rate\` is now a direct dep.
- Tests cover nil-limiter passthrough, callback-error early stop, observable throttling, limiter-error propagation, and the task handler asserts via a gomock matcher that the limiter is constructed at the default rps.

internal/database/labels_propagate.go
rmorlok: Done in 1cb1853 — added `pagination.EnumerateThrottled` (see thread on `task_reconcile_carry_forward_labels.go`). All five per-resource reconcile walks now go through it.

internal/util/pagination/enumerate_throttled.go
```diff
@@ -0,0 +1,43 @@
...
+	"context"
+
+	"golang.org/x/time/rate"
+)
+
+// EnumerateFunc is the shape of an Enumerate method on a list builder. It
+// matches the existing pagination iteration contract (a callback that
+// receives one PageResult at a time) so any builder's Enumerate method
+// can be passed as a method value.
+type EnumerateFunc[T any] func(context.Context, func(PageResult[T]) (keepGoing KeepGoing, err error)) error
```
rmorlok: Let's split this type out into a separate file in this package. Then look at the internal/database where we already use similar definitions and start using this type.

rmorlok: Done in 3f6dc32.

- Moved \`EnumerateFunc[T]\` and a sibling \`EnumerateCallback[T]\` (the per-page callback shape) into \`internal/util/pagination/enumerate.go\`.
- Replaced the inline \`func(pagination.PageResult[T]) (keepGoing pagination.KeepGoing, err error)\` shape with \`pagination.EnumerateCallback[T]\` everywhere it appeared:
  - All six database list executors: \`actors\`, \`connections\`, \`connectors\`, \`connector_versions\`, \`encryption_keys\`, \`namespaces\` (interface declarations + implementations).
  - The iface-layer wrappers in \`internal/core/iface/\` and their core implementations.
  - \`internal/request_log\`'s \`ListRequestExecutor\`, the SQL implementation, and the hand-rolled mock.
- No behavior change — the named types are structurally identical to the inline form.

PR #210, opened 2026-05-03
Reconcile orphaned config-sourced connectors during migration

### LINKED ISSUE #15: Don't auto-destroy connectors that have connections
When you start up and define connectors via the mechanism that loads from a config file, don't auto-destroy connectors that are no longer in the config file if they have connections remaining. The connector should transition the most recent version to active from primary to avoid new connections, and it should log a message indicating that API action is required to remove connections and then remove the connector.

rmorlok:
Closes #15.

## Summary
- When the config-file connector migration runs, connectors previously loaded from config but no longer present are now reconciled instead of left in place.
- Orphans with no live connections are soft-deleted; orphans with live connections have their primary version demoted to `active` (so no new connections can be created) and a warning is logged telling the operator to remove the remaining connections via the API before the connector itself can be removed.
- Config-sourced connector versions are tagged with an `apxy/cxr/source=config` system label so the orphan-cleanup pass only touches config-managed connectors and leaves API-created ones alone.

## Implementation notes
- `internal/database/connection.go` — added `ListConnectionsBuilder.ForConnectorId(id)` so we can cheaply check whether a connector still has connections.
- `internal/database/connector_version.go` — added `DB.DeleteConnector(ctx, id)` (soft-deletes every version of a connector); the `UpsertConnectorVersion` update path now lets caller-supplied apxy-prefixed labels override stored apxy labels (existing callers don't pass apxy keys, so behavior is unchanged for them).
- `internal/database/interface.go` + regenerated `internal/database/mock/db.go` for `DeleteConnector`.
- `internal/core/service_migration.go` — `migrateConnector` now returns the resolved connector id and tags the version with `apxy/cxr/source=config`; `MigrateConnectors` runs `cleanupOrphanedConfigConnectors` after the upsert loop.

## Test plan
- [x] `go test ./internal/database/... ./internal/core/...` on SQLite
- [x] `go test ./internal/database/... ./internal/core/...` on PostgreSQL
- [x] Full `go test ./...` on SQLite and PostgreSQL
- [x] `./scripts/preflight.sh`
- [x] New tests in `internal/core/service_migration_test.go` covering: orphan with no connections is deleted; orphan with live connections is demoted to active; API-created connectors are not touched

### DIFF
--- a/internal/database/connector_version.go
+++ b/internal/database/connector_version.go
@@ -379,8 +379,10 @@ func (s *service) UpsertConnectorVersion(ctx context.Context, cv *ConnectorVersi
 				return errors.New("cannot modify non-draft connector")
 			}
 
-			// Preserve apxy/ system labels — the caller passes only the
-			// user portion.
+			// Preserve apxy/ system labels — the caller usually passes only
+			// the user portion, but if they include apxy/-prefixed entries
+			// (e.g. system-owned provenance markers), let those override the
+			// stored apxy values.
 			var existingLabels Labels
 			if scanErr := sqb.
 				Select("labels").
@@ -390,8 +392,16 @@ func (s *service) UpsertConnectorVersion(ctx context.Context, cv *ConnectorVersi
 				Scan(&existingLabels); scanErr != nil {
 				return scanErr
 			}
+			newUser, newApxy := SplitUserAndApxyLabels(cv.Labels)
 			_, existingApxy := SplitUserAndApxyLabels(existingLabels)
-			mergedLabels := MergeApxyAndUserLabels(cv.Labels, existingApxy)
+			mergedApxy := make(Labels, len(existingApxy)+len(newApxy))
+			for k, v := range existingApxy {
+				mergedApxy[k] = v
+			}
+			for k, v := range newApxy {
+				mergedApxy[k] = v
+			}
+			mergedLabels := MergeApxyAndUserLabels(newUser, mergedApxy)
 
 			result, err := sqb.Update(ConnectorVersionsTable).
 				Set("state", cv.State).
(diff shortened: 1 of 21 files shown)


internal/database/connector_version.go
```diff
@@ -390,8 +392,16 @@ func (s *service) UpsertConnectorVersion(ctx context.Context, cv *ConnectorVersi
 				Scan(&existingLabels); scanErr != nil {
 				return scanErr
 			}
+			newUser, newApxy := SplitUserAndApxyLabels(cv.Labels)
```
rmorlok: If we are going to make this change for the connector version resource type, the behavior needs to be consistent for all resource types. Make the change to other resources and make sure there is test coverage for it. Also make sure that from the api layer, the user cannot set system labels (I assume this behavior is unaffected).

internal/database/connector_version.go
```diff
@@ -558,6 +568,37 @@ func (s *service) SetConnectorVersionState(ctx context.Context, id apid.ID, vers
 	})
 }
 
+// DeleteConnector soft-deletes all versions of the connector identified by id.
+// Returns ErrNotFound if no live versions exist for the id.
+func (s *service) DeleteConnector(ctx context.Context, id apid.ID) error {
```
rmorlok, line 564: This logic needs a test.

internal/database/connector_version.go
rmorlok: Done. Extracted `MergeUpsertLabels` in `labels.go` as the shared "stored apxy/ preserved by default, caller-supplied apxy/ overrides per key" helper. Both `UpsertConnectorVersion` and `UpsertActor` (the only two upsert paths in the database layer that take labels and have an existing-row branch) now go through it, so they behave the same way. Tests added in `labels_test.go`, `connector_version_test.go`, and `actor_test.go`.

For the API layer: I audited every `database.Labels(req.Labels).Validate()` call in the route handlers — these were technically allowing apxy/ keys through. They've all been switched to `database.ValidateUserLabels(...)`, which rejects the reserved namespace. I also discovered the namespace-create and encryption-key-create endpoints weren't validating labels at all (so they would have silently persisted apxy/ keys); they now validate too. New route-level tests cover `/actors`, `/connections`, `/connectors`, `/encryption-keys`, and `/namespaces` returning 400 with "reserved" in the message when an apxy/-prefixed label is supplied.

internal/database/connector_version.go
rmorlok: Done — added a `DeleteConnector` subtest block in `internal/database/connector_version_test.go` covering: soft-deletes every live version, returns `ErrNotFound` when no live versions exist, returns `ErrNotFound` when called twice (idempotency), does not affect other connectors, and rejects `apid.Nil`.

Issue #167, opened 2026-04-27
OAuth tests P0: callback state security (CSRF, replay, cross-tenant)


rmorlok:
Parent: #159
Phase: 1 (P0)
Spec: #159 scenario 4
Blocked by: #162

## Cases

- Missing \`state\`.
- Unknown \`state\`.
- Tampered \`state\`.
- Replayed \`state\`.
- Valid state for a different user.
- Valid state for a different tenant/workspace.
- Valid state for a different integration connection.

## Verify

- Callback is rejected.
- Authorization code is not exchanged.
- Tokens are not stored.
- Credentials are not attached to the wrong user, tenant, or connection.
- Suspicious callback is logged without sensitive values.

## Observability

- \`oauth callback rejected\` events with category metadata.

rmorlok:
## Test approach: mixed

This scenario has both interactive and non-interactive legs. Decide per case:

- If the case hinges on user interaction (consent, login, callback navigation), drive it through the marketplace UI with chromedp (see [`integration_tests/oauth2/standard_flow_test.go`](https://github.com/rmorlok/authproxy2/blob/main/integration_tests/oauth2/standard_flow_test.go)).
- If the case hinges on programmatic tampering or token-endpoint behavior, use the go-oauth2-server `/test/*` control plane. A real session cookie can be seeded via chromedp once and reused for follow-on programmatic requests.

rmorlok:
All 7 cases covered by the merged PR series:

- Missing / unknown / tampered / replayed state — #215
- Valid state for a different user (cross-actor) — #216
- Valid state for a different tenant/workspace (cross-namespace) — #226
- Valid state for a different integration connection (cross-connection) — #226

Each case has a structured `oauth callback rejected` event with category metadata, and tests assert callback rejection, no token exchange, no token persisted, and no connection mutation.

Defense-in-depth namespace guards (`namespace_mismatch_actor` / `namespace_mismatch_connection`) are exercised via synthetic state-envelope injection in #226.

PR #213, opened 2026-05-03
Bind OAuth2 state to namespace and encrypt Redis payload

rmorlok:
## Summary
- Adds a `Namespace` field to the OAuth2 round-trip `state` and validates it on callback against both the inbound actor's namespace and the target connection's namespace, closing the cross-tenant replay/forgery vector.
- Replaces the plain-JSON Redis payload with an AES-GCM AEAD blob (encrypted via the global key, stored in `<ekv_id>:<base64>` form). Tampering with the Redis value now fails authentication on decrypt rather than producing a forged state — closes the manual-Redis-mutation attack vector.
- Routes `RecordCancelSessionAfterAuth` and `saveStateToRedis` through a single `writeStateToRedis` helper so the encrypted-write path is never bypassed.

This is PR 2 of the #167 workstream.

## Test plan
- [x] `go test ./internal/auth_methods/oauth2/...` — five new unit tests cover round-trip with namespace, ciphertext-not-plaintext, AEAD tamper rejection, actor-namespace mismatch, connection-namespace mismatch.
- [x] `go test ./...` — full suite passes.
- [x] `./scripts/preflight.sh` — passes.
- [ ] Integration tests on PR (CI).

### DIFF
--- a/internal/auth_methods/oauth2/state.go
+++ b/internal/auth_methods/oauth2/state.go
@@ -2,7 +2,6 @@ package oauth2
 
 import (
 	"context"
-	"encoding"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -15,6 +14,7 @@ import (
 	"github.com/rmorlok/authproxy/internal/config"
 	coreIface "github.com/rmorlok/authproxy/internal/core/iface"
 	"github.com/rmorlok/authproxy/internal/database"
+	"github.com/rmorlok/authproxy/internal/encfield"
 	"github.com/rmorlok/authproxy/internal/encrypt"
 	"github.com/rmorlok/authproxy/internal/httpf"
 	sconfig "github.com/rmorlok/authproxy/internal/schema/config"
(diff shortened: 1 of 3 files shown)


internal/auth_methods/oauth2/state.go
```diff
@@ -31,20 +32,53 @@ type state struct {
 	ExpiresAt              time.Time `json:"expires_at"`
 }
 
-func (s *state) MarshalBinary() ([]byte, error) {
-	return json.Marshal(s)
+func (s *state) IsValid() bool {
+	return s.ActorId != apid.Nil && s.ConnectorId != apid.Nil && s.ConnectionId != apid.Nil && !s.ExpiresAt.IsZero()
```
rmorlok: Shouldn't this also be requiring the namespace to be present?

rmorlok: Good catch — empty namespace is never legitimate (`ValidateNamespacePath` rejects `""`), so leaving it permissive in `IsValid` would let a forged/malformed state through whenever the caller's namespace also happened to be empty. Pushed b3ab731 to require it, with a test covering the empty-state case.

PR #215, opened 2026-05-05
Add OAuth2 callback state security integration tests (#167)

rmorlok:
## Summary

Direct-HTTP rejection tests for the four state-validation scenarios from #167:

- `TestCallbackRejection_MissingState` — no `state` param → category `missing_state`
- `TestCallbackRejection_UnknownState` — bogus apid never persisted → category `unknown_state`
- `TestCallbackRejection_TamperedState` — Redis ciphertext flipped post-`_initiate` → category `tampered_state` (AEAD authentication fails)
- `TestCallbackRejection_ReplayedState` — state consumed by a successful callback, then redelivered → category `unknown_state`

Every test asserts: 302 lands on `error_pages.internal_error`, exactly one `oauth callback rejected` event with the right category, no token row, and zero `/token` calls observed at the provider.

Adds two helpers to support these:
- `helpers.LogCapture` — buffered JSON slog sink swapped in before the dependency manager builds its cached logger, so every derived logger flows through it.
- `env.ForgeOAuth2CallbackURL(state, code)` — builds the kind of callback URL a real browser would never produce (missing/unknown/tampered/replayed state).

Why direct HTTP (not chromedp) for these four cases is documented in `callback_state_security_test.md`. The cross-actor, cross-tenant, and cross-connection cases (5–7 in #167) are PR 4 / PR 5.

Builds on #214 (rejection event emission) and #213 (namespace binding + AES-GCM Redis encryption).

## Test plan

- [x] `go test -tags integration -run TestCallbackRejection ./oauth2/...` — all four pass (1.5s)
- [x] `./scripts/preflight.sh` clean

### DIFF
--- a/integration_tests/helpers/log_capture.go
+++ b/integration_tests/helpers/log_capture.go
@@ -1,0 +1,109 @@
+package helpers
+
+import (
+	"bytes"
+	"encoding/json"
+	"log/slog"
+	"sync"
+	"testing"
+
+	sconfig "github.com/rmorlok/authproxy/internal/schema/config"
+)
+
+// LogCapture is a thread-safe sink for slog records emitted during a test.
+// It plugs into the integration test environment via SetupOptions.LogCapture,
+// replacing the configured logging block with a JSON handler that writes
+// into an in-memory buffer the test can inspect after the fact.
+//
+// Tests use this to assert on observability events the production code
+// emits — e.g., the "oauth callback rejected" events from PR 1 (#214).
+// The capture sits at the root logger level, so any descendant logger
+// (per-service, per-component, per-connection) routes through it.
+type LogCapture struct {
+	mu     sync.Mutex
+	buf    bytes.Buffer
+	level  slog.Level
+	logger *slog.Logger
+}
+
+// NewLogCapture builds a capture that records every record at Debug or
+// above. Tests filter by level/category/etc. when reading.
+func NewLogCapture() *LogCapture {
+	c := &LogCapture{level: slog.LevelDebug}
+	return c
+}
+
+// Records returns every captured record as a parsed map. Each call rescans
+// the entire buffer — cheap relative to the rest of an integration test
+// and preserves the simplest possible API.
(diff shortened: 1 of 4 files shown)


integration_tests/helpers/log_capture.go
```diff
@@ -0,0 +1,109 @@
...
+			t.Fatalf("LogCapture: failed to decode record: %v", err)
+		}
+		records = append(records, rec)
+	}
+	return records
+}
+
+// RejectionEvents returns just the "oauth callback rejected" events. The
+// message string is the public contract from PR 1 — alerts and tests both
+// key off it.
```
rmorlok: This is a generic helper method in the integration test suite. The comment shouldn't be referring to specific requirements.

integration_tests/helpers/log_capture.go
```diff
@@ -0,0 +1,109 @@
...
+func (c *LogCapture) RejectionEvents(t *testing.T) []map[string]any {
+	t.Helper()
+	out := []map[string]any{}
+	for _, r := range c.Records(t) {
+		if r["msg"] == "oauth callback rejected" {
```
rmorlok: Instead of hardcoding this, let's generalize this method to be a generic helper. It should take a string of what the message should be, and callers could use it with the appropriate message (which might be stored as a constant or something for the tests).

integration_tests/helpers/log_capture.go
rmorlok: Removed the oauth-specific references from the LogCapture doc comments and from the SetupOptions.LogCapture field comment in setup.go — both are now framed as generic integration-test helpers.

integration_tests/helpers/log_capture.go
rmorlok: Renamed RejectionEvents → RecordsWithMessage(t, msg) so the helper takes the message string as a parameter. The oauth2 test file now defines a local rejectionEventMessage constant and passes it in, keeping LogCapture domain-agnostic.

Issue #168, opened 2026-04-27
OAuth tests P0: authorization-code exchange failures


rmorlok:
Parent: #159
Phase: 1 (P0)
Spec: #159 scenario 5
Blocked by: #162

## Cases

- Authorization code expired.
- Authorization code already used.
- Invalid authorization code.
- Invalid client ID.
- Invalid client secret.
- Redirect URI mismatch.
- Provider returns \`invalid_grant\`.
- Provider returns \`invalid_client\`.
- Provider token endpoint returns 500/502/503/504.

## Verify

- No connection is created on failed exchange.
- No partial credentials are stored.
- Retryable failures are retried according to policy.
- Non-retryable failures are surfaced clearly.
- Invalid provider/client configuration is distinguished from user authorization failure.

## Observability

- \`token exchange failed\` events with classification metadata.

rmorlok:
## Test approach: go-oauth2-server `/test/*` (preferred)

The primary goal of this scenario is non-interactive (token-endpoint behavior, refresh, revocation, timing, response shape, etc.) — there is no user-flow leg under test. Use the go-oauth2-server `/test/*` control plane to seed state and short-circuit the authorize step rather than booting chromedp.

Reach for chromedp only if the scenario grows to depend on a real consent/login/redirect leg.

rmorlok:
All three PRs merged: #228 (PR A — structured failure events), #245 (PR B — non-retryable rejection tests), #248 (PR C — retry policy + transient/5xx tests).

PR #228, opened 2026-05-09
Emit structured oauth token exchange failure events (#168)

rmorlok:
## Summary

Foundational PR for #168. Adds a stable category enum + `oauth token exchange failed` slog events for the authorization-code → token exchange leg, mirroring the `oauth callback rejected` pattern from #214. The follow-up rejection-test and retry-test PRs (PRs B and C of the #168 workstream) key their assertions off this category metadata.

- `tokenExchangeCategory` enum covers RFC 6749 §5.2 provider responses (`invalid_grant`, `invalid_client`, `invalid_request`, `unauthorized_client`, `unsupported_grant_type`, `invalid_scope`) plus operational failure modes (`network_error`, `provider_5xx`, `provider_4xx_other`, `malformed_response`, `provider_denied`, `missing_code`, `state_cleanup_error`, `internal_error`).
- `classifyTokenEndpointStatus(status, body)` maps the response to a category. 5xx is always transient regardless of body so retry policy (PR C) can key off category without re-parsing. 4xx parses the RFC error JSON; unrecognized values fall through to `provider_4xx_other` while preserving the raw `provider_error` field for diagnostics.
- `emitTokenExchangeFailure` is the single entrypoint, populated from `tokenExchangeAttrsFromConn` so callers don't have to repeat boilerplate.

## Behavior unchanged

Failures still flow through `HandleAuthFailed` → `SetupStepAuthFailed` on the connection — only observability is added. No retry behavior in this PR; PR C introduces a bounded transient-retry policy keyed on `provider_5xx` / `network_error`.

## Test plan

- [x] `go test ./internal/auth_methods/oauth2/...` — 11 new unit tests for emit + classify
- [x] `go test ./integration_tests/oauth2 -tags=integration -run 'TestStandardFlow|TestUserDenialFlow|TestCallbackRejection_'` — no regressions
- [x] `./scripts/preflight.sh` — passes


integration_tests/helpers/setup.go
rmorlok: We are starting to get to a lot of parameters with several different methods to initiate oauth2 connections.

Let's move this to and related methods to a InitiateOAuth2Connection(t *testing.T, connectorID apid.ID, returnToUrl, options OAuth2Option...) style where options could cover things like namespace, etc. This would allow us to consolidate InitiateOAuth2ConnectionAsActor and  InitiateOAuth2Connection.

rmorlok: Same here. Move to an option pattern to allow for these other things to be specified rather than having multiple methods to accomplish the additional specificity.

rmorlok: This setup.go file is getting too large and has a lot responsibilities. Let's split it out where there would be setup_oauth2.go which would cover the IntegrationTestEnv methods that are specific to OAuth2. If there are other broad categories that would make sense, do that as well.

rmorlok: Done — `InitiateOAuth2Connection` now accepts variadic `OAuth2Option` values. The `…AsActor` variant is gone; callers pass `helpers.WithActor(externalID, namespace)` when they need to override the default ("test-actor" in root). Same option type is shared across the OAuth2 helpers (Initiate, DeliverCallback) so callers learn one pattern.

rmorlok: Done — `DeliverOAuth2Callback` is now `(t, callbackURL, opts ...OAuth2Option)`; the `…AsActor` variant is gone. Same `helpers.WithActor(externalID, namespace)` option works here.

rmorlok: Addressed: extracted all OAuth2-specific helpers from `setup.go` into `integration_tests/helpers/setup_oauth2.go`. The new file owns `OAuth2ConnectorOptions`, `NewOAuth2Connector`, `PublicOAuthCallbackURL`, `ForgeOAuth2CallbackURL`, `InitiateOAuth2Connection`, `FollowOAuth2Redirect`, `DeliverOAuth2Callback`, `OAuth2StateForTest`, `WriteOAuth2StateForTest`, and `GetOAuth2Token`, plus the new `OAuth2Option` / `WithActor` plumbing.

Issue #224, opened 2026-05-08
Terraform provider: authproxy_rate_limit resource (native HCL)


rmorlok:
Part of #3. Depends on the admin-api CRUD PR.

Add a new `authproxy_rate_limit` resource to the Terraform provider at `terraform/provider/`. Unlike `authproxy_connector`, this resource exposes every field as native HCL (typed nested blocks), not as a JSON-blob string — so authors get autocomplete, validation, and field-level plan diffs without `jsonencode(...)`.

## Scope

- New resource type registered with the provider
- HCL schema with typed nested blocks for `selector` (`label_selector`, `methods`, `path_match`, `request_types`), `bucket` (`dimensions`), and `algorithm` (one of `fixed_window` / `sliding_window` / `token_bucket`)
- Plan-time validation that exactly one algorithm subtype is set
- Mutable in place via `PATCH`; namespace change forces replacement
- Client methods in `terraform/provider/internal/client` calling the admin-api CRUD endpoints
- Acceptance tests covering create / read / update / delete and switching algorithm subtype (replace)
- Generated docs in `terraform/provider/docs/resources/`
- Example under `terraform/provider/examples/resources/authproxy_rate_limit/`

## Out of scope

- Server-side resource model / API (already in earlier PRs)

## Acceptance criteria

- [ ] Resource registered, schema validated
- [ ] Acceptance tests pass
- [ ] Docs and example committed
- [ ] No `jsonencode(...)` blobs anywhere in the example

Issue #223, opened 2026-05-08
Rate limits: proxy-path enforcement


rmorlok:
Part of #3. Depends on the sync, evaluation, algorithms/store, and request-log attribution PRs.

Wire the cached rules, evaluation engine, and counter store into the proxy request path so rate limits actually take effect end-to-end.

## Scope

- Hook in the proxy round-tripper (or a thin pre-send middleware) that for each governed request:
  1. Resolves the request's type, verb, final upstream URL, label snapshot, and reserved bucket-dimension values
  2. Iterates currently-cached rules; calls evaluation engine to find matches
  3. For each match in `enforce` mode, calls into the algorithm/store to check + increment
  4. If any enforce-mode rule rejects, returns a synth 429 with `Retry-After` and an identifying body; writes a request-log entry with `ResponseSource = rate_limit`, `RateLimitId`, `RateLimitMode = enforce`, `RateLimitBucket`, and the full match set in `RateLimitMatched`
  5. If only observe-mode rules match (or in addition to passing enforce-mode rules), passes the request through but writes log entry with the observe attribution populated
- Coexistence with existing connector-level reactive 429 — both can fire on the same request
- Handles `proxy` and `probe` request types by default (other types governed only when explicitly opted in by a rule's `requestTypes`)

## Out of scope

- Schema, sync, evaluation, algorithms, log-record fields — all exist by the time this PR lands

## Acceptance criteria

- [ ] End-to-end integration tests covering: single rule rejection, multi-rule "all apply" most-restrictive, observe mode, bucket isolation, fail-open on Redis outage, coexistence with connector-level reactive 429
- [ ] Synth 429 body and Retry-After header are correct per algorithm
- [ ] Request-log entries on rejection and observe-only matches carry full attribution
- [ ] Probe traffic governed by default; OAuth2 token/refresh/revocation traffic governed only when explicitly listed in `requestTypes`

Issue #222, opened 2026-05-08
Request log: attribute 429s to upstream, connector limiter, or rate-limit resource


rmorlok:
Part of #3. Mostly independent of the rate-limit resource PRs except where it stamps fields they will populate.

Make every 429 in the request log unambiguous about its source. Adds new fields to `LogRecord`, migrates the SQL and Clickhouse providers, and updates the existing connector-level reactive 429 path to write a properly-attributed entry when it short-circuits a request — closing the current gap where those synthetic 429s look identical to real upstream ones.

## Scope

- `LogRecord` (`internal/request_log/log_record.go`) gains:
  - `ResponseSource` enum: `upstream` (default) | `connector_rate_limiter` | `rate_limit`
  - `RateLimitId` (apid.ID, populated when a `RateLimit` resource matched, including observe mode)
  - `RateLimitMode` (`enforce` | `observe`)
  - `RateLimitBucket` (resolved dimension → value map)
  - `RateLimitMatched` (full match set as `[]{id, mode, bucket}`)
- Migrations + columns in `provider_sql.go` and `provider_clickhouse.go`
- Update `internal/ratelimit/roundtripper.go`:
  - Real upstream 429 → entry with `ResponseSource = upstream`
  - Connection-in-cooldown short-circuit → entry with `ResponseSource = connector_rate_limiter`, `ResponseStatusCode = 429` (this is the bug fix)
- Request-log list endpoint accepts `response_source` and `rate_limit_id` filters
- The new `rate_limit` source value is reserved here but populated by the enforcement PR

## Out of scope

- Rate-limit resource code paths (separate PRs); this PR only adds the schema and updates the existing 429 path

## Acceptance criteria

- [ ] LogRecord schema additions + migrations for both providers
- [ ] Existing connector-level reactive limiter writes log entries with `ResponseSource = connector_rate_limiter` on short-circuit; tests assert this
- [ ] Real upstream 429s log with `ResponseSource = upstream`; tests assert this
- [ ] List endpoint filters work; tested
- [ ] No behavioral change to non-429 responses (default `upstream` keeps existing entries valid)

Issue #221, opened 2026-05-08
Rate limits: algorithms and Redis-backed counter store


rmorlok:
Part of #3. Depends on the schema/DB foundation PR.

Implement the three rate-limit algorithms backed by a shared Redis counter store, including fail-open behavior on Redis unavailability.

## Scope

- Counter store interface and Redis implementation, sharded by `(rate_limit_id, bucket_key)`
- Lua scripts for atomic check-and-increment per algorithm
- Algorithms:
  - `fixed_window` (`window`, `limit`)
  - `sliding_window` (`window`, `limit`, `mode: log | counter`)
  - `token_bucket` (`capacity`, `refill_rate`)
- `Retry-After` computation per algorithm (remaining-window for window algorithms, time-to-next-token for token bucket)
- Fail-open on Redis unavailability: allow the request, increment a metric, emit a structured warning

## Out of scope

- Selector matching / bucket resolution (separate PR)
- Enforcement integration (separate PR) — algorithms are unit-tested here in isolation

## Acceptance criteria

- [ ] Each algorithm has unit tests for window boundaries, refill behavior, and concurrent increments (via miniredis or equivalent)
- [ ] `sliding_window` covers both `log` and `counter` modes
- [ ] Retry-After values are correct under each algorithm
- [ ] Redis-unavailable path returns "allow + warn" without panicking; tested

PR #242, opened 2026-05-09
Rate limits — Step 3: per-process cache + refresher

### LINKED ISSUE #219: Rate limits: background sync to proxy-node cache
Part of #3. Depends on the schema/DB foundation PR.

Add a background sync task that pulls `RateLimit` definitions from the database into an in-memory cache on each proxy node, modeled on the encryption-key sync (`internal/encrypt/task_sync_keys_to_database.go`).

## Scope

- Redis sentinel (timestamp key) and mutex pattern matching the encryption-key sync
- In-memory rule cache exposed via an interface that the enforcement layer (later PR) will consume — thread-safe reads, atomic swap on refresh
- Configurable interval, defaulting to 5 minutes
- Initial sync at process start before the proxy begins serving

## Out of scope

- Selector/bucket evaluation engine and the algorithms (separate PRs) — this PR only owns getting fresh definitions into memory

## Acceptance criteria

- [ ] Task implementation + registration; sentinel + mutex behavior tested
- [ ] Cache returns the latest sync's rules; partial-failure modes don't blow away the cache (keep last-known-good)
- [ ] Interval is configurable; default 5 minutes
- [ ] Tests cover initial sync, periodic re-sync, and concurrency (multiple workers competing for the mutex)

rmorlok:
## Summary

Closes #219. Third of the eight sub-issues under #3 (rate limit support); builds on #217 (schema/DB) and lands alongside #218 (admin/api CRUD).

Adds the in-memory rate-limit rule cache + the background goroutine that keeps it fresh from the database. The proxy enforcement layer (#223) will consume the read-only \`Cache\` interface to look up matching rules at request time without round-tripping to the database.

## What's in

**`internal/ratelimit/cache.go`** — \`Cache\` (read-only) and \`MutableCache\` (Replace) interfaces, plus a concrete atomic-swap impl. Reads return a copied slice so callers can never mutate cache state. Snapshot version increments monotonically on each Replace.

**`internal/ratelimit/refresher.go`** — \`Refresher\` with \`Sync()\` (single pull) and \`Run()\` (periodic loop). Last-known-good preservation: a failed DB query leaves the previous snapshot in place. Configurable interval, default 5 minutes; minimum floor 5s to guard against pathological config. \`StartRefresher(ctx, ...)\` is the convenience wrapper used by the proxy bootstrap.

**`internal/service/dependency_manager.go`** — \`GetRateLimitCache()\` lazy singleton + \`StartRateLimitRefresher(ctx)\` lifecycle helper.

**`internal/service/api/server.go`** and **`admin_api/server.go`** — boot the refresher before the gin server starts serving, so the initial snapshot lands before any rate-limit evaluation can run.

## Architectural note (vs. issue text)

The ticket called for an asynq cron task plus Redis sentinel + mutex "matching the encryption-key sync". Those mechanisms make sense when the cache is shared (one writer, many readers); ours is per-process in-memory, so:

- An asynq task running on the worker can't populate per-process caches in api/admin-api.
- A Redis mutex would only serialise refreshes that per-process caches don't share.

Until proxy fleets grow large enough to stress the database from synchronised refreshes, a self-contained per-process goroutine is the simpler primitive that meets the rest of the spec (thread-safe reads, atomic swap, last-known-good, configurable interval, initial sync at process start). Cross-process invalidation on admin-API writes is **deferred** — current behaviour is "wait for the next tick" — and is best added later via Redis pub/sub when actually needed. Happy to revisit if you'd prefer the asynq+Redis design upfront.

## Test plan

- [x] \`go test -race ./internal/ratelimit/...\` — 15 sub-tests passing including concurrent-read/replace under the race detector
- [x] \`go test ./...\` — full repo green

## Out of scope (later sub-issues)

- Selector / bucket evaluation engine (#220)
- Algorithms + Redis counter store (#221)
- Request-log 429 attribution + connector-limiter fix (#222)
- Proxy-path enforcement (#223 — consumer of \`Cache\`)
- Terraform \`authproxy_rate_limit\` resource (#224)

### DIFF
--- a/internal/service/api/server.go
+++ b/internal/service/api/server.go
@@ -180,6 +180,11 @@ func Serve(cfg config.C) {
 		panic(err)
 	}
 
+	// Boot the rate-limit cache refresher before serving traffic so the
+	// initial snapshot lands before the proxy starts evaluating rules.
+	stopRateLimitRefresher := dm.StartRateLimitRefresher(context.Background())
+	defer stopRateLimitRefresher()
+
 	var wg sync.WaitGroup
 	wg.Add(1)
 	go func() {
--- a/internal/ratelimit/cache.go
+++ b/internal/ratelimit/cache.go
@@ -1,0 +1,91 @@
+package ratelimit
+
+import (
+	"sync"
+	"time"
+
+	"github.com/rmorlok/authproxy/internal/database"
+)
+
+// Cache is the read-side of the in-memory rate-limit rule cache. The
+// proxy-path enforcement layer (#223) consumes this interface to look up
+// matching rules at request time without round-tripping to the database.
+//
+// Reads are safe for concurrent use; values returned from All() are a
+// snapshot — callers must not mutate them, and a subsequent Replace() will
+// not affect the slice they hold.
+type Cache interface {
+	// All returns every rule in the current snapshot. The returned slice is
+	// owned by the caller and won't be mutated by concurrent Replace() calls.
+	All() []*database.RateLimit
+
+	// SnapshotTime is when the current snapshot was last installed via
+	// Replace(). Zero if no snapshot has been installed yet.
+	SnapshotTime() time.Time
+
+	// SnapshotVersion increments each time a snapshot is installed. Useful
+	// for callers that want to debounce repeated cache reads or detect
+	// staleness.
+	SnapshotVersion() uint64
+}
+
+// MutableCache extends Cache with the write entry-point used by the Refresher.
+// Kept separate from Cache so the enforcement layer's dependency surface stays
+// read-only.
+type MutableCache interface {
+	Cache
+
+	// Replace atomically swaps the snapshot. The supplied slice is taken as-is
+	// — callers must not mutate it after the call.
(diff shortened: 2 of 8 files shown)


internal/service/api/server.go
```diff
@@ -180,6 +180,11 @@ func Serve(cfg config.C) {
 		panic(err)
 	}
 
+	// Boot the rate-limit cache refresher before serving traffic so the
+	// initial snapshot lands before the proxy starts evaluating rules.
+	stopRateLimitRefresher := dm.StartRateLimitRefresher(context.Background())
```
rmorlok, line 185: Isn't this also going to be needed on the public service to deal with the fact that we serve connection initiate dropbowns from calls to the 3rd party, which ultimately requires proxy calls?

rmorlok: Yes — confirmed and fixed in commit 36a0ab6. The public service runs both `NewConnectionsProxyRoutes` (when `EnableProxy()` is true) and connection-initiate, which triggers setup-flow calls to 3rd parties for the dropdown population you mentioned. Boot the same refresher at process start there.

PR #259, opened 2026-05-10
Rate limits — Step 6: 429 attribution in the request log

### LINKED ISSUE #222: Request log: attribute 429s to upstream, connector limiter, or rate-limit resource

rmorlok:
## Summary

Closes #222. Sixth of the eight sub-issues under #3. Mostly independent of the other rate-limit PRs except where it stamps fields they will populate.

Adds attribution fields to the request log so every 429 records its source — closing the existing gap where the connector-level reactive limiter's synthetic 429s were indistinguishable from real upstream ones, and reserving a value for the proxy-side rate-limit resource enforcement (#223).

## What's in

**LogRecord schema additions**
- \`ResponseSource\` enum: \`upstream\` (default) | \`connector_rate_limiter\` | \`rate_limit\`. The \`rate_limit\` value is reserved here; population happens in #223.
- \`RateLimitId\`, \`RateLimitMode\` (\`enforce\`|\`observe\`), \`RateLimitBucket\` (resolved dimension → value map), and \`RateLimitMatched\` (full match set as a list of \`{id, mode, bucket}\`). Populated by the proxy-side rate-limit enforcement (#223); the connector-level reactive limiter does not populate these (it has no rule id).

**Plumbing — context-shared Attribution pointer**
- \`internal/request_log/attribution.go\` — \`ResponseSource\` constants + \`IsValidResponseSource\`, \`RateLimitMatch\` struct, \`Attribution\` struct + context helpers. Pointer is shared across the round-tripper chain so inner middlewares mutate the same struct that the outer log-writer reads on the way back.
- request-log RoundTripper installs an empty Attribution on the request context if one isn't already present, then calls \`ApplyAttributionToLogRecord(ctx)\` after \`SetLogRecordFieldsFromRequestInfo\` so the LogRecord is populated regardless of which middleware stamped attribution.
- ratelimit RoundTripper writes \`Source=connector_rate_limiter\` on the synthetic-429 short-circuit. Real upstream 429s leave the Attribution alone — they fall through to the default \`upstream\`.

**Storage**
- Migrations \`000002\` for postgres and sqlite. New columns: \`response_source TEXT NOT NULL DEFAULT 'upstream'\`, \`rate_limit_id\` / \`rate_limit_mode\` TEXT, \`rate_limit_bucket\` / \`rate_limit_matched\` JSON. Indexes on \`response_source\` and (\`rate_limit_id\` where non-empty).
- ClickHouse \`Migrate()\` DDL extended with the same columns.
- SQL + ClickHouse insert paths and scan path round-trip the new fields. Bucket and Matched marshal via \`attribution_codec.go\` which treats nil/empty as \`{}\` / \`[]\` so the columns always hold valid JSON.

**List filters**
- \`ListFilters\` gains \`ResponseSource\` + \`RateLimitId\`; \`SetX\` setters and \`WithX\` builder methods on the public \`ListRequestBuilder\` interface.
- SQL + ClickHouse + Mock builders all forward to the SQL \`buildQuery()\` that emits \`sq.Eq\` filters.
- \`ListRequestsQuery\` in routes binds \`response_source\` and \`rate_limit_id\` form params; \`ApplyToBuilder\` validates \`response_source\` against the enum.

## Notes for the reviewer

- **Why a context-pointer instead of a header**: stamping a header on the synthetic response (e.g., extending the existing \`X-Authproxy-Ratelimited: true\`) would also work for distinguishing connector_rate_limiter from upstream, but the rate-limit resource enforcement (#223) needs to convey rich data (rule id, bucket map, full match set) that doesn't belong on the wire. A pointer in the context handles both cases uniformly.
- **Default backfill**: existing rows get \`response_source = 'upstream'\` via the \`NOT NULL DEFAULT\` migration. New writes always set it explicitly via \`ApplyAttributionToLogRecord\`, which falls back to upstream when no attribution was installed.
- **No behavioural change to non-429 responses**: a successful 200 sees \`response_source = upstream\` (the default). Existing entries keep validating.

## Test plan

- [x] \`go test ./internal/request_log/ ./internal/ratelimit/\` — 9 + 2 new sub-tests pass
- [x] \`go test ./...\` — full repo green

## Out of scope (later sub-issues)

- Proxy-path enforcement (#223 — populates the \`rate_limit\` source value + the RateLimit* fields)
- Terraform \`authproxy_rate_limit\` resource (#224)

### DIFF
--- a/internal/request_log/attribution_codec.go
+++ b/internal/request_log/attribution_codec.go
@@ -1,0 +1,64 @@
+package request_log
+
+import "encoding/json"
+
+// marshalRateLimitBucket renders the bucket map to a stable JSON form
+// suitable for storage. An empty / nil map serializes to "{}" so the DB
+// column always holds valid JSON (matches the migration default).
+func marshalRateLimitBucket(b map[string]string) (string, error) {
+	if len(b) == 0 {
+		return "{}", nil
+	}
+	out, err := json.Marshal(b)
+	if err != nil {
+		return "", err
+	}
+	return string(out), nil
+}
+
+// unmarshalRateLimitBucket parses a bucket JSON blob from the DB. nil /
+// empty / "{}" all yield a nil map so callers don't need to special-case
+// "missing" vs "explicitly empty" — semantically the same for log
+// readers.
+func unmarshalRateLimitBucket(data []byte) (map[string]string, error) {
+	if len(data) == 0 {
+		return nil, nil
+	}
+	var out map[string]string
+	if err := json.Unmarshal(data, &out); err != nil {
+		return nil, err
+	}
+	if len(out) == 0 {
+		return nil, nil
+	}
+	return out, nil
+}
+
+// marshalRateLimitMatched renders the matched list to JSON. Empty / nil
+// → "[]" so the DB column always holds valid JSON.
+func marshalRateLimitMatched(m []RateLimitMatch) (string, error) {
+	if len(m) == 0 {
+		return "[]", nil
+	}
+	out, err := json.Marshal(m)
+	if err != nil {
+		return "", err
+	}
+	return string(out), nil
+}
+
+// unmarshalRateLimitMatched parses a matched-list JSON blob. nil / empty
+// / "[]" yield a nil slice.
+func unmarshalRateLimitMatched(data []byte) ([]RateLimitMatch, error) {
+	if len(data) == 0 {
+		return nil, nil
+	}
+	var out []RateLimitMatch
+	if err := json.Unmarshal(data, &out); err != nil {
+		return nil, err
+	}
+	if len(out) == 0 {
+		return nil, nil
+	}
+	return out, nil
+}
--- a/internal/request_log/request_info.go
+++ b/internal/request_log/request_info.go
@@ -1,6 +1,8 @@
 package request_log
 
 import (
+	"context"
+
 	"github.com/rmorlok/authproxy/internal/httpf"
 )
 
@@ -16,4 +18,43 @@ func SetLogRecordFieldsFromRequestInfo(er *LogRecord, ri httpf.RequestInfo) {
 	er.ConnectorVersion = ri.ConnectorVersion
 	er.ConnectionId = ri.ConnectionId
 	er.Labels = ri.Labels
+}
+
+// ApplyAttributionToLogRecord stamps the LogRecord with whatever the
+// proxy stack recorded on the request context — the response source
+// plus, when a rate-limit resource matched, the rule id / mode / bucket
+// / full match set. When no attribution was installed (older code paths
+// or non-proxy traffic), defaults the source to ResponseSourceUpstream
+// so the column is always populated.
+//
+// Designed as a separate call from SetLogRecordFieldsFromRequestInfo so
+// the existing signature isn't disturbed; the request-log round-tripper
+// invokes both in sequence.
+func ApplyAttributionToLogRecord(er *LogRecord, ctx context.Context) {
+	attr := AttributionFromContext(ctx)
+	if attr == nil {
+		if er.ResponseSource == "" {
+			er.ResponseSource = ResponseSourceUpstream
+		}
+		return
+	}
+
+	if attr.Source != "" {
+		er.ResponseSource = attr.Source
+	} else if er.ResponseSource == "" {
+		er.ResponseSource = ResponseSourceUpstream
+	}
+
+	if !attr.RateLimitId.IsNil() {
+		er.RateLimitId = attr.RateLimitId
+	}
+	if attr.RateLimitMode != "" {
+		er.RateLimitMode = attr.RateLimitMode
+	}
+	if len(attr.RateLimitBucket) > 0 {
+		er.RateLimitBucket = attr.RateLimitBucket
+	}
+	if len(attr.RateLimitMatched) > 0 {
+		er.RateLimitMatched = attr.RateLimitMatched
+	}
 }
--- a/internal/routes/request_log.go
+++ b/internal/routes/request_log.go
@@ -43,6 +43,8 @@ type ListRequestsQuery struct {
 	Path                     *string  `form:"path"`
 	PathRegex                *string  `form:"path_regex"`
 	LabelSelector            *string  `form:"label_selector"`
+	ResponseSource           *string  `form:"response_source"`
+	RateLimitId              *apid.ID `form:"rate_limit_id" swaggertype:"string"`
 }
 
 func (q *ListRequestsQuery) ApplyToBuilder(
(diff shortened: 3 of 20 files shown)


internal/routes/request_log.go
```diff
@@ -43,6 +43,8 @@ type ListRequestsQuery struct {
 	Path                     *string  `form:"path"`
 	PathRegex                *string  `form:"path_regex"`
 	LabelSelector            *string  `form:"label_selector"`
+	ResponseSource           *string  `form:"response_source"`
+	RateLimitId              *apid.ID `form:"rate_limit_id" swaggertype:"string"`
```
rmorlok, line 47: Check if any future tickets address updating the sdks/js. If not, add the update in this PR.

internal/routes/request_log.go
```diff
@@ -43,6 +43,8 @@ type ListRequestsQuery struct {
 	Path                     *string  `form:"path"`
 	PathRegex                *string  `form:"path_regex"`
 	LabelSelector            *string  `form:"label_selector"`
+	ResponseSource           *string  `form:"response_source"`
```
rmorlok, line 46: Same as the othe rcomment, make sure this gets added to the js sdk. Also update the admin UI with this and the rate limit id columns. Response source should be visible by default, rate limit id would be hidden by default. Also the rate limit id field should link to the rate limiter page in admin ui that needs to be created.

internal/request_log/request_info.go
```diff
@@ -17,3 +19,42 @@ func SetLogRecordFieldsFromRequestInfo(er *LogRecord, ri httpf.RequestInfo) {
 	er.ConnectionId = ri.ConnectionId
 	er.Labels = ri.Labels
 }
+
+// ApplyAttributionToLogRecord stamps the LogRecord with whatever the
+// proxy stack recorded on the request context — the response source
+// plus, when a rate-limit resource matched, the rule id / mode / bucket
+// / full match set. When no attribution was installed (older code paths
+// or non-proxy traffic), defaults the source to ResponseSourceUpstream
+// so the column is always populated.
```
rmorlok, line 28: Are there other places in the codebase where we should be looking to apply this attribution now so that it's consistently applied?

internal/request_log/attribution_codec.go
```diff
@@ -0,0 +1,64 @@
+package request_log
+
+import "encoding/json"
+
+// marshalRateLimitBucket renders the bucket map to a stable JSON form
+// suitable for storage. An empty / nil map serializes to "{}" so the DB
+// column always holds valid JSON (matches the migration default).
+func marshalRateLimitBucket(b map[string]string) (string, error) {
```
rmorlok, line 8: Please add tests for functions in this file.

internal/routes/request_log.go
rmorlok: Added in commit bc5639b. `sdks/js/src/requests.ts` now exposes `response_source` and `rate_limit_id` on `ListRequestsParams`, plus a new `ResponseSource` enum, `RateLimitMatch` interface, and the `rate_limit_*` fields on `RequestEntryRecord`. `dist/` is git-ignored so consumers regen with `yarn build`.

internal/routes/request_log.go
rmorlok: Added in commit bc5639b. Admin UI `Requests` page gains two columns:

- **Source** (visible by default) — renders `response_source` as a small chip. `upstream` is default color; the two synthetic-429 sources use `warning` so operators spot them at a glance.
- **Rate Limit** (`hideInitial: true`) — shows `rate_limit_id` when present, rendered as a `<Link to=`/rate-limits/{id}`>` (react-router-dom) so it'll resolve once the rate-limiter detail page is added.

SDK side also covered (see prior comment).

internal/request_log/request_info.go
rmorlok: Audited the codebase:

**LogRecord writers**: only `internal/request_log/roundtripper.go` (two paths: error at line ~136 and success at line ~230). Both already call `ApplyAttributionToLogRecord(record, ctx)`. `internal/request_log/full_log.go:107` constructs a LogRecord via `ToRecord()` but doesn't store it — its callers are the same two paths in the roundtripper.

**429 producers**:
- `internal/ratelimit/roundtripper.go:218` — synthetic short-circuit, stamped with `connector_rate_limiter` ✅
- `internal/ratelimit/roundtripper.go:102` — real upstream 429 path, correctly leaves attribution unset → defaults to `upstream` ✅
- `internal/schema/config/key_data_vault_retry.go:34` — vault-key retry on 429. This is an internal request from authproxy to a key-storage vault, doesn't go through the proxy request log, so no attribution needed.

So the two call sites already added are the complete set.

internal/request_log/attribution_codec.go
rmorlok: Added `attribution_codec_test.go` in commit bc5639b. 10 sub-tests covering both marshallers + both unmarshallers: nil/empty produce valid JSON (`{}` / `[]`) so the DB's NOT NULL invariant holds, round-trip preserves order for the matched list, and bad JSON surfaces as an error from the unmarshaler.

Issue #220, opened 2026-05-08
Rate limits: selector matching and bucket key resolution


rmorlok:
Part of #3. Depends on the schema/DB foundation PR.

Pure-function evaluation layer: given a `RateLimit` rule and a request context, decide whether the rule matches and compute the bucket key. No Redis, no algorithms, no enforcement — just deterministic evaluation that the enforcement PR will call into.

## Scope

- `Match(rule, requestContext) (matched bool, bucketKey BucketKey)` style API
- Selector clauses (all ANDed):
  - `labelSelector` evaluated against the per-request label snapshot using existing `internal/database/label_selector.go`
  - `methods` membership
  - `pathMatch` evaluated against the **final upstream URL** (not the proxy-input URL); kinds: `prefix`, `glob`, `regex`
  - `requestTypes` membership (default `[proxy, probe]` when unspecified on the rule)
- Bucket-key resolution:
  - Reserved dimensions: `actor`, `connection`, `connector`, `connector_version`, `namespace`, `method`
  - `labels/<key>` references the per-request label snapshot
  - Empty dimensions list = single global bucket per rule
  - Stable, deterministic serialization of the resolved tuple

## Out of scope

- Algorithm execution and Redis counters
- Request log writes
- Enforcement wiring

## Acceptance criteria

- [ ] Unit tests for each selector clause (including negative cases) and combinations
- [ ] Unit tests for bucket-key resolution including missing-dimension behavior and label references
- [ ] `pathMatch` test cases for each `kind`, including edge cases for glob/regex
- [ ] Default `requestTypes` covered

PR #264, opened 2026-05-11
Rate limits — Step 7: proxy-path enforcement

### LINKED ISSUE #223: Rate limits: proxy-path enforcement

rmorlok:
## Summary

Closes #223. Seventh of the eight sub-issues under #3 — the integration centrepiece that ties together the cache (#219), matcher (#220), algorithms (#221), and log-attribution (#222) into the proxy request path so rate limits actually take effect end-to-end.

## What's in

**`internal/ratelimit/enforcer.go`** — `EnforcerFactory` implementing `httpf.RoundTripperFactory` + `EnforcerRoundTripper`. On each request:

1. Iterates the in-memory rule cache and runs \`Match()\` on each rule. Request type, methods, label selector, path match all ANDed; the rule's \`requestTypes\` clause filters by traffic type (default \`[proxy, probe]\`).
2. For every match (enforce or observe), constructs a \`Limiter\` and calls \`Decide\` — both modes keep counters hot so flipping a rule from observe to enforce shouldn't reset its bucket.
3. Enforce-mode rejections are collected; the one with the longest \`Retry-After\` wins (most-restrictive). The request short-circuits with a synth 429 carrying \`X-Authproxy-Ratelimited\`, \`X-Authproxy-Ratelimit: <rule id>\`, and a JSON body \`{ rate_limit_id, retry_after_seconds }\`.
4. Stamps \`request_log.Attribution\` — firing rule on top-level fields (\`Source = rate_limit\`), plus the full \`RateLimitMatched\` list so observers see every rule that contributed. When only observe matches occurred, \`Source\` stays default (upstream — the response really came from upstream) but the observe match's id/mode/bucket are recorded.

**`internal/ratelimit/bucket_key.go`** — added \`AsMap()\` helper for storing the resolved bucket on \`LogRecord.RateLimitBucket\`.

**`internal/httpf/factory.go`** — **middleware-ordering bugfix.** The previous loop made \`request_log\` *innermost* despite the comment claiming it was outermost. That meant inner middlewares (the existing reactive 429 limiter, and now the enforcer) saw a nil \`Attribution\` from #222 and their stamping was a silent no-op in production. Moved \`request_log\` to the end of the middleware slice so it actually becomes outermost in execution order; the comment now matches the behaviour. **As a side benefit, synthetic 429s from either rate limiter now show up in the request log** — closing a latent gap that pre-dated this work.

**`internal/service/dependency_manager.go`** — \`GetRateLimitEnforcerFactory\` pulls the cache + Redis client. \`GetHttpf\` registers the enforcer ahead of the connector-level reactive \`Factory\` so a proxy-side rule rejection short-circuits before any upstream 429 backoff is applied.

## Notes for the reviewer

- **Most-restrictive selection**: when multiple enforce rules reject, the one with the longest \`Retry-After\` wins. Implementations that prefer "smallest remaining-capacity" would also be defensible; the issue spec just says "most-restrictive wins". I went with the wait-time interpretation because that's what callers actually observe in the response.
- **Why observe mode increments counters**: so an operator validating a rule in production sees the same behaviour the enforce-mode flip will produce, and the bucket is already warmed up. Cost: observe mode does the Redis round-trip even though it can't reject.
- **Pulling actor id from labels**: \`RequestInfo\` doesn't carry \`ActorID\` directly — \`httpf.ForActor()\` stamps \`apxy/act/-/id\` on the per-request label snapshot. The enforcer reads that. Tradeoff is a string parse vs. a wider \`RequestInfo\` struct; chose the former to keep the existing surface unchanged. Constant lives in \`enforcer.go\` (\`actorIDLabelKey\`).
- **Middleware ordering fix**: I called it out above and in the commit message — it's a small but consequential change. Re-tested every \`./internal/...\` package; nothing else depended on the wrong order.

## Test plan

- [x] \`go test ./internal/ratelimit/\` — 14 new sub-tests pass (each acceptance-criterion bullet has at least one)
- [x] \`go test -race ./internal/ratelimit/\` — race-detector clean
- [x] \`go test ./...\` — full repo green, including the existing factory tests against the new ordering

## What's left for #3

- Terraform \`authproxy_rate_limit\` resource (#224) — only remaining sub-issue.

### DIFF
--- a/internal/httpf/factory.go
+++ b/internal/httpf/factory.go
@@ -36,14 +36,21 @@ func CreateFactory(
 	logger *slog.Logger,
 	additionalMiddlewares ...RoundTripperFactory,
 ) F {
-	// Order matters here to determine the order of middlewares.
-	// Request logging is outermost so all requests (including rate-limited ones) are logged.
-	middlewares := []RoundTripperFactory{requestLog}
+	// Order matters: in New() below, each middleware wraps the previous
+	// parent, so the *last* entry in this slice becomes the outermost in
+	// execution order (the first to see the request, the last to see the
+	// response on the way back). Request logging must be outermost so it
+	// records every request — including ones that inner middlewares
+	// short-circuit with a synth 429 — and so it can install the
+	// per-request Attribution onto the context before any inner middleware
+	// reads or writes it.
+	var middlewares []RoundTripperFactory
 	for _, m := range additionalMiddlewares {
 		if m != nil {
 			middlewares = append(middlewares, m)
 		}
 	}
+	middlewares = append(middlewares, requestLog)
 
 	return &clientFactory{
 		cfg:         cfg,
--- a/internal/ratelimit/bucket_key.go
+++ b/internal/ratelimit/bucket_key.go
@@ -29,6 +29,23 @@ func (k BucketKey) IsGlobal() bool {
 	return len(k.Components) == 0
 }
 
+// AsMap renders the bucket key as a map suitable for storage on a request
+// log entry (LogRecord.RateLimitBucket). Order is lost — the map is for
+// human consumption, not for downstream bucket-key equality. Component
+// names are preserved verbatim, so duplicate dimension names would
+// collide; the schema rejects duplicates at validation time so this is
+// not a real concern.
+func (k BucketKey) AsMap() map[string]string {
+	if k.IsGlobal() {
+		return nil
+	}
+	out := make(map[string]string, len(k.Components))
+	for _, c := range k.Components {
+		out[c.Name] = c.Value
+	}
+	return out
+}
+
 // String renders a stable, canonical string suitable for use as a Redis
 // sub-key. The format escapes the field separator ('|') and the
 // name-value separator ('=') in values to avoid collisions, e.g.
--- a/internal/httpf/factory_ordering_test.go
+++ b/internal/httpf/factory_ordering_test.go
@@ -1,0 +1,133 @@
+package httpf
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// recordingFactory wraps every constructed RoundTripper so we can record
+// the order in which each layer runs against an actual request. Used by
+// the middleware-ordering test below.
+type recordingFactory struct {
+	name      string
+	recorder  *orderRecorder
+	skipBuild bool // when true, NewRoundTripper returns nil — exercises the if-nil branch
+}
+
+type orderRecorder struct {
+	mu    sync.Mutex
+	calls []string
+}
+
+func (r *orderRecorder) record(name string) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	r.calls = append(r.calls, name)
+}
+
+func (r *orderRecorder) snapshot() []string {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	out := make([]string, len(r.calls))
+	copy(out, r.calls)
+	return out
+}
+
+func (rf *recordingFactory) NewRoundTripper(_ RequestInfo, transport http.RoundTripper) http.RoundTripper {
+	if rf.skipBuild {
+		return nil
+	}
+	return &recordingRT{name: rf.name, recorder: rf.recorder, transport: transport}
+}
+
+type recordingRT struct {
+	name      string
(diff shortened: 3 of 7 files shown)


internal/httpf/factory.go
```diff
@@ -36,14 +36,21 @@ func CreateFactory(
...
+	// short-circuit with a synth 429 — and so it can install the
+	// per-request Attribution onto the context before any inner middleware
+	// reads or writes it.
+	var middlewares []RoundTripperFactory
 	for _, m := range additionalMiddlewares {
 		if m != nil {
 			middlewares = append(middlewares, m)
 		}
 	}
+	middlewares = append(middlewares, requestLog)
```
rmorlok, line 53: Is this a change on the PR or was this a bug from previous work? If so, is there a gap here about testing expectations that should have covered this?

internal/ratelimit/bucket_key.go
```diff
@@ -29,6 +29,23 @@ func (k BucketKey) IsGlobal() bool {
 	return len(k.Components) == 0
 }
 
+// AsMap renders the bucket key as a map suitable for storage on a request
+// log entry (LogRecord.RateLimitBucket). Order is lost — the map is for
+// human consumption, not for downstream bucket-key equality. Component
+// names are preserved verbatim, so duplicate dimension names would
+// collide; the schema rejects duplicates at validation time so this is
+// not a real concern.
+func (k BucketKey) AsMap() map[string]string {
```
rmorlok, line 38: Please add a test for this.

internal/httpf/factory.go
rmorlok: Pre-existing bug. The comment in `factory.go` claimed `request_log` was outermost, but the loop made it innermost — so the existing connector-level reactive limiter's synthetic-429 attribution (added in #222) silently did nothing in production because the Attribution pointer was never on the context by the time the limiter ran.

Testing gap acknowledged: there was no test pinning the middleware execution order, so the discrepancy between comment and code was undetectable. Added `internal/httpf/factory_ordering_test.go` (commit ee07d8c) with two tests that exercise the chain end-to-end via gentleman + an httptest upstream and pin the contract: `requestLog → additionalMiddlewares (last-in-slice first) → upstream`. Future ordering changes will fail this test loudly rather than degrade silently.

internal/ratelimit/bucket_key.go
rmorlok: Added in commit ee07d8c. Three sub-tests in `bucket_key_test.go`: global key returns nil (so JSON serialization produces `null` / omits the field rather than `{}`), populated key round-trips to the expected map, and empty values are preserved (so log filtering can distinguish unauthenticated buckets like `actor=""` from populated ones).

Issue #255, opened 2026-05-10
[api-key] Probe-driven health: counters, health_state, transitions


rmorlok:
Part of #243 — see that issue, Part 5.

## Scope

Generalized probe-driven health checks across all auth types (OAuth2 and api-key — this is the unified reauth signal called for in #243).

Currently `service.runProbeForConnection` (`internal/core/task_probe.go:43-92`) only logs probe outcomes; periodic-probe failures do not affect connection state. This sub-issue closes that gap.

### Storage

- Per-probe counters: `consecutive_failures` and `consecutive_successes`. Storage choice during implementation: denormalized on the connection (e.g. JSON map keyed by probe id) vs. a small `connection_probe_health` table. Either is acceptable.
- Connection-level `health_state` enum: `healthy` | `unhealthy`. **Distinct from `ConnectionState`** (lifecycle: `created` / `ready` / `disabled` / `disconnecting` / `disconnected`). A connection can be `Ready` and `unhealthy` simultaneously.

### Execution

Extend `service.runProbeForConnection`:

- On success: increment success counter for that probe; reset failure counter for that probe.
- On failure: increment failure counter for that probe; reset success counter.
- On `consecutive_failures >= failure_threshold` for any probe: flip `health_state` to `unhealthy`.
- On `consecutive_successes >= recovery_threshold` (for the previously failing probes) while `health_state == unhealthy`: flip back to `healthy` and reset all counters for the connection.
- Probes continue running while unhealthy — recovery is automatic.
- `ConnectionState` is unaffected by probe outcomes.

### Events

Emit a structured event on each `healthy ↔ unhealthy` transition, consistent with the OAuth2 callback failure event pattern from #228.

### `last_validated_at`

On probe success against an api-key connection, update `api_key_credentials.last_validated_at` for the active credential row.

## Test plan

- [ ] Single failure does not flip state
- [ ] N consecutive failures (where N == `failure_threshold`) flip to `unhealthy`
- [ ] `recovery_threshold` consecutive successes flip back to `healthy`; counters reset
- [ ] Mixed probe outcomes (one passing, one failing): `unhealthy` if any probe is over its threshold
- [ ] Cross-auth: same logic applies for an OAuth2 connection
- [ ] Threshold-cross emits the structured event
- [ ] api-key probe success updates `last_validated_at`

## Dependencies

- Probe threshold schema sub-issue
- Proxy injection sub-issue (so `proxy_http` probes work for api-key, not just OAuth2)

[...]

Issue #253, opened 2026-05-10
[api-key] internal/auth_methods/api_key proxy package


rmorlok:
Part of #243 — see that issue for full project context.

## Scope

New `internal/auth_methods/api_key/` package implementing `iface.Proxy`, mirroring the structure of `internal/auth_methods/oauth2/` and `internal/auth_methods/no_auth/`.

### Behavior

1. Load the active `api_key_credentials` row for the connection.
2. Decrypt via the existing encryption service.
3. Inject per `placement_snapshot`:
   - `bearer` → `Authorization: Bearer <key>`
   - `header` → `<HeaderName>: <Prefix><key>`
   - `query` → URL-encode `<ParamName>=<key>` and merge into the request URL (preserving existing query parameters)
   - `basic` → `Authorization: Basic <base64(username:key)>` (system performs the encoding; user is never asked for the encoded form)

### Wire-up

Extend `internal/core/connection_proxy.go:18-43` with the api-key dispatch branch:

```go
if _, ok := auth.Inner().(*config.AuthApiKey); ok {
    return apiKeyFactory.NewApiKey(c)
}
```

### Side benefit

`proxy_http` probes against api-key connectors begin working (today they hit `ErrProxyNotImplemented`). This is a prerequisite for the probe-runtime sub-issue to actually validate api-key credential health.

## Test plan

- [ ] Each placement produces the correct `http.Request` (header values, query string, encoding)
- [ ] `basic` correctly base64-encodes `username:key`
- [ ] `query` placement preserves existing query parameters when merging
- [ ] `header` placement honors `prefix` (with and without trailing space)
- [ ] `proxy_http` probe runs end-to-end against a stub upstream

## Dependencies

- AuthApiKey schema sub-issue
- `api_key_credentials` storage sub-issue

Issue #252, opened 2026-05-10
[api-key] Connection initiate: auto-form + persist credentials


rmorlok:
Part of #243 — see that issue for full project context.

## Scope

Wire api-key into the connection initiate flow.

### Auto-form generation

At config-load time, when a connector declares `auth.type: api-key` without an explicit `setup_flow.preconnect` step, synthesize a `SetupFlowStep` collecting:

- `bearer` / `header` / `query`: single `api_key` (rendered as a secret input).
- `basic`: `<username_field>` + `api_key` (key rendered as a secret).

Defaults for labels and help text. Produces a normal `SetupFlowStep` indistinguishable from a hand-written one.

An explicit `setup_flow.preconnect` in YAML suppresses synthesis. The connector author owns ensuring the step yields the credential field names the proxy injection layer expects.

### Initiate dispatch

Extend `service_connection_initiate.go:87-131` to handle `AuthApiKey` and present the credential-collection step (auto-generated or explicit).

### Submit handler

On step submit:
1. Validate against the step's `json_schema`.
2. Encrypt and persist into `api_key_credentials` (single active row + soft-delete of any prior).
3. Advance to `SetupPhaseVerify`.

`http` (raw) probes execute via the existing path. `proxy_http` probes will return `ErrProxyNotImplemented` until the proxy sub-issue lands — note that limitation in tests.

### Greenhouse stub cleanup

Update or remove the `test_data/system.yaml:58` greenhouse stub now that api-key has a real schema.

## Test plan

- [ ] Auto-form generation produces correct `SetupFlowStep` for each placement
- [ ] Explicit `setup_flow.preconnect` suppresses synthesis
- [ ] Submit persists encrypted credentials and advances to verify
- [ ] Submit's "no prior credential decrypted into form" invariant holds (foundation for the reauth sub-issue)
- [ ] Greenhouse YAML loads without error after cleanup

## Dependencies

- AuthApiKey schema sub-issue
- `api_key_credentials` storage sub-issue

rmorlok:
Note from #243 / #261 work: it is **fine to break the greenhouse stub** that exists today (`auth.type: api-key` with no `placement`) when this work lands. The stub appears in:

- `dev_config/default.yaml`
- `test_data/system.yaml`
- `test_data/other-system.yaml`
- `internal/schema/config/test_data/valid.yaml`
- `internal/schema/config/test_data/invalid-bad-service-port.yaml`
[...]

Issue #243, opened 2026-05-09
Add support for API Key Integrations


rmorlok:
## Goals (original brief)

Today, AuthProxy supports no-auth and OAuth2 connector types. The goal of this project is to add support for API key authenticated 3rd parties.

Authproxy should support:
- Header-based (most common for modern APIs):
  - `Authorization: Bearer <key>` (SendGrid, OpenAI, Mapbox, etc.)
- Custom headers (AWS, Azure)
- Query param (older/simple APIs):
  - `?api_key=...`, `?appid=...`, `?key=...`
- Basic Auth (less common but still used):
  - Encode key with username (api or account ID)

The application should support API keys using the current connection initiate flow where the needed values are requested as fields to the user. It should fit into the normal pre-auth, auth, configure flow. If APIs require an encoding of the API key (e.g. base64 from what the user would know), that should be configurable in terms of how the values are rendered.

The system should take into account how reauth happens in this situation, avoiding replay of the credentials. The API keys should be an encrypted form within the database. This is likely already accounted for in the current config data storage mechanism we have.

---

## Background — what exists today

- `AuthType` is declared with `OAuth2`, `api-key`, and `no-auth` variants in `internal/schema/connectors/auth.go:12-18`. The OAuth2 path is fully implemented; `AuthApiKey` is a 21-line stub with no fields and no proxy implementation.
- `service_connection_initiate.go:87-131` returns `unsupported connector auth type` when it encounters an api-key connector.
- The setup-flow framework (`internal/schema/connectors/setup_flow.go:440-493`) is fully YAML-driven (json_schema + ui_schema). Dynamic user-input collection is largely already solved.
- `EncryptedConfiguration` on the connection (AES-GCM, generic blob) plus the `oauth2_tokens` table (encrypted columns + soft-delete on rotation) are the two existing patterns for credential storage. The encryption service (`internal/encrypt/service.go`) is generic.
- Probes already exist (`internal/schema/connectors/probe.go`, `internal/core/task_probe.go`) and run in two modes: one-shot during `SetupPhaseVerify`, plus Asynq-scheduled periodic execution. **However, periodic-probe failures are currently logged only — they do not affect connection state, and there is no consecutive-failure counter.** This is the gap the reauth requirements below close.
[...]

PR #266, opened 2026-05-11
Add api_key_credentials table with encrypted storage (#251)

### LINKED ISSUE #251: [api-key] api_key_credentials table and encrypted storage

## Scope

Add a new `api_key_credentials` table parallel to `oauth2_tokens`. Mirrors the OAuth2 credential pattern: secrets live separate from the user-form `EncryptedConfiguration` blob, with a clear lifecycle and audit trail.

### Schema

See #243 Part 3 for the illustrative columns. Highlights:

- `encrypted_api_key` and `encrypted_username` use `encfield.EncryptedField` (existing AES-GCM, no new crypto).
- Encryption registration mirrors `internal/database/oauth2_token.go:21-30`.
- `placement_snapshot` (jsonb) captures the placement config at submission time so YAML changes don't break in-flight credentials.
- Audit columns: `created_at`, `last_rotated_at`, `rotated_by_actor_id`, `last_validated_at`, `deleted_at`.

### Operations

- `Insert(connection_id, encrypted_api_key, encrypted_username, placement_snapshot)` — also soft-deletes the prior active row for that connection in the same transaction.
- `GetActiveForConnection(connection_id) → row | not found`.
- `UpdateLastValidated(id, ts)` — called by probes on success (wired up in the probe-runtime sub-issue).

## Out of scope

- The form that collects credentials (initiate-flow sub-issue).
- The proxy that decrypts and injects (proxy sub-issue).
- "No replay" UI behavior (reauth sub-issue).

## Test plan

- [ ] Migration up and down
- [ ] `Insert` + `GetActiveForConnection` returns the latest row
- [ ] `Insert` soft-deletes prior active row in the same transaction (single active invariant)
- [ ] Encryption registration: re-encryption task touches both encrypted columns

## Dependencies

None — foundational PR. Can land in parallel with the schema sub-issue.

rmorlok:
Closes #251. Part of #243.

## Summary
- New \`api_key_credentials\` table (parallel SQLite + Postgres migrations \`000005_create_api_key_credentials\`).
- New \`ApiKeyCredential\` model in \`internal/database/api_key_credential.go\` with the standard cols/fields/values pattern.
- Operations: \`InsertApiKeyCredential\` (soft-deletes any prior active row in the same transaction → single-active invariant), \`GetActiveApiKeyCredential\`, \`UpdateApiKeyCredentialLastValidated\`, \`DeleteAllApiKeyCredentialsForConnection\`.
- Encrypted columns \`encrypted_api_key\` and \`encrypted_username\` (the latter populated only for the \`basic\` placement) reuse the existing \`encfield.EncryptedField\` AES-GCM type. Registered with the re-encryption registry via JOIN on \`connections\` for namespace resolution — same pattern as \`oauth2_tokens\`.
- \`placement_snapshot\` (jsonb on Postgres, text on SQLite) freezes the connector's \`ApiKeyPlacement\` config at submission time so an in-flight YAML change can't break previously-submitted credentials. New \`apiKeyPlacementDB\` Valuer/Scanner mirrors the \`rateLimitDefDB\` pattern.
- New \`PrefixApiKeyCredential = \"akc_\"\` apid prefix, registered in \`validPrefixes\`.
- Mocks regenerated via \`go generate ./internal/database/...\`.

## Schema deviation from #251
The requirements doc lists \`last_rotated_at\` and \`rotated_by_actor_id\`. With the insert-new-row + soft-delete-prior model adopted in this PR, each row's own \`created_at\` already records when it became active, so \`last_rotated_at\` would just duplicate it. \`rotated_by_actor_id\` is renamed to \`created_by_actor_id\` to match per-row semantics — it captures whoever submitted (initial) or rotated to (subsequent) the credential. The \"last rotation\" event is the \`created_at\` of the active row.

## What's intentionally NOT in this PR
- No call sites — the connection-initiate flow (#252) is what calls \`InsertApiKeyCredential\`, the proxy package (#253) is what reads via \`GetActiveApiKeyCredential\`, and probes (#255) are what call \`UpdateApiKeyCredentialLastValidated\`. This PR is foundation only.

## Test plan
- [x] \`Validate()\` covers required fields + prefix checks (5 subtests).
- [x] Round-trip insert + GetActive returns the expected row with placement_snapshot deep-copied.
- [x] Two inserts on one connection: prior row soft-deleted, exactly one active row remains, GetActive returns the newer one.
- [x] Inserts on sibling connections don't interfere.
- [x] Basic placement with encrypted username persists and round-trips.
- [x] \`UpdateLastValidated\` writes the column and surfaces \`ErrNotFound\` for missing ids.
- [x] \`DeleteAllForConnection\` soft-deletes only the target connection's rows; sibling connection untouched.
- [x] Encrypted-field registration smoke test asserts \`encrypted_api_key\` + \`encrypted_username\` are both registered with the right JOIN config.
- [x] Each of the four placement variants round-trips through \`placement_snapshot\`.
- [x] \`go test ./...\` green against PostgreSQL (\`AUTH_PROXY_TEST_DATABASE_PROVIDER=postgres\`).
- [x] \`go test ./internal/database/ -run TestApiKeyCredential\` green against SQLite.


internal/database/migrations/postgres/000005_create_api_key_credentials.up.sql
```diff
@@ -0,0 +1,19 @@
+create table api_key_credentials
+(
+    id                    text primary key,
+    connection_id         text not null,
+    encrypted_api_key     jsonb,
+    encrypted_username    jsonb,
```
rmorlok: Instead of having two encrypted columns here, let's only have one for the encrypted credentials that may have substructure to include the username versus the API key. That will keep re-encryption jobs simpler as there will only be one column to handle.

rmorlok: Collapsed into a single `encrypted_credentials` column in 400ad59. The api-key / username substructure now lives inside the encrypted blob — the database is agnostic to it, and re-encryption jobs only touch one column per row. `InsertApiKeyCredential` drops its `encryptedUsername` parameter; the auth layer in #253 will own marshaling `{api_key, username?}` into the plaintext that gets encrypted.

Issue #239, opened 2026-05-09
OTel #10: Dev sample — grafana/otel-lgtm in docker-compose


rmorlok:
Part of #225. Depends on #230, #231, #232, and ideally #238 (so something visible in all three signals end-to-end).

Provide a zero-setup local observability sample so contributors can see AuthProxy traces, metrics, and logs in Grafana with one command.

## Scope

- Add a `grafana/otel-lgtm` (Grafana + Tempo + Loki + Prometheus + bundled OTel Collector) container to `docker-compose.yml` under a new `observability` profile. Launch via `docker compose --profile observability up -d`.
- Pre-provision Grafana datasources and at least:
  - One dashboard demonstrating proxy RED metrics (rate / errors / duration), with a panel filtered by an allowlisted connector label such as `type`.
  - Trace search docs / saved query showing how to find proxy spans by connector or by `connection_id`.
- Default `dev_config/default.yaml` adds a `telemetry:` block pointing at the sample's OTLP endpoint, gated on a flag or env var so it only activates if the sample is running. Default behavior in dev (no `--profile observability`) remains unchanged.

## Acceptance criteria

- `docker compose --profile observability up -d` brings the stack up cleanly. Grafana is reachable on a documented port.
- Running AuthProxy locally (with telemetry enabled) populates the Grafana dashboard within seconds of the first traffic.
- `docker compose down` (or `./scripts/teardown-docker.sh`) tears it down cleanly.
- `./scripts/preflight.sh` passes.

## Out of scope

- Production-grade dashboards or alerting.
- Anything beyond the bundled all-in-one image (no separate Tempo / Prom containers).

Issue #225, opened 2026-05-08
Add OpenTelemetry Support


rmorlok:
# Add OpenTelemetry Support

Add comprehensive OpenTelemetry instrumentation to AuthProxy across all four services (`api`, `admin-api`, `public`, `worker`), with first-class configuration, a dev-friendly Grafana sample, and a tightly-bounded label projection model for proxy traffic.

## Delivery breakdown

This project is delivered as the following sub-tickets, in order. Each is sized as a single PR. All carry the `project:opentelemetry` label.

- [x] #230 — OTel #1: Foundation — telemetry config block, schema, and SDK bootstrap (no-op when disabled)
- [x] #231 — OTel #2: Inbound HTTP instrumentation (Gin server spans, metrics, error capture)
- [x] #232 — OTel #3: Outbound proxy instrumentation + label allowlist projection
- [x] #233 — OTel #4: Trace context propagation (opt-in, per-connector override)
- [x] #234 — OTel #5: Database (Postgres / SQL) instrumentation
- [x] #235 — OTel #6: Redis instrumentation
- [x] #236 — OTel #7: Asynq worker + scheduler instrumentation
- [x] #237 — OTel #8: OAuth2 lifecycle telemetry
- [x] #238 — OTel #9: Logs bridge — slog → OTel + trace_id/span_id correlation
- [x] #239 — OTel #10: Dev sample — grafana/otel-lgtm in docker-compose
- [x] #240 — OTel #11: Documentation — docs/telemetry.md, README TOC, AGENTS.md, docs index

Dependencies: #230 must land first (everything depends on it). #231–#238 are largely parallelizable after #230, with the noted soft dependencies in each ticket. #239 needs at least #230/#231/#232 (and ideally #238) so all signals are visible. #240 lands last so docs match shipped behavior.

## Goals

- Emit OTel **traces, metrics, and logs** from every AuthProxy service.
- Capture **errors/exceptions** as span events wherever practical.
- Provide **configurable label projection** from incoming proxy requests onto telemetry, with cardinality controls.
- Ship a **Grafana-based dev sample** in `docker-compose.yml` so contributors can see telemetry locally with zero setup.
- Be **off by default** so existing deployments are unaffected by the upgrade.

## Signals

All three OTel signals are in scope:

- **Traces** — server spans, client spans, task spans, DB/Redis spans.
- **Metrics** — HTTP server RED metrics, proxy-specific metrics, worker queue metrics, auth lifecycle counters.
[...]

PR #270, opened 2026-05-11
OpenTelemetry — Step 5: database (SQL) instrumentation (#234)

### LINKED ISSUE #234: OTel #5: Database (Postgres / SQL) instrumentation
Part of #225. Depends on #230.

Instrument `database/sql` access with OTel spans and metrics.

## Scope

- Instrument the `database/sql` driver registration (e.g., via `otelsql` from contrib, or equivalent wrapping).
- Per-query spans (operation, table where derivable, duration).
- Per-query metrics (count, duration histogram, optional connection-pool gauges if cheap).
- Applies to both Postgres (production / dev) and SQLite (in-memory test DB) so tests behave consistently.
- Honor `telemetry.signals.traces` / `telemetry.signals.metrics` toggles.

## Acceptance criteria

- Unit / integration test asserts that a query made via the instrumented DB produces a span and a metric record (using an in-memory exporter).
- When telemetry is disabled (#230 default), DB calls go through with no overhead.
- Existing DB tests (`go test ./internal/database`) continue to pass unchanged.

## Out of scope

- Query parameter capture (PII risk).
- Slow-query detection / alerting beyond histogram emission.

rmorlok:
## Summary

Adds per-query OTel spans + metrics for the database/sql layer (Postgres + SQLite), wired in via a new `database.WithTelemetry` option on the existing constructors.

- Uses [XSAM/otelsql](https://github.com/XSAM/otelsql) v0.41.0 (latest line that targets Go 1.24, paired with OTel v1.39+) to wrap \`sql.Open\` when telemetry is enabled. Drop-in for both Postgres (\`pgx\`) and SQLite (\`go-sqlite3\`), so dev / test / prod all emit the same shape.
- Reports semconv \`db.system\` per driver (\`postgresql\` / \`sqlite\`).
- When the metrics signal is on, additionally registers \`otelsql.RegisterDBStatsMetrics\` so the standard go-sql connection-pool gauges (open / in-use / idle / wait counts) show up. Best-effort: registration failures don't fail the connection.
- \`database.WithTelemetry\` is a non-breaking functional option — existing callers (notably \`test_db.go\`) keep their current signature and stay uninstrumented.
- When telemetry is disabled, providers are no-op, or both trace and metric signals are off, the constructor falls through to plain \`sql.Open\` — zero overhead for unconfigured deployments.

## Test plan

- [x] \`go test ./internal/database/...\` — new tests cover:
  - Query through instrumented DB emits at least one span and at least one metric stream
  - Default-off: with no \`WithTelemetry\` option, no spans / metrics emitted
  - Providers in no-op mode + telemetry config absent: no spans / metrics
  - Both signals off (traces=false, metrics=false): no spans / metrics
- [x] Existing internal/database tests continue to pass unchanged
- [x] \`./scripts/preflight.sh\` passes (including integration_tests \`go list -mod=readonly\`)
- [x] \`go test ./...\` clean across the whole tree
- [x] \`golangci-lint run\` clean
- [ ] Manual verification with a live collector (deferred to #239 dev sample)

Closes #234. Part of the OpenTelemetry project — parent #225.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	cloud.google.com/go/secretmanager v1.16.0
 	github.com/ClickHouse/clickhouse-go/v2 v2.43.0
 	github.com/Masterminds/squirrel v1.5.4
+	github.com/XSAM/otelsql v0.41.0
 	github.com/alicebob/miniredis/v2 v2.35.0
 	github.com/aws/aws-sdk-go-v2 v1.41.2
 	github.com/aws/aws-sdk-go-v2/config v1.32.9
--- a/go.sum
+++ b/go.sum
@@ -27,6 +27,8 @@ github.com/PuerkitoBio/purell v1.1.1 h1:WEQqlqaGbrPkxLJWfBwQmfEAE1Z7ONdDLqrN38tN
 github.com/PuerkitoBio/purell v1.1.1/go.mod h1:c11w/QuzBsJSee3cPx9rAFu61PvFxuPbtSwDGJws/X0=
 github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578 h1:d+Bc7a5rLufV/sSk/8dngufqelfh6jnri85riMAaF/M=
 github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578/go.mod h1:uGdkoq3SwY9Y+13GIhn11/XLaGBb4BfwItxLd5jeuXE=
+github.com/XSAM/otelsql v0.41.0 h1:uZifjQhZhv5EDYJh+IVk1DiYxQZJBlNSen0MBFnfxB8=
+github.com/XSAM/otelsql v0.41.0/go.mod h1:NMQT0PiKoFILp9QgjQz+D5mvW+9mT0suR7OejqrtMaM=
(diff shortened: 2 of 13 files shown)


rmorlok (commented):
The `internal/request_log` has a separate postgres/sqlite/clickhouse database setup. You missed instrumenting there.

rmorlok:
Addressed in 3926ab0:

- New \`database.OpenInstrumentedSQL(driverName, dsn, dbSystem, opts...)\` and \`database.OpenInstrumentedConnector(connector, dbSystem, opts...)\` for the connector-based ClickHouse path (\`clickhouse.Connector(opts)\` → \`otelsql.OpenDB(connector, ...)\`)
- \`request_log\` factories take \`...database.Option\` and forward to all four DB call sites:
  - \`NewSqlRecordStore\` / \`NewSqlRecordRetriever\` (Postgres + SQLite)
  - \`NewClickhouseRecordStore\` / \`NewClickhouseRecordRetriever\`
- \`NewStorageService\` accepts \`...database.Option\` and \`DependencyManager.GetLogStorageService\` passes \`database.WithTelemetry(...)\` through
- Tagged with the appropriate semconv \`db.system\` per provider (\`postgresql\` / \`sqlite\` / \`clickhouse\`)
- New test in \`internal/request_log/telemetry_test.go\` pins span + metric emission when telemetry is on, and confirms the no-op fast path when it isn't (otelsql only spans Ping if its SpanOptions flag is set, so the probe issues a real \`SELECT 1\` through QueryContext)

PR #273, opened 2026-05-11
Docs: rate limits + labels guides, docs/ index, main README TOC

rmorlok:
## Summary

Now that the eight #3 PRs are merged and rate-limit support is live end-to-end, this adds customer-facing documentation.

## What's in

**\`docs/rate-limits.md\` (~360 lines)**
Opens with a customer guide:
- When to use the connector-level reactive 429 handler vs. proxy-side rate-limit resources.
- Defining a rate limit via API and via Terraform (with side-by-side examples).
- Selectors — label selector, methods, path match (prefix / glob / regex), request types (default \`[proxy, probe]\` + opt-in for OAuth2 sub-types).
- Buckets — reserved dimensions (\`actor\`, \`connection\`, \`connector\`, etc.) and \`labels/<key>\` references.
- Algorithms — fixed_window, sliding_window (log + counter), token_bucket, with the relevant tradeoffs.
- \`enforce\` vs. \`observe\` mode, including why observe still increments counters.
- What the caller sees on rejection — 429, \`Retry-After\`, \`X-Authproxy-Ratelimited\`, \`X-Authproxy-Ratelimit\`, body shape.
- Multi-rule "all apply, most-restrictive wins" semantics.
- Namespace inheritance and permissions.
- Connector-level reactive 429 handling — full options table including the \`exponential_backoff\` sub-options.
- Request log attribution — \`response_source\` enum, \`rate_limit_id\`/\`mode\`/\`bucket\`/\`matched\` fields, list-endpoint filters.

Continues with an implementation reference for operators:
- Architecture diagram (cache → matcher → limiter → Redis).
- Middleware chain order (request log → telemetry → enforcer → reactive 429 → upstream).
- Bucket key canonical serialization rules.
- Per-algorithm internals (Lua scripts, state shape, retry-after computation).
- Fail-open semantics on Redis outage.
- Coexistence with the reactive limiter (which fires first, what wins on attribution).
- Performance characteristics.
- Known limitations (cross-process invalidation on admin writes deferred, etc.).

**\`docs/labels.md\` (~190 lines)**
Companion page covering the label system end to end:
- Where labels live across resource types.
- Label vs. annotation tradeoffs.
- The \`apxy/\` system label namespace — identifier labels (\`apxy/<rt>/-/id\`, \`apxy/<rt>/-/ns\`) and the resource-type token table.
- Carry-forward through the hierarchy with a worked three-level example showing exactly what ends up on a connection.
- Propagation on parent change (targeted refresh + daily reconciler).
- The per-request label snapshot — what's in it and how the actor contribution gets re-keyed.
- Label selector syntax with examples for both user and system labels.
- The four-endpoint label/annotation API surface every resource exposes.

**\`docs/README.md\`**
Index for the \`docs/\` directory grouped into feature guides, operational guides, reference, and architecture diagrams. Renders on GitHub as the directory landing page.

**\`README.md\` updates**
- Adds a top-level Table of Contents linking to existing sections via anchors.
- Adds a "Feature documentation" section near the top with links to the new docs.
- Adds a "Rate Limits" sub-section under Key Concepts that introduces the feature and points at \`docs/rate-limits.md\`.
- Expands the existing "Labels and Annotations" section with a pointer to \`docs/labels.md\`.
- Reorganises the bottom "Additional Documentation" section into a full "Documentation" table linking feature guides + operational guides + Terraform provider docs / examples.

## Notes for the reviewer

- The two long docs are **opinionated**: customer concerns first, internals near the end so an integrator can stop reading at the point that's relevant to them.
- Every API path, header name, default value, and field name was verified against the code on \`main\` (e.g., \`X-Authproxy-Ratelimited\` / \`X-Authproxy-Ratelimit\`, the \`rate_limit_matched\` field shape, the connector-level \`exponential_backoff\` defaults).
- Cross-links between \`labels.md\` and \`rate-limits.md\` use relative paths so they work both on GitHub and locally.
- I left the existing \`docs/index.md\` alone — it looked like it serves a different purpose (the GitHub Pages landing rather than the docs directory's GitHub-rendered README).

## Test plan

- [x] \`./scripts/preflight.sh\` — clean (docs-only change but worth confirming nothing slipped)
- [x] Spot-checked every relative link target — all files / sections exist.
- [x] Main README TOC anchors verified against actual heading text.

### DIFF
--- a/docs/labels.md
+++ b/docs/labels.md
@@ -58,7 +58,6 @@ Every namespace-scoped resource carries `labels` and `annotations` columns:
 | `connection` | Inherits labels from its connector version, plus the namespace. |
 | `encryption_key` | Inherits from its namespace. |
 | `rate_limit` | Inherits from its namespace. See [Rate limits](rate-limits.md). |
-| `oauth2_token` | Inherits from its connection. |
 | `request_log` entry | The frozen label snapshot of the request that produced it. |
 
 Request log entries store the per-request label snapshot (the values that were active when the request ran), not a live reference — so a label change on the connection does not retroactively change old log entries.
@@ -100,7 +99,6 @@ The `<rt>` token comes from the resource's apid prefix (strip the trailing `_`):
 | Connector version | `cxr` |
 | Encryption key | `ek` |
 | Rate limit | `rl` |
-| OAuth2 token | `tok` |
 
 For example, a connection with id `cxn_abc123` in namespace `root.acme` always carries:
 
@@ -125,7 +123,6 @@ namespace
    ├── encryption_key                   (labels carry forward through "ns/")
    ├── connector (version)              (labels carry forward through "cxr/")
    │      └── connection                (labels carry forward through "cxn/")
-   │             └── oauth2_token       (labels carry forward through "tok/")
    ├── rate_limit                       (labels carry forward through "ns/")
    └── (child namespace)                (labels carry forward through "ns/")
 ```
@@ -160,14 +157,18 @@ apxy/cxn/-/ns:          root.acme        # implicit identifier
 
 ### Propagation on parent change
 
-Carry-forward is **materialised on create**. If the parent's labels change later, the values on existing children are stale until the propagation job runs.
+Carry-forward is **materialised on create** — and **eventually consistent** afterwards. If the parent's labels change later, the values on existing children are stale until the propagation job catches up; depending on fleet size, fan-out depth, and how rate-limited the reconciler is, that delay can be anywhere from a few seconds to several hours.
 
 There are two propagation paths:
 
-1. **Targeted refresh.** Admin API mutators (e.g., updating a namespace's labels) enqueue a background task that re-derives every descendant's `apxy/` mirror — running each row's update in its own short transaction so concurrent reads aren't blocked.
-2. **Daily consistency checker.** A scheduled job walks every resource type, compares the on-disk labels to what the carry-forward rule would compute, and corrects any drift. Belt-and-braces for the targeted refresh.
+1. **Targeted refresh.** Admin API mutators (e.g., updating a namespace's labels) enqueue a background task that re-derives every descendant's `apxy/` mirror — running each row's update in its own short transaction so concurrent reads aren't blocked. Typical latency: seconds to a few minutes, longer for deep fan-outs.
+2. **Daily consistency checker.** A scheduled job walks every resource type, compares the on-disk labels to what the carry-forward rule would compute, and corrects any drift. Belt-and-braces for the targeted refresh; the worst-case bound is "next daily run plus the time the walk itself takes" — minutes to hours.
 
-Operational implication: a label change on a deep namespace can fan out to many descendants. The propagation job batches updates and runs out of band; do not rely on a label change being visible to in-flight proxy requests within the next few seconds.
+**Operational implications:**
+
+- Treat carry-forward labels as eventually-consistent reads. Application logic that needs to observe a fresh parent-label change should not rely on a child's mirrored `apxy/<parent>/...` having caught up yet.
+- A label change on a deep namespace can fan out to many descendants. The propagation job batches updates and runs out of band; do not rely on a label change being visible to in-flight proxy requests within the next few seconds.
+- Per-request label snapshots on **new** requests after the update reflect the parent change as soon as each individual descendant row is rewritten — so propagation progresses visibly through the fleet rather than landing atomically.
 
 ## Per-request label snapshot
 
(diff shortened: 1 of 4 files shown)


docs/labels.md
```diff
@@ -0,0 +1,234 @@
...
+
+| Resource | Notes |
+|---|---|
+| `namespace` | Top of the hierarchy. Its labels carry forward to every resource defined in it. |
+| `actor` | Users / service accounts. Labels carry forward to requests an actor initiates. |
+| `connector` (versioned) | Each version is a separate resource with its own labels. |
+| `connection` | Inherits labels from its connector version, plus the namespace. |
+| `encryption_key` | Inherits from its namespace. |
+| `rate_limit` | Inherits from its namespace. See [Rate limits](rate-limits.md). |
+| `oauth2_token` | Inherits from its connection. |
```
rmorlok: OAuth2 tokens don't have labels or annotations defined against them.

docs/labels.md
```diff
@@ -0,0 +1,234 @@
...
+
+| Resource | `<rt>` |
+|---|---|
+| Namespace | `ns` |
+| Actor | `act` |
+| Connection | `cxn` |
+| Connector version | `cxr` |
+| Encryption key | `ek` |
+| Rate limit | `rl` |
+| OAuth2 token | `tok` |
```
rmorlok: As noted, these don't carry labels, so let's remove references to oauth2 tokens from this doc.

docs/labels.md
```diff
@@ -0,0 +1,234 @@
...
+app.example.com/cohort: beta             # user-set
+apxy/ns/team:           alpha            # carried from parent namespace
+apxy/ns/env:            prod             # carried from root (transitively)
+apxy/cxn/-/id:          cxn_…            # implicit identifier
+apxy/cxn/-/ns:          root.acme        # implicit identifier
+```
+
+### Propagation on parent change
+
+Carry-forward is **materialised on create**. If the parent's labels change later, the values on existing children are stale until the propagation job runs.
```
rmorlok: Let's use the term eventually consistent here. Note that propagation can take on the order of minutes to hours.

docs/rate-limits.md
```diff
@@ -0,0 +1,471 @@
...
+AuthProxy has two complementary rate-limiting systems:
+
+1. **Connector-level reactive rate limiting** — built into every connector. When a 3rd party returns 429, AuthProxy parses the `Retry-After` header and blocks subsequent requests on that connection until the wait expires. This protects you from hammering the upstream after it has already told you to slow down.
+2. **Rate limit resources** — declarative, namespace-scoped rules you define via API or Terraform. They run **before** the upstream call and reject requests that would breach a quota you have configured (e.g., "no more than 60 writes per minute per actor against Salesforce").
+
+Both can fire on the same request and both stamp the request log so you can tell them apart. This page covers each in detail.
+
+## Table of contents
+
+- [Customer guide](#customer-guide)
```
rmorlok: Let's use the term user guide instead of customer.

docs/rate-limits.md
```diff
@@ -0,0 +1,471 @@
...
+
+The response carries the server-assigned id (e.g., `rl_AbcXyz…`), the materialised label set (including the implicit `apxy/rl/-/id` / `apxy/rl/-/ns` and any carried-forward namespace labels — see [Labels](labels.md)), and timestamps.
+
+Update with `PATCH /api/v1/rate-limits/{id}` (the request body can carry any of `definition`, `labels`, `annotations`; omit fields you don't want to change). Delete with `DELETE /api/v1/rate-limits/{id}`. List with `GET /api/v1/rate-limits` (supports `namespace`, `label_selector`, and pagination cursor params).
+
+Labels and annotations also have sub-resource endpoints — see [Labels — API surface](labels.md#api-surface).
+
+### Via Terraform
+
+The `authproxy_rate_limit` resource exposes every field as typed HCL — no `jsonencode` blobs:
```
rmorlok: Take out "— no `jsonencode` blobs"

docs/labels.md
rmorlok: Fixed in commit a18686b — dropped the `oauth2_token` row from the resource table.

docs/labels.md
rmorlok: Fixed in commit a18686b — removed the `OAuth2 token | `tok`` row from the resource-type token table and dropped the `oauth2_token` line from the carry-forward hierarchy diagram.

docs/labels.md
rmorlok: Fixed in commit a18686b — reworked the propagation section to frame it as **eventually consistent** with an explicit note that latency can range from seconds to several hours depending on fleet size and how rate-limited the reconciler is. Also split out an operational-implications list so the consequences (treat as eventually-consistent reads, propagation is non-atomic across the fleet, etc.) read as a checklist.

docs/rate-limits.md
rmorlok: Done in commit a18686b — renamed both the section heading and its TOC entry to "User guide".

docs/rate-limits.md
rmorlok: Done in commit a18686b — dropped the phrase. The intro now reads simply "The `authproxy_rate_limit` resource exposes every field as typed HCL:".

PR #274, opened 2026-05-11
Wire api-key auth into the connection initiate + submit flow (#252)

### LINKED ISSUE #252: [api-key] Connection initiate: auto-form + persist credentials

rmorlok:
Closes #252. Part of #243.

## Summary
- \`Connector.Normalize()\` synthesizes a credential-collection preconnect step (\`api_key\` for bearer/header/query, \`api_key\` + the configured username field for basic) when the connector did not declare its own \`setup_flow.preconnect\`. Explicit preconnect suppresses synthesis. Called from \`Connector.Validate\` so the synthesized step is part of the canonical connector definition that gets hashed and versioned.
- \`service_connection_submit.go\` extracts the credential field values from submitted form data, marshals them as a typed \`ApiKeyCredentialPlaintext\`, encrypts via the existing encrypt service, and persists into \`api_key_credentials\`. Credential fields are stripped from the merged config before \`SetConfiguration\` so plaintext never reaches the per-connection \`EncryptedConfiguration\` blob.
- When \`NextSetupStep\` returns \`SetupStepAuth\` and the connector is \`AuthApiKey\`, the submit path elides the auth phase entirely and calls \`HandleCredentialsEstablished\` directly — the unified post-auth handler then advances to verify / configure / ready as appropriate.
- New \`ApiKeyCredentialPlaintext\` type in \`internal/database\` surfaces the canonical plaintext shape for both encrypters (this PR) and decrypters (api-key proxy, #253). The database itself does not interpret the bytes.

## Three deferred items from #261 land alongside
1. \`AuthValidator\` is now type-asserted in \`Connector.Validate\` — api-key connectors get full placement validation at config load.
2. \`schema.json\` marks \`placement\` required on \`AuthApiKey\` — the JSON-schema layer now rejects api-key blocks without a placement.
3. The greenhouse stub gets a real placement (\`bearer\`) in \`dev_config/default.yaml\`, \`test_data/system.yaml\`, \`test_data/other-system.yaml\`, \`internal/schema/config/test_data/valid.yaml\`, \`internal/schema/config/test_data/invalid-bad-service-port.yaml\`, and the inline YAML + expected struct in \`internal/schema/config/root_test.go\`.

The matching \`invalid-api-key-missing-placement.yaml\` fixture is re-added so the schema test exercises the new requirement.

## Test plan
- [x] Synthesis tests for each placement variant, idempotence, explicit-override suppression, and the no-op for non-api-key connectors.
- [x] \`TestApiKeySubmit_BearerPersistsCredentialAndAdvancesToReady\` — bearer submit persists the credential, strips \`api_key\` from the connection config, advances to ready.
- [x] \`TestApiKeySubmit_BasicPersistsBothCredentialFields\` — basic submit persists \`{api_key, username}\` together, strips both from the config blob.
- [x] \`TestApiKeySubmit_TransitionsToVerifyWhenProbesPresent\` — with probes configured, advances to verify (and enqueues the verify task).
- [x] \`TestApiKeySubmit_NoReplay_PriorCredentialNeverDecryptedIntoForm\` — submit path must NOT call \`GetActiveApiKeyCredential\`. Foundation for the reauth no-replay guarantee.
- [x] Full module green against both SQLite and PostgreSQL.

### DIFF
--- a/internal/core/service_connection_submit.go
+++ b/internal/core/service_connection_submit.go
@@ -13,8 +13,13 @@ import (
 )
 
 // SubmitForm handles form data submission for a connection setup flow step.
-// It merges the submitted data into the connection's configuration, advances to the next step,
-// and returns the appropriate response (next form, auth redirect, or complete).
+//
+// Dispatch is by phase:
+//   - preconnect / configure: validate, merge submitted fields into the
+//     connection's EncryptedConfiguration, advance.
+//   - credentials: validate, route to the auth method's credential-submit
+//     handler (which encrypts + persists into api_key_credentials), advance via
+//     HandleCredentialsEstablished. Field data never enters EncryptedConfiguration.
 func (c *connection) SubmitForm(ctx context.Context, req iface.SubmitConnectionRequest) (iface.ConnectionSetupResponse, error) {
 	setupStep := c.GetSetupStep()
 	if setupStep == nil {
@@ -28,7 +33,7 @@ func (c *connection) SubmitForm(ctx context.Context, req iface.SubmitConnectionR
 
 	currentSetupStep := *setupStep
 
-	// Only preconnect and configure phases accept form submissions
+	// Only the indexed phases (preconnect, credentials, configure) accept form submissions.
 	if !currentSetupStep.Phase().IsIndexed() {
 		return nil, httperr.BadRequestf("cannot submit form for phase %q", currentSetupStep.Phase())
 	}
@@ -39,30 +44,21 @@ func (c *connection) SubmitForm(ctx context.Context, req iface.SubmitConnectionR
 		return nil, httperr.InternalServerError(httperr.WithInternalErrorf("failed to get current step: %w", err))
 	}
 
-	// Get existing configuration to merge into
+	// Credential submissions are dispatched to the auth method — no merge into
+	// the general config blob, no risk of field-name collisions with preconnect.
+	if currentSetupStep.Phase() == cschema.SetupPhaseCredentials {
+		return c.submitCredentialsStep(ctx, req, currentStep, connector)
+	}
+
+	// Preconnect / configure submissions follow the generic merge path.
 	existingConfig, err := c.GetConfiguration(ctx)
 	if err != nil {
 		return nil, httperr.InternalServerError(httperr.WithInternalErrorf("failed to get existing configuration: %w", err))
 	}
 
-	// Validate step id, validate data against schema, and merge only allowed fields
 	mergedConfig, err := currentStep.ValidateAndMergeData(req.StepId, req.Data, existingConfig)
 	if err != nil {
 		return nil, httperr.BadRequest(err.Error())
-	}
-
-	// For api-key connectors, the credential fields submitted in this step (api_key
-	// and, for the basic placement, the configured username field) belong in the
-	// api_key_credentials table — NOT in the connection's EncryptedConfiguration
-	// blob. Extract them, encrypt into a single opaque blob, persist, and strip
-	// from the merged config before saving so plaintext credentials never reach
-	// the general per-connection config.
-	if def := c.cv.GetDefinition(); def.Auth != nil {
-		if ak, ok := def.Auth.Inner().(*config.AuthApiKey); ok && ak.Placement != nil {
-			if err := c.persistApiKeyCredentialsFromConfig(ctx, mergedConfig, ak.Placement); err != nil {
-				return nil, err
-			}
-		}
 	}
 
 	if err := c.SetConfiguration(ctx, mergedConfig); err != nil {
@@ -75,17 +71,14 @@ func (c *connection) SubmitForm(ctx context.Context, req iface.SubmitConnectionR
 		return nil, httperr.InternalServerError(httperr.WithInternalErrorf("failed to determine next step: %w", err))
 	}
 
-	// Handle each possible next step
 	if nextStep.IsZero() {
 		// Flow complete
 		if err := c.SetSetupStep(ctx, nil); err != nil {
 			return nil, httperr.InternalServerError(httperr.WithInternalErrorf("failed to clear setup step: %w", err))
 		}
-
 		if err := c.SetState(ctx, database.ConnectionStateReady); err != nil {
 			return nil, httperr.InternalServerError(httperr.WithInternalErrorf("failed to set connection state to ready: %w", err))
 		}
-
 		return &iface.ConnectionSetupComplete{
 			Id:   c.GetId(),
 			Type: iface.ConnectionSetupResponseTypeComplete,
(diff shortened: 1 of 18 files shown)


internal/core/service_connection_submit.go
```diff
@@ -51,6 +51,20 @@ func (c *connection) SubmitForm(ctx context.Context, req iface.SubmitConnectionR
 		return nil, httperr.BadRequest(err.Error())
 	}
 
+	// For api-key connectors, the credential fields submitted in this step (api_key
+	// and, for the basic placement, the configured username field) belong in the
+	// api_key_credentials table — NOT in the connection's EncryptedConfiguration
+	// blob. Extract them, encrypt into a single opaque blob, persist, and strip
+	// from the merged config before saving so plaintext credentials never reach
+	// the general per-connection config.
+	if def := c.cv.GetDefinition(); def.Auth != nil {
```
rmorlok: How do we know the submission for this data is specific to the credentials for the api key? Couldn't the data be for a pre-connect data? If I'm understanding this correctly, you are merging all of the data into a single dict, and then using `persistApiKeyCredentialsFromConfig` to pull out fields that are specific to the API Key. Isn't this potentially prone to naming conflicts between the API key and user defined steps?

Is there a way where the auth method can define a setup step itself and then the data for that submission is passed to it based on that id?

internal/core/service_connection_submit.go
```diff
@@ -79,6 +93,17 @@ func (c *connection) SubmitForm(ctx context.Context, req iface.SubmitConnectionR
 	}
 
 	if nextStep.Equals(cschema.SetupStepAuth) {
+		// For api-key the credential was persisted in the preconnect submit above,
+		// so there is no separate auth phase: skip straight to verify / configure /
```
rmorlok: This goes to my earlier point -- is it possible to delegate the idea of there being an auth phase to the OAuth2 auth type? So preconnect and configure exist for all connector types, but the auth phase is OAuth2 specific, and this api key phase would implement a step called `credentials` or something.

internal/core/service_connection_submit.go
rmorlok: Good catch — fixed in 5443bcc. The naming-conflict risk is gone because the dispatcher no longer scans merged config for known credential names. Instead, a new `SetupPhaseCredentials` (indexed phase, sibling of preconnect/configure) is owned by the auth method, and submit routes by phase:

- preconnect / configure submit → generic merge into EncryptedConfiguration (no special handling for api-key fields)
- credentials submit → routed to the auth method's persist path; data never touches EncryptedConfiguration

`Connector.Normalize` now injects the synthesized step into `SetupFlow.Credentials` (renamed from the preconnect target), so an author's preconnect is never altered or interpreted.

Added a regression test (`TestApiKeySubmit_PreconnectFieldNamedApiKeyIsNotTreatedAsCredential`) that declares a preconnect step with a field literally called `api_key` and asserts the value lands in EncryptedConfiguration, not in the credential blob — exactly the case you flagged.

internal/core/service_connection_submit.go
rmorlok: Done in 5443bcc. `SetupPhaseAuth` is now strictly OAuth2's redirect-and-wait step. api-key has its own indexed phase, `SetupPhaseCredentials`, with form steps owned by the auth method. The auth-method-specific shortcut inside the generic submit handler is gone — `nextStep.Equals(SetupStepAuth)` now only triggers `initiateAuthStep`, which is OAuth2-only.

State machine summary:

- Phases: `preconnect` → (`credentials` for api-key | `auth` for OAuth2) → `verify`? → `configure`? → ready
- For NoAuth: preconnect → verify? → configure? → ready (no auth/credentials phase needed)

`Connector.Normalize` materializes the credentials phase for api-key only (other auth types unaffected). `NextSetupStep` chooses the transition based on `HasCredentials()` so the SetupFlow itself carries the signal — no auth-type check leaks into the generic state machine.

internal/core/service_connection_submit.go
```diff
@@ -61,33 +71,126 @@ func (c *connection) SubmitForm(ctx context.Context, req iface.SubmitConnectionR
...
+// submitCredentialsStep handles a submit against a step in the credentials
+// phase. The data is validated against the step's JSON Schema and then handed
+// to the auth method, which is responsible for materializing the credential
+// (encrypting and persisting). Plaintext field data never lands in the
+// connection's general EncryptedConfiguration.
+//
+// After the auth method persists the credential, HandleCredentialsEstablished
+// advances the connection to verify / configure / ready — the same code path
+// OAuth2 takes after its callback.
+func (c *connection) submitCredentialsStep(
```
rmorlok, line 111: The logic in this file needs tests.

rmorlok: Expanded test coverage in 6f3e227. Existing tests covered the happy paths (bearer, basic, verify-transition, no-replay, naming-conflict regression). Added six more covering the remaining branches in \`submitCredentialsStep\` + \`persistApiKeyCredentials\`:

- \`TransitionsToConfigureWhenNoProbes\` — third post-credentials transition (besides ready and verify).
- \`RejectsMismatchedStepId\` — schema-step-id validation surfaces before persistence.
- \`RejectsSchemaViolation\` — JSON Schema enforcement when \`api_key\` is missing.
- \`RejectsBasicMissingUsername\` — basic placement's username field is required.
- \`RejectsUnknownAuthTypeOnCredentialsPhase\` — defensive guard for a credentials phase declared on a non-api-key auth.
- \`DBErrorSurfacesUnchanged\` — \`InsertApiKeyCredential\` failure propagates out and the connection state is NOT advanced (no \`SetConnectionSetupStep\` / \`SetConnectionState\` expected).

All 11 tests in the file pass; full module + preflight green.

rmorlok:
Integration tests pass clean on the rerun (job 76095312017, 6m36s) — all 7 checks green. The earlier failure was `TestCallbackRejection_ActorMismatch` timing out with `websocket url timeout reached`, which is chromedp's internal startup timeout while waiting for headless Chrome's WebSocket endpoint to come up (~20s default).

This is a CI flake on Chrome cold-start, not anything in the diff:
- The test exercises the OAuth2 callback rejection path (the marketplace renders, victim's browser follows a forged callback). The code I changed is the connection-initiate dispatch and the new credentials-phase submit path — neither runs in this test's flow.
- The same test passed on a parallel PR (`oauth2-refresh-rotation`) running against current main.
- The retry was clean.

No code changes needed.

Issue #189, opened 2026-04-28
Token refresh failure should eventually trigger connection needing reauth


rmorlok:
Seeing errors like this in the DEV logs:

```
ERR probe failed service=worker probe_id=calendar-list namespace=root connection_id=cxn_MBsxs7WsBJ7Ev8Xa connector_id=cxr_ck5F3SAkK3CwDpIq connector_version=1 outcome=error duration=66.327583ms error="failed to refresh token: no access token in response"
```

After some set amount of attempts/time, the connection should transition to needing reauth rather than endlessly trying to refresh the tokens.

rmorlok:
Duplicate of #170 (specifically the "Provider returns malformed refresh response" / "Provider returns success without an access token" sub-cases of scenario 7).

The symptom — `failed to refresh token: no access token in response` looping without ever transitioning to reauth — is exactly what #170 covers. The classification (`tokenRefreshMalformedResponse`) is now in place from #169 / PR #265, and is treated as permanent, so the connection's `health_state` flips to unhealthy on the first such failure. The marketplace UI keys off `health_state=unhealthy` to render the reconnect prompt.

Tracking remaining work under #170. Closing this as a duplicate.

rmorlok:
Closed as duplicate of #170. See linked comment for details.

Issue #170, opened 2026-04-27
OAuth tests P0: refresh token failure modes + transient retry policy


rmorlok:
Parent: #159
Phase: 1 (P0) — combines spec scenarios 7 and 8 (deterministic and transient failure modes share helper plumbing)
Spec: #159 scenarios 7, 8
Blocked by: #162

## Scenario 7 — deterministic refresh failures

Cases:
- Refresh token expired.
- Refresh token revoked.
- Refresh token invalid.
- Provider returns \`invalid_grant\`.
- Provider returns malformed refresh response.
- Provider returns success without an access token.

Verify:
- Connection is marked reconnect-required.
- Proxy does not retry indefinitely.
- Existing token state is not corrupted by malformed responses.
- User-facing error indicates that reconnection is required.

## Scenario 8 — transient retry

Cases:
- Refresh endpoint fails once, then succeeds.
- Refresh endpoint fails several times, then succeeds.
- Refresh endpoint fails until retry budget is exhausted.

Verify:
- Retryable failures use bounded retries.
- Retry attempts are observable through logs/metrics.
- Success persists the new token state.
- Final failure is classified correctly.

## Observability

- \`token refresh failed\`, \`refresh retry attempted\`, \`refresh retry exhausted\`.

rmorlok:
## Test approach: go-oauth2-server `/test/*` (preferred)

The primary goal of this scenario is non-interactive (token-endpoint behavior, refresh, revocation, timing, response shape, etc.) — there is no user-flow leg under test. Use the go-oauth2-server `/test/*` control plane to seed state and short-circuit the authorize step rather than booting chromedp.

Reach for chromedp only if the scenario grows to depend on a real consent/login/redirect leg.

rmorlok:
Bug report #189 ("Token refresh failure should eventually trigger connection needing reauth") was closed as a duplicate — its symptom (`no access token in response` looping without reauth transition) is exactly the malformed-response sub-case of scenario 7 here.

rmorlok:
## Coverage Audit — closing

Scenarios 7 and 8 are fully covered. Tests landed across #276 (retry policy), #279 (scenario 7), and #281 (scenario 8).

### Scenario 7 — deterministic refresh failures
`TestProxyRefreshFailure_Scenarios` (in `proxy_refresh_failure_test.go`) drives five scripted refresh-endpoint responses through one shared rig:
[...]

PR #276, opened 2026-05-12
Add bounded retry policy to OAuth2 refresh path (#170 PR A)

rmorlok:
First of three PRs for issue #170 (refresh failure modes + retry policy). Also resolves the symptom reported in #189 — the refresh path was single-shot, so a transient upstream blip surfaced as a permanent refresh failure.

## Summary
- Adds `postRefreshWithRetry` helper on `oAuth2Connection`: 3 attempts with 200ms/400ms linear backoff, retrying transport errors and 5xx responses only. 4xx is returned immediately (refresh tokens are often single-use under rotation; a retried 4xx can actively make things worse).
- Refactors `refreshAccessToken` to use the helper. Plumbs `attempts` through `classifyAndRecordRefreshFailure` and onto the structured `oauth token refresh failed` event (emitted when > 0, omitted at zero so pre-HTTP failure paths don't log a misleading literal zero).
- Unit tests in `proxy_test.go` mirror the postTokenExchangeWithRetry test set: success first attempt, 4xx-not-retried (the load-bearing invariant of the policy), 5xx-then-success, 5xx exhaustion, transport-error retry, and context cancellation during backoff. Plus two new tests pinning the `attempts` field round-trips through the emitted event.

## Followup PRs (issue #170)
- **PR B**: integration tests for scenario 7 — deterministic refresh failures (`invalid_grant`, `invalid_client`, `provider_4xx_other`, malformed 200, 200-without-access-token).
- **PR C**: integration tests for scenario 8 — retry succeeds after N transient 5xxs, exhausts after budget, attempts count on the failure event.

## Test plan
- [x] \`go test ./internal/auth_methods/oauth2/...\` passes (full package, including 11 new test cases for the retry helper + attempts plumbing)
- [x] \`./scripts/preflight.sh\` passes
- [x] Existing integration tests \`TestProxyRefresh_ExpiredAccessTokenRefreshes\` and \`TestProxyRefresh_NoRefreshTokenFlipsUnhealthy\` still pass (refresh path end-to-end with the new helper in place)
- [ ] CI green

### DIFF
--- a/internal/auth_methods/oauth2/proxy.go
+++ b/internal/auth_methods/oauth2/proxy.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log/slog"
 	"net/http"
 	"net/url"
+	"time"
 
 	apauthcore "github.com/rmorlok/authproxy/internal/apauth/core"
 	"github.com/rmorlok/authproxy/internal/core/iface"
@@ -13,6 +15,23 @@ import (
 	"github.com/rmorlok/authproxy/internal/httperr"
 	"github.com/rmorlok/authproxy/internal/httpf"
 	gentleman "gopkg.in/h2non/gentleman.v2"
+)
+
+// Refresh-token retry policy. Mirrors the token-exchange policy in
+// callback.go: a small bounded number of attempts with linear backoff,
+// retrying transport errors and 5xx responses only. 4xx responses
+// (invalid_grant, invalid_client, etc.) are permanent — resubmitting the
+// same refresh token won't change the provider's decision and may, with
+// some providers, count toward a per-refresh-token attempt budget.
+const (
+	// tokenRefreshMaxAttempts is the total number of refresh-endpoint POST
+	// attempts (including the first). 3 = 1 try + 2 retries.
+	tokenRefreshMaxAttempts = 3
+	// tokenRefreshBackoffStep is the linear backoff between attempts:
+	// 200ms before retry 1, 400ms before retry 2. Same shape as
+	// tokenExchangeBackoffStep — short enough not to stall a proxied
+	// request perceptibly, long enough to ride out a node-local hiccup.
+	tokenRefreshBackoffStep = 200 * time.Millisecond
 )
 
 // errNoRefreshToken is returned by refreshAccessToken when the persisted
@@ -32,14 +51,14 @@ func (o *oAuth2Connection) refreshAccessToken(ctx context.Context, token *databa
 	m := o.tokenMutex()
 	err := m.Lock(ctx)
 	if err != nil {
-		return nil, o.classifyAndRecordRefreshFailure(ctx, tokenRefreshInternalError, 0, "", err)
+		return nil, o.classifyAndRecordRefreshFailure(ctx, tokenRefreshInternalError, 0, "", 0, err)
 	}
 	defer m.Unlock(ctx)
 
 	// Get the latest token to make sure we still need to refresh
 	token, err = o.db.GetOAuth2Token(ctx, o.connection.GetId())
 	if err != nil {
-		return nil, o.classifyAndRecordRefreshFailure(ctx, tokenRefreshInternalError, 0, "", err)
+		return nil, o.classifyAndRecordRefreshFailure(ctx, tokenRefreshInternalError, 0, "", 0, err)
 	}
 
 	if mode == refreshModeOnlyExpired && !token.IsAccessTokenExpired(ctx) {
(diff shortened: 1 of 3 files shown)


internal/auth_methods/oauth2/proxy.go
```diff
@@ -259,4 +274,72 @@ func (o *oAuth2Connection) ProxyRequestRaw(ctx context.Context, reqType httpf.Re
...
+func (o *oAuth2Connection) postRefreshWithRetry(
+	ctx context.Context,
+	client *gentleman.Client,
+	tokenEndpoint string,
+	values url.Values,
+) (*gentleman.Response, int, error) {
+	var lastResp *gentleman.Response
+	var lastErr error
+
+	for attempt := 1; attempt <= tokenRefreshMaxAttempts; attempt++ {
```
rmorlok, line 301: This is the second time we've added retry logic to the system. Instead of rolling our own retry every time, let's either pull in an appropriate library that handles context aware retries (e.g. https://github.com/sethvargo/go-retry) or roll our own standard retry implementation in the utils package and test it there. Look through the codebase for other places ware are using retries, come up with a set of requirements and suggest an approach in this PR. Update other callsites that do retries to use the new system.

rmorlok: Surveyed the codebase. Existing retry sites (excluding lock-acquisition retry, which is owned by `redislock`):

| Site | Shape | Attempts | Backoff | Retry when |
| ---- | ----- | -------- | ------- | ---------- |
| `callback.go` `postTokenExchangeWithRetry` | gentleman POST | 3 | linear 200ms step | 5xx + transport |
| `proxy.go` `postRefreshWithRetry` (this PR) | gentleman POST | 3 | linear 200ms step | 5xx + transport |
| `schema/config/key_data_vault_retry.go` `vaultRetryTransport` | `http.RoundTripper` | 5 | exp 2s→60s + jitter | 429 (Retry-After aware) |
| `core/task_disconnect_connection.go` revoke loop | generic op | N | linear `attempt × 1s` | any error |

The OAuth sites and the disconnect loop are pure in-house. The vault transport is a `http.RoundTripper`. Existing transitive deps: `cenkalti/backoff/v4` + `v5`, `hashicorp/go-retryablehttp` (neither directly imported).

## Common requirements (across sites)
- Bounded max attempts, returned to caller (load-bearing for our structured `attempts` field on failure events).
- Context-aware: cancellation during backoff must short-circuit with `ctx.Err()`.
- Retry classification supplied by the caller (5xx vs 429 vs `any-error` vary by site).
- Per-attempt observability hook so each site keeps its current structured "transient failure; retrying" Warn log.
- HTTP-client-agnostic — OAuth sites use gentleman, vault uses `net/http`.

## Site-specific needs that the generic helper must accept (not bake in)
- Backoff strategy: linear (OAuth, disconnect) vs exponential with jitter (vault).
- Retry-After header parsing — vault-only; vault keeps that logic, helper just receives "how long to wait".
- Per-call result type — OAuth returns `*gentleman.Response`, vault returns `*http.Response`. Generics keep this clean.

## Proposed approach: `internal/util/retry` thin wrapper over `cenkalti/backoff/v5`

```go
package retry

type Result[T any] struct {
    Value    T
    Attempts int
}

type Classifier[T any] func(value T, err error) bool // true = retry

type Options[T any] struct {
    MaxAttempts int
    Backoff     backoff.BackOff       // cenkalti/backoff strategy
    Classify    Classifier[T]
    OnRetry     func(attempt int, value T, err error) // per-attempt hook
}

func Do[T any](ctx context.Context, opts Options[T], op func(ctx context.Context) (T, error)) (Result[T], error)
```

Why this shape:
- **`cenkalti/backoff/v5` for the backoff math** — already a transitive dep, well-tested, handles ctx cancellation correctly, provides constant/linear/exponential/jittered strategies out of the box. No new direct dep, no rolling our own backoff math.
- **Generic `T`** so each callsite keeps its native response type (gentleman vs `http.Response`).
- **Classifier callback** so each site decides retryability — OAuth sites pass "5xx or err != nil", vault passes "status == 429", disconnect passes "err != nil".
- **OnRetry hook** preserves each site's current per-attempt Warn log without leaking message strings into the helper.
- **`Result[T].Attempts` returned** matches the load-bearing pattern from this PR / #168 — the structured failure event needs to know the attempt count.

Why not the alternatives:
- **`sethvargo/go-retry`** — functionally equivalent but adds a new direct dep when `cenkalti/backoff` is already transitively pulled in. Same shape, no new advantage.
- **`hashicorp/go-retryablehttp`** — locked to `*http.Client`, would force vault to migrate and would not fit gentleman without a shim. The classification API is also less general (status-code list, not predicate).
- **Pure in-house** — would re-implement what `cenkalti/backoff` already does for ctx-cancellation/backoff-math.

## Suggested migration plan

1. Land `internal/util/retry` with the wrapper + tests for the core invariants (max-attempts honored, ctx cancellation during backoff, classifier respected, OnRetry called the expected number of times, `Result.Attempts` correct on success and exhaustion).
2. Migrate this PR's `postRefreshWithRetry` to use it.
3. Migrate `postTokenExchangeWithRetry` to use it (one-call swap).
4. Migrate `vaultRetryTransport` — replaces its hand-rolled exponential-backoff-with-jitter using `backoff.NewExponentialBackOff`. Retry-After header parsing stays in the transport, just feeds the helper.
5. Migrate the disconnect revoke loop. Generic op, classifier returns `err != nil`.

Steps 1–3 fit in one PR (the new helper + both OAuth callsites). Steps 4 and 5 are independent and can be separate small PRs since they're in unrelated subsystems.

Want me to land this as the next PR (replacing the PR-A-as-is), or as a follow-up after PR A merges with this PR's helper migrated then?

rmorlok: Create a new ticket for the work including your analysis. I'll have a different agent land that separately.

Issue #232, opened 2026-05-09
OTel #3: Outbound proxy instrumentation + label allowlist projection


rmorlok:
Part of #225. Depends on #230.

Instrument the outbound proxy path (`internal/httpf` + gentleman) with client spans, proxy metrics, and the configurable label projection model from #225.

## Scope

- New roundtripper (or extension to the existing one) in `internal/httpf` that emits:
  - **Client spans** per outbound request, including connector identity if available (`connector_id`, `connection_id`, `connector_version`).
  - **Metrics** — proxy request count and duration histogram dimensioned by connector + status class; bytes-in / bytes-out histograms.
- **Label projection** from `httpf.RequestInfo.Labels` (the already-computed effective set produced by `ForConnection` then `ForLabels`):
  - Apply `telemetry.proxy.span_attribute_labels` allowlist to decide which keys become span attributes.
  - Apply `telemetry.proxy.metric_dimension_labels` allowlist (separate list) to decide which keys become metric dimensions.
  - **Telemetry does no merging.** It reads from the effective set and filters by allowlist.
  - Keys not present in the effective set produce no attribute / dimension (they are absent, not empty string).
  - Optional `telemetry.proxy.metric_dimension_value_cap` collapses overflow values to `other` (off by default).

## Acceptance criteria

- Unit tests cover:
  - A request-supplied label correctly **wins over** a connection-inherited value of the same key in the projected attributes/dimensions (validates that telemetry reads from the effective set, not from any individual layer).
  - A label key in `span_attribute_labels` but not `metric_dimension_labels` appears on the span and is absent from metric dimensions (and vice versa).
  - An unlisted key is dropped from both span and metric output.
  - Connector identifying labels (e.g., `type`, `env`) project only when their key is in an allowlist.
  - `metric_dimension_value_cap` collapses overflow values when configured.
- Integration test verifies that proxying through a real (test) HTTP target emits both span and metric.
- `./scripts/preflight.sh` passes.

## Out of scope

- W3C trace context injection on outbound requests (#4 in this project — opt-in).
- Database / Redis / Asynq instrumentation (subsequent tickets).

PR #310, opened 2026-05-14
OpenTelemetry — Step 8: OAuth2 lifecycle telemetry (#237)

### LINKED ISSUE #237: OTel #8: OAuth2 lifecycle telemetry (token exchange, refresh, revocation)
Part of #225. Depends on #230 and #232 (outbound HTTP spans flow through the instrumented `httpf`).

Add spans and metrics around the OAuth2 lifecycle in `internal/auth_methods/oauth2`.

## Scope

- Spans wrapping each meaningful OAuth2 operation: authorization-code exchange, token refresh, token revocation. Errors recorded via `RecordError` / `SetStatus`.
- Metrics:
  - `authproxy_oauth2_refresh_attempts_total{result=success|failure}`
  - `authproxy_oauth2_refresh_failures_total{reason=...}` (high-level reason class only — no PII).
  - `authproxy_oauth2_revocations_total{result=success|failure}`
  - Counters dimensioned by connector (per allowlist-style projection consistent with #232 — only allowlisted labels become dimensions).
- Honor `telemetry.signals.traces` / `telemetry.signals.metrics` toggles.

## Acceptance criteria

- Unit / integration tests for the existing OAuth2 flows assert that the relevant spans and metrics are emitted on success and failure paths.
- Token contents and refresh payloads are **not** captured in span attributes.

## Out of scope

- Per-vendor (Google, Greenhouse, etc.) custom telemetry.

rmorlok:
## Summary

Adds OTel spans + counters for the three OAuth2 lifecycle operations (authorization-code → token exchange, refresh-token grant, token revocation). Piggybacks on the existing structured-event emit sites from #168 / #170 — those already classify failures by category, so this PR maps each classification directly onto OTel without reorganising the business logic.

**Spans** (kind=client; errors recorded via \`RecordError\` + \`SetStatus(Error)\`):
- \`oauth2.token_exchange\` wraps \`exchangeCodeAndAdvance\`
- \`oauth2.refresh\` wraps \`refreshAccessToken\`
- \`oauth2.revoke\` wraps \`RevokeTokens\`

Token contents, refresh payloads, and other secrets are never captured as span attributes.

**Counters** (Int64 monotonic):
- \`authproxy.oauth2.refresh.attempts.total{result, connector_id}\`
- \`authproxy.oauth2.refresh.failures.total{reason, connector_id}\`
- \`authproxy.oauth2.revocations.total{revocation_kind, result, connector_id}\`
- \`authproxy.oauth2.token_exchange.attempts.total{result, connector_id}\`
- \`authproxy.oauth2.token_exchange.failures.total{reason, connector_id}\`

\`reason=\` reuses the existing \`tokenRefreshCategory\` / \`tokenExchangeCategory\` strings (\`invalid_grant\`, \`network_error\`, \`provider_5xx\`, \`malformed_response\`, \`internal_error\`, ...) so operators get a single stable enumeration across logs + metrics + span events.

**Plumbing**: \`oauth2.NewFactory\` takes variadic \`FactoryOption\` (non-breaking — the connections-management route that doesn't need lifecycle telemetry stays unchanged). \`core.NewCoreService\` and \`oauth2.NewTaskHandler\` mirror the shape. \`DependencyManager\` threads providers + telemetry config through the three runtime paths that drive lifecycle operations:

1. **Core service** — used by proxy-time refresh + revoke
2. **Worker task handler** — used by the periodic refresh-expiring-tokens cron
3. **Public OAuth2 route** — used by the callback handler that drives token exchange

Every method on \`*telemetry\` is nil-safe and gates on the resolved signal set — disabled providers, both signals off, or a nil telemetry each degrade to a no-op fast path.

## Test plan

- [x] \`go test ./internal/auth_methods/oauth2/...\` — new tests pin:
  - Refresh success + failure counters emitted
  - Revocation counter emitted (both kinds + both results)
  - Token exchange success + failure counters emitted
  - \`withSpan\` success leaves status unset; failure records error + sets status to Error + adds exception event
  - No-op fast paths: providers disabled, all signals off
  - Nil \`*telemetry\` receiver is safe (every method + \`withSpan\` invokes fn directly)
- [x] Existing OAuth2 tests pass unchanged
- [x] \`./scripts/preflight.sh\` passes
- [x] \`go test ./...\` clean across the whole tree
- [x] \`golangci-lint run\` clean
- [ ] Manual verification with a live collector + real OAuth provider (deferred to #239 dev sample)

Closes #237. Part of the OpenTelemetry project — parent #225.


internal/auth_methods/oauth2/telemetry.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+// metricsActive reports whether metrics are enabled. Safe to call on nil.
+func (t *telemetry) metricsActive() bool {
+	return t != nil && t.metricsEnabled
+}
+
+// connectorAttr returns the connector_id attribute or an empty slice when
+// the id is unset. Connector id is bounded-cardinality (one per connector
+// version row), so it's a safe metric dimension.
```
rmorlok: There could be hundreds of connectors, potentially up to a few thousand. What is a safe value here for bounded cardinality?

rmorlok: Addressed in 27fe702 by dropping `authproxy.connector_id` from every metric attribute. With 1000+ connectors × {result, reason, revocation_kind} we'd hit unacceptable active-series counts on Prometheus, and there's no good universal cap that's right for both a 10-connector and 10,000-connector deployment.

The id stays on **spans** unconditionally — span attributes can carry high-cardinality identifiers because spans aren't a time series; they're stored per-trace and indexed by Tempo/Jaeger. Per-connector breakdowns of OAuth2 lifecycle activity belong in traces (or in the structured request log, which already indexes `connection_id`).

The counters are now aggregate:
- `refresh.attempts.total{result}`
- `refresh.failures.total{reason}`
- `revocations.total{revocation_kind, result}`
- `token_exchange.attempts.total{result}`
- `token_exchange.failures.total{reason}`

Added a test (`TestTelemetry_MetricsExcludeConnectorIDAttribute`) that pins the cardinality contract by asserting `authproxy.connector_id` is absent from every emitted counter data point.

If a deployment later wants per-connector metric breakdowns, the right path is the same #232 allowlist + value-cap on connection labels (e.g. `type`, `env` — bounded by definition), not raw connector ids. Happy to add that plumbing in a follow-up if it'd be useful.

rmorlok: Followed up with the allowlist projection too in 82a8804. The spec from #225 / #237 called for "per allowlist-style projection consistent with #232" and I'd missed that on the first cut, so this round adds it.

What landed:

- Extracted the `labelProjector` from `internal/httpf` to `internal/aptelemetry` as exported `aptelemetry.LabelProjector` (the projector type is now shared between httpf and oauth2; no duplicate state machine).
- OAuth2's `recordRefresh*` / `recordRevocation` / `recordTokenExchange*` methods take `connection.GetLabels()` and pass them through the projector. The same `telemetry.proxy.metric_dimension_labels` allowlist + `metric_dimension_value_cap` from #232 now govern both proxy and OAuth2 metrics.
- `connector_id` stays a span attribute only (still absent from metric dimensions per the prior round of this review).

Behavior:
- Unlisted connection-label keys are dropped from metric attributes.
- Allowlisted keys (e.g. `type=google_drive`, `env=prod`) appear on every OAuth2 counter so dashboards can break down refresh / revocation / exchange activity by connector class.
- Cap exhaustion collapses overflow values to `"other"` — same semantics as proxy metrics.

Two new tests pin this:
- `TestTelemetry_AllowlistedLabelsProject` — allowlisted keys land on metric attrs, unlisted keys (e.g. `tenant`, `internal`) do not.
- `TestTelemetry_ValueCapAppliesToProjectedLabels` — third distinct value of an allowlisted key collapses to `"other"`.

Plus the projector-only tests moved to `internal/aptelemetry/labelprojector_test.go` so the shared type has its own focused test surface.

Issue #298, opened 2026-05-13
Admin UI: rate-limit dry-run page + SDK + nav wiring


rmorlok:
Part of #294. The visible end of the dry-run feature.

Depends on #296 (backend endpoint) and #297 (\`<RequestForm>\` component).

## Scope

**SDK addition** in \`sdks/js/src/rateLimits.ts\`:

\`\`\`ts
export interface RateLimitDryRunRequest { request: ...; context: ...; }
export interface RateLimitDryRunResponse {
    request_label_snapshot: Record<string, string>;
    matched: Array<{ rate_limit_id: string; namespace: string; effective_mode: string;
                     bucket_key: string; algorithm_summary: string;
                     would_allow: boolean; remaining: number; retry_after_ms: number;
                     peek_failed: boolean; }>;
    not_matched: Array<{ rate_limit_id: string; reason: string; }>;
}
export const dryRunRateLimits = (req: RateLimitDryRunRequest) =>
    client.post<RateLimitDryRunResponse>('/api/v1/rate-limits/_dry_run', req);
\`\`\`

Add to the \`rateLimits\` namespace object.

**New page** at \`/rate-limits/_dry_run\` (\`ui/admin/src/pages/RateLimitDryRun.tsx\`):

- Two-column layout: \`<RequestForm>\` on the left, results panel on the right.
- "Run dry run" button submits the request; results render in place.
- **Matched** section: each row shows id (link to detail page), would-allow chip (green / red), remaining, retry-after (when not allowed), bucket key, algorithm summary, and effective mode (chip).
- **Not matched** section: id (link) + one-line miss reason.
- **Computed label snapshot** in a collapsible block at the bottom for debugging.
- Surface \`peek_failed: true\` as a warning-coloured row badge ("couldn't read counter; runtime would fail-open").

**Entry points:**

- Add route + breadcrumb in \`ui/admin/src/routes.tsx\` (parent breadcrumb is "Rate Limits").
- Add a secondary "Dry run" button on the Rate Limits list page header, next to "Create rate limit".
- No side-nav entry — the dry-run is discoverable from the Rate Limits section.

## Acceptance

- \`./scripts/preflight.sh\` and \`yarn workspace @authproxy/admin {typecheck,lint,test}\` clean.
- Manual exercise via Chrome DevTools MCP against a local stack. Screenshots in the PR description: empty form, populated form with results (mix of matched + not-matched), and a peek-failed state (kill Redis or block the connection).
[...]

Issue #295, opened 2026-05-13
Rate limits: Limiter.Peek() — read-only counter inspection


rmorlok:
Part of #294. Backend prerequisite for the rate-limit dry-run endpoint.

## Why

The dry-run endpoint needs to answer "would this request be rate-limited *right now*" without actually consuming a token / incrementing a counter. The current `Limiter.Decide()` always writes (atomic check-and-increment via Lua). Adding `Peek()` lets the dry-run share the same math without polluting the runtime counters.

## Scope

Add to the `Limiter` interface in \`internal/ratelimit/limiter.go\`:

\`\`\`go
// Peek returns what Decide would return for bucketKey, given the current
// counter state, without writing. Same fail-open semantics as Decide.
Peek(ctx context.Context, bucketKey BucketKey) (Decision, error)
\`\`\`

Implement across all three algorithms:

- **Token bucket** — read \`tokens\` + \`last_refill_ms\`, compute the refill that *would* happen, return Allowed iff tokens-after-refill >= 1. Don't write.
- **Fixed window** — read the current window's counter, return Allowed iff count < limit. Don't write the counter or its TTL.
- **Sliding window** — read both window counters and recompute the weighted sum; same comparison as Decide but no \`ZADD\` / \`INCR\`.

\`Decision.Remaining\` and \`Decision.RetryAfter\` should be populated the same way \`Decide\` populates them.

## Acceptance

- New \`Peek\` method on \`Limiter\`, implemented in all three limiter files.
- Unit tests per algorithm covering: (a) under limit → Allowed=true with correct Remaining, (b) over limit → Allowed=false with correct RetryAfter, (c) Peek does not change Redis state (call Peek twice, observe identical results and unchanged Redis state).
- Fail-open on Redis errors with \`FailedOpen=true\`, matching \`Decide\`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #294, opened 2026-05-13
Add UI to support knowing for a given request which rate limits will apply


rmorlok:
## Goal

Provide a "would this request be rate-limited?" diagnostic so operators can validate selectors / buckets / algorithms without sending real traffic. The endpoint must match on a synthesized request, report which rate limits apply (and which would 429), and **must not increment counters**. The admin UI piece should expose this through a reusable HTTP-request form that we can later reuse to drive real proxy requests.

## Server

**New endpoint:** `POST /api/v1/rate-limits/_dry_run`

Request:

```json
{
  "request": {
    "method": "POST",
    "path": "/services/data/v54.0/sobjects/Account",
    "request_type": "proxy",
    "headers": { "X-Foo": "bar" }
  },
  "context": {
    "actor_id": "act_abc",
    "connection_id": "conn_xyz",
    "namespace": "root.acme"
  }
}
```

- `context.connection_id` (when given) hydrates the connector + per-request label snapshot the way the runtime would.
- Otherwise `context.namespace` + `context.actor_id` are used directly; labels are an explicit override field on `context` (so the dry-run can reproduce non-connection contexts).
- Ruleset is read from **the same cache the enforcer uses**, so the result reflects propagation delay.

Response:

```json
{
  "request_label_snapshot": { "apxy/connector/-/id": "salesforce", ... },
  "matched": [
    {
      "rate_limit_id": "rl_abc",
      "namespace": "root.acme",
      "effective_mode": "enforce",
      "bucket_key": "actor=act_abc|labels/team=acme",
      "algorithm_summary": "token bucket 60 @ 1/s",
      "would_allow": true,
      "remaining": 42,
      "retry_after_ms": 0,
      "peek_failed": false
    }
  ],
  "not_matched": [
    { "rate_limit_id": "rl_xyz", "reason": "method GET does not satisfy POST|PATCH|PUT" }
  ]
}
```

`peek_failed=true` mirrors the runtime's fail-open behaviour when Redis is unreachable — the UI surfaces it as "couldn't read counter; runtime would fail-open."

### New Limiter capability — `Peek()`

The current limiters atomically check-and-increment via Lua. Add a sibling read-only Lua script per algorithm that reads counter state and computes the same `Decision` (`Allowed`, `RetryAfter`, `Remaining`) **without writing**. New method on the Limiter interface:

```go
Peek(ctx context.Context, bucketKey BucketKey) (Decision, error)
```

Fail-open semantics match `Decide()`.

[...]

PR #311, opened 2026-05-14
Add rate-limits _dry_run admin-api endpoint

### LINKED ISSUE #296: Rate limits: _dry_run admin-api endpoint
Part of #294. Backend endpoint that powers the dry-run admin UI.

Depends on #295 (Limiter.Peek()).

## Scope

Add \`POST /api/v1/rate-limits/_dry_run\` to admin-api.

**Request:**

\`\`\`json
{
  "request": {
    "method": "POST",
    "path": "/services/data/v54.0/sobjects/Account",
    "request_type": "proxy",
    "headers": { "X-Foo": "bar" }
  },
  "context": {
    "actor_id": "act_abc",
    "connection_id": "conn_xyz",
    "namespace": "root.acme",
    "labels": { "team": "acme" }
  }
}
\`\`\`

- When \`context.connection_id\` is given, hydrate connector + per-request label snapshot the way the runtime does (carry-forward from namespace → connector → connection).
- Otherwise use \`context.namespace\`, \`context.actor_id\`, and \`context.labels\` as the raw inputs. \`labels\` is a manual override that takes precedence over carry-forward.

**Response:**

\`\`\`json
{
  "request_label_snapshot": { "apxy/connector/-/id": "salesforce" },
  "matched": [
    {
      "rate_limit_id": "rl_abc",
      "namespace": "root.acme",
      "effective_mode": "enforce",
      "bucket_key": "actor=act_abc|labels/team=acme",
      "algorithm_summary": "token bucket 60 @ 1/s",
      "would_allow": true,
      "remaining": 42,
      "retry_after_ms": 0,
      "peek_failed": false
    }
  ],
  "not_matched": [
    { "rate_limit_id": "rl_xyz", "reason": "method GET does not satisfy POST|PATCH|PUT" }
  ]
}
\`\`\`

**Pipeline:**

1. Hydrate \`RequestContext\` from the inputs.
2. Pull the ruleset from the **enforcer cache** (same source the runtime reads), not directly from the DB.
3. Run the existing \`matcher\` against the ruleset; collect matched + not-matched (with a one-line miss reason per rule).
4. For each matched rule, call the new \`Limiter.Peek()\` to populate \`would_allow\` / \`remaining\` / \`retry_after_ms\` / \`peek_failed\`.
5. Include both enforce-mode and observe-mode matches in \`matched\` — operators should see everything that fires.

## Acceptance

- New route + handler registered on admin-api with appropriate auth (admin namespace permissions).
- Swagger annotations + regenerated docs (via \`./scripts/preflight.sh\`).
- Integration test covering: connection-driven hydration, raw-context input, label override, an observe-mode match, a 429 result (would_allow=false with retry_after_ms), and a not-matched rule with each kind of miss reason (method, path, label-selector, request_type).
- Verify counter state is unchanged after the call (read counter, dry-run, read counter, assert equal).

rmorlok:
Closes #296.

## Summary

Adds \`POST /api/v1/rate-limits/_dry_run\`, the diagnostic endpoint behind #294 that lets operators validate "would this request be rate-limited?" without sending real traffic. The endpoint runs a synthesized request through the matcher against the same in-memory rule cache the enforcer uses (so cache-propagation delay is visible) and calls the new \`Limiter.Peek\` per match — counters are not incremented.

## Pipeline

1. Hydrate \`RequestContext\` from the input. \`context.connection_id\` (when given) pulls namespace / connector / labels from the connection the way \`httpf.ForConnection\` does at runtime. Otherwise \`context.namespace\` + \`context.actor_id\` + \`context.labels\` are used directly; \`context.labels\` always wins over carry-forward.
2. Cascade-filter rules from the enforcer cache to those at the request's namespace or any ancestor (matches the runtime's visibility).
3. Run the matcher; for each not-matched rule, a per-clause miss reason is reported via a new \`MatchExplain\` (request_type / method / label selector / path match).
4. For each matched rule, construct a Limiter (from #295) and call \`Peek\` to populate \`would_allow\` / \`remaining\` / \`retry_after_ms\` / \`peek_failed\`. Observe-mode matches are included — the whole point of a dry-run is to see everything that fires.

## Auth

The route runs under the existing rate-limit auth surface with \`verb=get\`. Callers must be able to read rate-limits in the request's namespace; rules outside that namespace are filtered out before the matcher runs.

## Wire shape

Request:

```json
{
  \"request\": { \"method\": \"POST\", \"path\": \"/v1/things\", \"request_type\": \"proxy\", \"headers\": {} },
  \"context\": { \"namespace\": \"root.acme\", \"actor_id\": \"act_test\", \"labels\": { \"team\": \"acme\" } }
}
```

Response:

```json
{
  \"request_label_snapshot\": { \"team\": \"acme\" },
  \"matched\": [
    {
      \"rate_limit_id\": \"rl_abc\",
      \"namespace\": \"root.acme\",
      \"effective_mode\": \"enforce\",
      \"bucket_key\": \"actor=act_test\",
      \"algorithm_summary\": \"token bucket 60 @ 1/s\",
      \"would_allow\": true,
      \"remaining\": 59,
      \"retry_after_ms\": 0,
      \"peek_failed\": false
    }
  ],
  \"not_matched\": [
    { \"rate_limit_id\": \"rl_xyz\", \"namespace\": \"root\", \"reason\": \"method \\\"POST\\\" does not match the rule's method list\" }
  ]
}
```

## Test plan

- [x] `./scripts/preflight.sh` (clean — swagger regenerated and committed)
- [x] `go test ./internal/routes/ ./internal/ratelimit/ -count=1` (passes — 10 new dry-run tests)

Per-scenario coverage in the new `TestRateLimits/dry_run` subtree:

- 400 when \`request.method\` missing
- 400 when neither \`context.connection_id\` nor \`context.namespace\` given
- Matched rule reports peek result without consuming counter (two consecutive calls return identical remaining)
- Each miss-reason category surfaces correctly (method, request_type, path prefix)
- Manual \`context.labels\` flow into the \`request_label_snapshot\` and satisfy a label-selector rule
- Observe-mode match is included with \`effective_mode=\"observe\"\`
- 403 when caller lacks permission in the request's namespace
- Rules outside the request's namespace are filtered out (neither matched nor not-matched)
- **Counter-state-unchanged invariant**: 5 dry-runs against a capacity-2 bucket all report would_allow=true with remaining=1 — proves Peek is not consuming behind the scenes

## Unblocks

#298 (the dry-run admin UI page).


internal/routes/rate_limits.go
```diff
@@ -444,6 +509,233 @@ func (r *RateLimitsRoutes) delete(gctx *gin.Context) {
...
+		apgin.WriteError(gctx, nil, httperr.Forbidden(err.Error(), httperr.WithInternalErr(err)))
+		val.MarkErrorReturn()
+		return
+	}
+
+	// Run against the same ruleset the enforcer sees so operators
+	// observe propagation delay if any. The cache is empty until the
+	// refresher has installed its first snapshot — we treat that as
+	// "no rules to evaluate" rather than an error.
+	rules := r.rlCache.All()
```
rmorlok: Please move the logic for actually testing the rate limit (hydrating, looking up data in the database, etc) into the internal/core package. Routes should be limited to request validation and response formatting.

rmorlok: Moved in df371d5. The handler is now ~30 lines (bind → core call → validate namespace → marshal); all hydration, cache filtering, matcher, and Peek work lives in `core.DryRunRateLimit` and is covered by direct unit tests in `internal/core/service_rate_limit_dry_run_test.go`. Plumbed the rate-limit cache into the core service via a new `WithRateLimitCache` functional option so existing `NewCoreService` callers don't have to change.

internal/core/iface/rate_limit.go
```diff
@@ -35,3 +35,71 @@ type ListRateLimitsBuilder interface {
...
+	Request DryRunRequestPayload
+
+	// Context is the identity the request runs under. When ConnectionId
+	// is set, the implementation hydrates Namespace + Connector +
+	// Labels from the connection (the way httpf.ForConnection does at
+	// runtime). Manual Labels always merge on top.
+	Context DryRunRequestContext
+}
+
+type DryRunRequestPayload struct {
```
rmorlok: Let's reuse the `ProxyRequest` struct. That will mean that labels move to the request instead of the context. This will affect the routes side data as well.

rmorlok: Done in c2e4132. `DryRunRateLimitRequest.Request` is now `iface.ProxyRequest` directly; `Labels` moved to the request (off the context); `request_type` stays as a sibling field on the dry-run input since the real proxy entrypoint takes it separately from the request body too (`Proxy.ProxyRequest(ctx, reqType, req)`). Route + swagger updated; tests rewritten to use `url` and request-level labels.

Issue #250, opened 2026-05-10
[api-key] AuthApiKey schema with placement variants


rmorlok:
Part of #243 — see that issue for full project context, goals, and out-of-scope notes.

## Scope

Extend `AuthApiKey` (`internal/schema/connectors/auth_api_key.go`) into a fully specified type with placement variants. **Schema and validation only** — no runtime behavior.

### Placements

| Placement | Emits | Configurable |
|---|---|---|
| `bearer` | `Authorization: Bearer <key>` | none |
| `header` | `<HeaderName>: <Prefix><key>` | `header_name` (required), `prefix` (optional) |
| `query` | `?<ParamName>=<key>` (URL-encoded) | `param_name` (required) |
| `basic` | `Authorization: Basic <base64(username:key)>` | `username_field` (required) |

### YAML shape

See #243 Part 1 for the illustrative shape.

### Validation

- Exactly one placement variant must be specified.
- `header_name` / `param_name` must match valid HTTP header / URL query token grammar.
- For `basic`, the `username_field` name must be referenced in the credential-collection form (auto-generated or explicit).

## Out of scope (other sub-issues)

- Auto-form generation and initiate-flow dispatch
- `api_key_credentials` storage
- Proxy injection
- Greenhouse stub cleanup

## Test plan

- [ ] YAML unmarshal + validation tests for each placement (positive cases)
- [ ] Validation rejects missing required fields per placement
- [ ] Validation rejects multiple placement variants set
- [ ] Validation rejects malformed header / query field names

## Dependencies

None — foundational PR.

PR #317, opened 2026-05-14
Add api-key proxy package (#253)

### LINKED ISSUE #253: [api-key] internal/auth_methods/api_key proxy package

rmorlok:
Closes #253. Part of #243.

## Summary
New \`internal/auth_methods/api_key/\` package implementing \`iface.Proxy\`. With #250–#252 already merged, api-key connections can be created end-to-end but couldn't actually proxy outbound requests. This closes that gap.

Per-request flow:
1. Load active \`api_key_credentials\` row for the connection.
2. Decrypt the credentials blob (\`ApiKeyCredentialPlaintext\` JSON).
3. Apply the \`placement_snapshot\` stored on the row (or fall back to the live connector definition if no snapshot — defensive for rows that predate snapshot capture).
4. Inject per placement:
   - \`bearer\` → \`Authorization: Bearer <key>\`
   - \`header\` → \`<HeaderName>: <Prefix><key>\`
   - \`query\` → \`<ParamName>=<key>\` appended to URL (existing params preserved)
   - \`basic\` → \`Authorization: Basic <base64(username:key)>\` (RFC 7617)

## Design note: pure-data inject helper
The credential→auth mapping is \`computeAuthApplication\`, which returns the (header name, header value) + (query name, query value) pair as plain data. The proxy applies the result via gentleman's \`SetHeader\` / \`SetQuery\`. This keeps unit tests fast and assertable without booting an HTTP transport — \`inject_test.go\` covers every placement and error branch directly against the return value, and \`proxy_test.go\` covers the surrounding \`resolveAuth\` (DB read + decrypt + placement resolution).

## Wire-up
- \`internal/core/service_api_key.go\`: lazy factory accessor mirroring \`getOAuth2Factory\`. One factory per service, shared across connections.
- \`internal/core/connection_proxy.go\`: dispatch branch for \`AuthApiKey\`.
- \`internal/core/service.go\`: factory field + once guard.

## Side benefit
\`proxy_http\` probes against api-key connectors now run end-to-end (previously \`ErrProxyNotImplemented\`). Prerequisite for #255 to actually validate api-key health.

## Mock change
\`internal/core/mock/connection.go\` gains a \`ConnectorVersionEntity\` field so the resolveAuth fallback test can inject a live connector definition. Default behaviour (nil) is unchanged.

## Test plan
- [x] Bearer, header (with and without prefix), query, basic — \`computeAuthApplication\` unit tests.
- [x] Basic correctly base64-encodes \`username:key\` per RFC 7617, including the colon-in-username edge case.
- [x] Query value passes through verbatim (URL encoding is the transport layer's job).
- [x] Header honors prefix (with trailing space, with non-space separator like \`v1_\`).
- [x] Missing required fields surface errors: nil placement, empty api_key, missing header_name / param_name / username.
- [x] \`resolveAuth\` happy path with placement_snapshot.
- [x] \`resolveAuth\` falls back to live connector definition when snapshot is nil.
- [x] DB error from \`GetActiveApiKeyCredential\` propagates.
- [x] Malformed plaintext (non-JSON) surfaces \`unmarshal\` error.
- [x] Basic placement end-to-end resolves both username and key from the decrypted plaintext.
- [x] Full module green on SQLite + Postgres.

### DIFF
--- a/internal/auth_methods/api_key/proxy.go
+++ b/internal/auth_methods/api_key/proxy.go
@@ -1,0 +1,104 @@
+package api_key
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+
+	apauthcore "github.com/rmorlok/authproxy/internal/apauth/core"
+	"github.com/rmorlok/authproxy/internal/core/iface"
+	"github.com/rmorlok/authproxy/internal/database"
+	"github.com/rmorlok/authproxy/internal/httpf"
+	cschema "github.com/rmorlok/authproxy/internal/schema/connectors"
+)
+
+// ProxyRequest authenticates an outbound proxied request with the connection's
+// active api-key credential and returns the upstream response.
+//
+// Each call loads the active credential fresh from the database so that a
+// rotation (which inserts a new row and soft-deletes the prior) takes effect
+// immediately on the next request — no in-memory cache to invalidate.
+func (a *apiKeyConnection) ProxyRequest(ctx context.Context, reqType httpf.RequestType, req *iface.ProxyRequest) (*iface.ProxyResponse, error) {
+	app, err := a.resolveAuth(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	r := a.httpf.
+		ForRequestType(reqType).
+		ForConnection(a.connection).
+		ForActor(apauthcore.ActorFromContext(ctx)).
+		ForLabels(req.Labels).
+		New().
+		UseContext(ctx).
+		Request()
+
+	req.Apply(r)
+	if app.HeaderName != "" {
+		r.SetHeader(app.HeaderName, app.HeaderValue)
+	}
+	if app.QueryName != "" {
+		r.SetQuery(app.QueryName, app.QueryValue)
+	}
+
+	resp, err := r.Do()
+	if err != nil {
+		return nil, err
+	}
+	return iface.ProxyResponseFromGentlemen(resp)
+}
+
+// ProxyRequestRaw is not yet implemented for api-key. Mirrors the current
+// state of no_auth — when the streaming-proxy path is added, both auth methods
+// will adopt the same pattern.
+func (a *apiKeyConnection) ProxyRequestRaw(ctx context.Context, reqType httpf.RequestType, req *iface.ProxyRequest, w http.ResponseWriter) error {
+	return nil
+}
+
+// resolveAuth fetches the active credential row, decrypts the plaintext, and
+// computes the authentication application for the placement that was in effect
+// when the credential was submitted. The placement_snapshot stored on the
+// credential row is used in preference to the live connector definition so an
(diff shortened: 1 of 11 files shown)


internal/auth_methods/api_key/proxy.go
```diff
@@ -0,0 +1,104 @@
...
+func (a *apiKeyConnection) resolveAuth(ctx context.Context) (authApplication, error) {
+	cred, err := a.db.GetActiveApiKeyCredential(ctx, a.connection.GetId())
+	if err != nil {
+		return authApplication{}, fmt.Errorf("failed to load api-key credential: %w", err)
+	}
+
+	placement := cred.PlacementSnapshot
+	if placement == nil {
+		// Fall back to the connector definition's placement when the credential
+		// row didn't snapshot one (rows inserted before snapshot wiring shipped).
```
rmorlok: You do not need to support cases before this was added. Assume no production deploys for now.

rmorlok: Dropped in c761fc7. \`resolveAuth\` now requires \`PlacementSnapshot\` and surfaces an error if it's missing — no fallback, no live-definition lookup. Also reverted the \`mock/connection.go\` \`ConnectorVersionEntity\` field that was added only to support the fallback test, since it's no longer needed.

Net: \`proxy.go\` -16 lines, \`mock/connection.go\` back to its prior shape.

PR #324, opened 2026-05-15
Remove ProxyRequestRaw — no callers, can re-add when needed

### LINKED ISSUE #319: Streaming + raw proxy: SSE responses, large uploads, ap curl
## Goal

Support HTTP endpoints that stream data (LLM token streams, SSE) and proxy to systems that handle large chunked uploads (S3 multipart). The wrapped JSON `_proxy` path can't do this because it buffers the entire request and response — we need a path that streams byte-for-byte while still applying connector authentication, rate limits, and request logging.

## HTTP contract

A new header-driven endpoint:

```
ANY /api/v1/connections/{id}/_proxy_raw
X-AuthProxy-Upstream-URL: https://api.openai.com/v1/chat/completions
X-AuthProxy-Label: <key>=<value>        # repeated; one per label
[ any other headers the client sets pass through, minus Authorization
  and hop-by-hop headers ]
```

- **Method** matches inbound (any).
- **Upstream URL** lives in `X-AuthProxy-Upstream-URL`. Required; matches the existing wrapped path's "client supplies the full URL" model.
- **Labels** ride on repeated `X-AuthProxy-Label` headers using `<key>=<value>` syntax. Multiple headers OK (one per label). Server splits on the first `=`. Label keys with `/` (e.g. `apxy/connector/-/id`) are valid in header values but not header names, hence the value-encoded form.
- **Body** streams through — no buffering. Inbound body is piped onto the outbound request body; response body is copied back with `http.Flusher` flushes between reads.
- **Auth header** is stripped from inbound and replaced by the connector's authentication (Bearer / API key / nothing).
- **Hop-by-hop headers** stripped per RFC 7230.

## Layering

Today, `connection.ProxyRequest` dispatches to the auth method, which both attaches credentials and runs the HTTP call. For streaming this entanglement is awkward — the streaming machinery doesn't want to live inside oauth2/no_auth/api_key. So we factor proxy orchestration out of auth methods entirely.

New shape:

```go
// internal/auth_methods/...
type Authenticator interface {
    Attach(ctx context.Context, req *http.Request) error
    // For credentials that can be refreshed (OAuth2): called when a
    // request returned 401. Returning nil signals "retry once with my
    // freshly-attached credentials." Returning an error surfaces the
    // failure to the caller.
    RecoverFrom401(ctx context.Context) error
}
```

- Each auth method (oauth2, no_auth, api_key) implements `Authenticator`.
- `internal/proxy/` (new package) owns both `ProxyRequest` (wrapped JSON) and `ProxyRequestRaw` (streaming). It builds the outbound request, calls `Authenticator.Attach`, sends, and (for the wrapped path) reads the full response into `ProxyResponse`. For the raw path it streams the response back to a supplied `http.ResponseWriter` with flushes.
- `Connection.ProxyRequest` / `ProxyRequestRaw` become thin shims that delegate to `internal/proxy/` with the right `Authenticator`.

This makes the streaming and wrapped paths share orchestration code (rate-limit round-tripper, telemetry, attribution, refresh-on-401), and lets each auth method stay narrow.

## Streaming impl

- Use `http.Client.Do` (not gentleman's `Send()` which reads the full body). The factory still produces the client, so the rate-limit enforcer round-tripper and OTel instrumentation stay attached.
- Inbound `*http.Request.Body` is moved directly onto the outbound `*http.Request.Body`.
- Response body copy uses `io.Copy(w, resp.Body)` with a `bufio.Writer` flushed after every read so SSE / chunked responses appear at the client byte-for-byte.
- Trailers passed through.
- Rate limits: existing `EnforcerRoundTripper` already gives one Decide-per-request, so raw mode inherits it.

## Request logging

- **Metadata always**: method, URL, status, headers, timings, attribution, rate-limit match set.
- **Body capture is opt-in by size and known length**:
  - Request: if inbound `Content-Length` is set and `<= max_request_size`, tee bytes through `splitReadCloser` to blob storage. Otherwise (chunked, oversized, unknown length) record `body_skipped: streaming` or `body_skipped: too_large` and pass through without tee.
  - Response: same logic mirror-applied against `maxFullResponseSize` and the response's `Content-Length`.
- Reuses existing `internal/request_log/io.go` machinery + `captureConfig` knobs — no new schema.

## CLI

Three commands:

1. **Rename `raw-proxy` → `signing-proxy`**. The existing JWT-signing reverse-proxy that lets clients hit the AuthProxy admin/api with proper auth — name was always misleading.
2. **New `ap proxy --connection <id>`**: a local reverse-proxy that forwards through the new `_proxy_raw` endpoint. Two modes:
   - Caller hits `http://localhost:9999/v1/chat/completions` → `ap proxy` derives the upstream URL by appending the path to the connection's configured base URL (or to an explicit `--upstream-base <url>` flag) and sets `X-AuthProxy-Upstream-URL` for the user.
   - Caller sets `X-AuthProxy-Upstream-URL` directly → `ap proxy` passes it through unmodified.
3. **New `ap curl <curl args>`**: same argv as real `curl`. Internally boots an `ap proxy` listener on an ephemeral port, swaps the user-supplied URL for `http://127.0.0.1:<port>/<path>`, and runs the request. Streaming responses go to stdout exactly like real curl.

## Integration tests

- **Chunked upload to MinIO**: drive a 50MB multipart `aws s3 cp` (or equivalent) through `_proxy_raw` to the docker-compose MinIO; assert byte-identical file at the destination.
- **SSE streaming response**: in-process fake LLM server that emits `data: {token}\n\n` events with `time.Sleep(50ms)` between each. Assert AuthProxy delivers each event to the client within ~100ms of the server emitting it (proves end-to-end no buffering).

## Decisions locked

- Method scope: all HTTP methods accepted.
- Auth verb: reuse `connections:proxy`; no new verb.
- Label header format: repeated `X-AuthProxy-Label: <key>=<value>`. Split on first `=`. Values may contain `=` only if percent-encoded (deferred until we hit one).

## Sub-tickets

Tracked separately so each piece can land independently. Numbered for natural landing order — (1) should land first because (2) builds on the new orchestration layer; (3)/(4)/(5) can go in parallel after (2).

- [ ] #329 — (1) Factor proxy orchestration out of auth methods into `internal/proxy/`; add `Authenticator` interface
- [ ] #330 — (2) Implement raw streaming endpoint + header parsing
- [ ] #331 — (3) Size-bounded body capture for the raw path
- [ ] #332 — (4) CLI: rename + `ap proxy` + `ap curl`
- [ ] #333 — (5) Integration tests: MinIO chunked upload + SSE stream

Sub-ticket references will be linked back here as each is filed.

rmorlok:
Closes #319.

## Summary

`ProxyRequestRaw` was sketched for a "direct proxy without JSON wrapping" path but never landed any production callers — both `oauth2` and `no_auth` implementations just `return nil`, and the only invocation outside the implementations themselves was a delegation shim in `connection_proxy.go`. Removing it now per the issue's "or remove" option.

If a streaming / no-wrapper variant becomes warranted later, this can be resurrected behind a real implementation.

## Touched

- `internal/core/iface/proxy.go` — removed method from `Proxy` interface.
- `internal/core/iface/connection.go` — removed method from `Connection` interface; dropped now-unused `net/http` import.
- `internal/core/connection_proxy.go` — dropped the dispatch shim.
- `internal/core/mock/connection.go` — removed the hand-written stub; dropped now-unused `net/http` import.
- `internal/auth_methods/oauth2/{interface.go,proxy.go}` — removed from interface and concrete impl.
- `internal/auth_methods/no_auth/{interface.go,proxy.go}` — same.
- `internal/auth_methods/oauth2/mock/oauth2.go`, `internal/auth_methods/no_auth/mock/noauth.go` — regenerated via `go generate`.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./internal/core/... ./internal/auth_methods/... -count=1` passes
- [x] `./scripts/preflight.sh` clean (no swagger drift)

Diff: `-77 +2` lines.

### DIFF
--- a/internal/auth_methods/no_auth/interface.go
+++ b/internal/auth_methods/no_auth/interface.go
@@ -2,7 +2,6 @@ package no_auth
 
 import (
 	"context"
-	"net/http"
 
 	connIface "github.com/rmorlok/authproxy/internal/core/iface"
 	"github.com/rmorlok/authproxy/internal/httpf"
@@ -11,5 +10,4 @@ import (
 //go:generate mockgen -source=./interface.go -destination=./mock/noauth.go -package=mock
 type NoAuthConnection interface {
 	ProxyRequest(ctx context.Context, reqType httpf.RequestType, req *connIface.ProxyRequest) (*connIface.ProxyResponse, error)
-	ProxyRequestRaw(ctx context.Context, reqType httpf.RequestType, req *connIface.ProxyRequest, w http.ResponseWriter) error
 }
--- a/internal/auth_methods/no_auth/mock/noauth.go
+++ b/internal/auth_methods/no_auth/mock/noauth.go
@@ -6,12 +6,11 @@ package mock
 
 import (
 	context "context"
-	http "net/http"
 	reflect "reflect"
 
 	gomock "github.com/golang/mock/gomock"
 	iface "github.com/rmorlok/authproxy/internal/core/iface"
-	"github.com/rmorlok/authproxy/internal/httpf"
+	httpf "github.com/rmorlok/authproxy/internal/httpf"
 )
 
 // MockNoAuthConnection is a mock of NoAuthConnection interface.
@@ -50,18 +49,4 @@ func (m *MockNoAuthConnection) ProxyRequest(ctx context.Context, reqType httpf.R
 func (mr *MockNoAuthConnectionMockRecorder) ProxyRequest(ctx, reqType, req interface{}) *gomock.Call {
 	mr.mock.ctrl.T.Helper()
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "ProxyRequest", reflect.TypeOf((*MockNoAuthConnection)(nil).ProxyRequest), ctx, reqType, req)
-}
-
-// ProxyRequestRaw mocks base method.
-func (m *MockNoAuthConnection) ProxyRequestRaw(ctx context.Context, reqType httpf.RequestType, req *iface.ProxyRequest, w http.ResponseWriter) error {
-	m.ctrl.T.Helper()
-	ret := m.ctrl.Call(m, "ProxyRequestRaw", ctx, reqType, req, w)
-	ret0, _ := ret[0].(error)
-	return ret0
-}
-
-// ProxyRequestRaw indicates an expected call of ProxyRequestRaw.
-func (mr *MockNoAuthConnectionMockRecorder) ProxyRequestRaw(ctx, reqType, req, w interface{}) *gomock.Call {
-	mr.mock.ctrl.T.Helper()
-	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "ProxyRequestRaw", reflect.TypeOf((*MockNoAuthConnection)(nil).ProxyRequestRaw), ctx, reqType, req, w)
 }
--- a/internal/auth_methods/no_auth/proxy.go
+++ b/internal/auth_methods/no_auth/proxy.go
@@ -2,7 +2,6 @@ package no_auth
 
 import (
 	"context"
-	"net/http"
 
 	apauthcore "github.com/rmorlok/authproxy/internal/apauth/core"
 	"github.com/rmorlok/authproxy/internal/core/iface"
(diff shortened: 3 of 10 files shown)


rmorlok:
Going to look at implementing this instead.

PR #327, opened 2026-05-16
Probe-driven health: per-probe counters + transitions (#255)

### LINKED ISSUE #255: [api-key] Probe-driven health: counters, health_state, transitions

rmorlok:
Closes #255. Part of #243.

## Summary
Generalizes the connection health-state signal across auth methods. With OAuth2 refresh-failure → unhealthy already wired by #265, this PR adds the probe-driven half: periodic probe outcomes track per-probe consecutive counters and transition \`health_state\` on threshold cross. Works uniformly for OAuth2 and api-key.

## Storage
New \`connection_probe_health\` table (sqlite + postgres migrations \`000008\`), one row per \`(connection_id, probe_id)\`. Counters update via \`INSERT ... ON CONFLICT DO UPDATE\` — single statement keeps Postgres out of the \"transaction aborted\" state that a separate INSERT + UPDATE recovery would trigger on PK collision.

DB methods:
- \`RecordProbeSuccess\` — increment successes, reset failures
- \`RecordProbeFailure\` — increment failures, reset successes
- \`GetConnectionProbeHealth\` — single (connection, probe) lookup
- \`ListConnectionProbeHealth\` — all rows for a connection
- \`ResetConnectionProbeHealth\` — zero all counters

## Runtime (\`internal/core/connection_probe_health.go\`)
\`recordPeriodicProbeOutcome\` is called from \`runProbeForConnection\` after every \`probe.Invoke\`.

- **On failure**: if the just-incremented failure count ≥ \`probe.EffectiveFailureThreshold()\`, \`MarkHealthState(unhealthy)\`.
- **On success**: if currently unhealthy AND this probe's success streak ≥ \`recovery_threshold\` AND **no other probe is still at/over its failure_threshold**, \`MarkHealthState(healthy)\` + reset all counters.
- The cross-probe check is the key recovery-correctness rule: a recovery on probe A doesn't restore health while probe B is still failing.

On any success against an **api-key** connection, the active credential's \`last_validated_at\` is stamped — gives operators a per-connection \"credential last checked OK\" timestamp without scraping logs.

Health bookkeeping is best-effort: a counter-write or transition failure is logged at \`task_probe.go\` but does not invalidate the probe outcome itself.

## Interface change
\`iface.Probe\` gains \`EffectiveFailureThreshold()\` / \`EffectiveRecoveryThreshold()\` accessors so the runtime reads thresholds without type-asserting back to \`cschema.Probe\`.

## Test plan
Runtime tests:
- [x] Single failure does NOT flip.
- [x] Nth failure (== threshold) flips unhealthy + emits structured event.
- [x] Subsequent failures while unhealthy are idempotent (no duplicate event).
- [x] Success while healthy is a no-op (no recovery work scheduled).
- [x] Success at \`recovery_threshold\` flips healthy + resets all counters.
- [x] Success below \`recovery_threshold\` does NOT flip.
- [x] Another probe over threshold blocks recovery (load-bearing correctness).
- [x] api-key probe success stamps \`last_validated_at\`.
- [x] OAuth2 probe success skips \`last_validated_at\`.

DB layer:
- [x] First-outcome insert (success and failure paths).
- [x] Increment + reset semantics for both directions.
- [x] Distinct rows per probe + per connection.
- [x] \`ErrNotFound\` on missing row.
- [x] Reset zeros all counters; reset on unknown connection is no-op.

Full module + preflight green on SQLite + Postgres.


internal/database/migrations/postgres/000008_create_connection_probe_health.up.sql
```diff
@@ -0,0 +1,15 @@
+create table connection_probe_health
```
rmorlok: Instead of having a probe health table, let's have a probe outcome table that tracks each outcome as a separate row. Then any transitions against it will be a query against the rows of the table rather than trying to maintain counters directly.

This change will require a database job that cleans up this table. The job would walk the set of connections, and for each connection remove old probe outcomes that are no long needed. It should inspect into the connection's probe definition to understand how many need to be kept for heath tracking purposes and always keep at least that many. Assuming that requirement is met, the job should delete probes older than a configured retention value which should default to 24 hours, which is how often the task would run.

rmorlok: Done in 5fe5ea6. Switched the storage model entirely:

- Dropped \`connection_probe_health\` (counter table).
- Added \`connection_probe_outcomes\` (append-only event log). Each probe invocation writes one row carrying \`outcome\`, \`occurred_at\`, and an optional \`error_message\` for failures.

Runtime (\`recordPeriodicProbeOutcome\`) now inserts an outcome, reads up to N recent rows (N = the relevant threshold), and walks them from the head to count consecutive matching events. No counters to maintain or reset — a success naturally breaks a failure streak just by being the most-recent row.

Daily cleanup job:
- New \`taskTypeProbeOutcomeCleanup\` registered as \`@every 24h\`.
- Walks all non-deleted connections; for each, reads the distinct probe ids that have recorded outcomes; for each \`(connection, probe)\`, keeps at least \`max(failure_threshold, recovery_threshold)\` most-recent rows + deletes anything older than \`DefaultProbeOutcomeRetention\` (24h).
- Probe ids that no longer appear in the connector definition still get pruned using a defaults-derived keep-minimum so stale events can't accumulate forever.
- Retention is parameterizable via the task payload — an operator can enqueue an ad-hoc cleanup with a stricter retention if needed.

Test coverage shifted to match: 8 DB-layer tests (insert / get-recent / keep-min semantics) + 10 runtime tests (streak counting, recovery rules, cross-probe blocking) + 2 cleanup helper tests. All pass on SQLite and Postgres.

Issue #181, opened 2026-04-27
OAuth tests P2: proxy-initiated disconnect / revocation


rmorlok:
Parent: #159
Phase: 3 (P2)
Spec: #159 scenario 26
Blocked by: #162

If supported:

- Proxy calls provider revocation endpoint.
- Local tokens are deleted, disabled, or tombstoned.
- Future proxied calls require reconnect.
- Revocation failure is handled according to product policy.

rmorlok:
## Test approach: go-oauth2-server `/test/*` (preferred)

The primary goal of this scenario is non-interactive (token-endpoint behavior, refresh, revocation, timing, response shape, etc.) — there is no user-flow leg under test. Use the go-oauth2-server `/test/*` control plane to seed state and short-circuit the authorize step rather than booting chromedp.

Reach for chromedp only if the scenario grows to depend on a real consent/login/redirect leg.

PR #328, opened 2026-05-16
Add OAuth2 provider timeout integration tests (#177 scenario 18)

### LINKED ISSUE #177: OAuth tests P1: provider timeouts
Parent: #159
Phase: 2 (P1)
Spec: #159 scenario 18
Blocked by: #162

## Cases

- Authorization endpoint timeout.
- Token endpoint timeout.
- Refresh endpoint timeout.
- Revocation endpoint timeout.
- Upstream API timeout.

## Verify

- Timeout is retried only where safe.
- User receives clear failure after retry exhaustion.
- Connection state remains consistent.

rmorlok:
## Summary

End-to-end coverage for scenario 18 in #177: how the proxy behaves when a provider endpoint hangs (or, more precisely, when the connection terminates mid-flight). The proxy has no HTTP client timeout configured — `httpf.factory` wraps `http.DefaultTransport` without a `Timeout`, and `RefreshTimeout` only feeds the Redis mutex — so the faithful reproduction is `ScriptAction{DropConnection: true}` on the test provider, which hijacks the writer and closes the conn. From the proxy's perspective, a server-side timeout, a TCP reset, and a process crash are all the same transport error on the pending request.

Three tests, one per leg the proxy itself drives:

- **`TestTokenExchangeTimeout_RetriesAndExhausts`** — `/token` POST during the code-exchange leg. Asserts 3 attempts (full `tokenExchangeMaxAttempts` budget), exactly one failure event with `category=network_error` / `attempts=3` / no `provider_status_code`, no token persisted, connection lands in `auth_failed`, and a clean 302 to `return_to_url?setup=pending`.
- **`TestTokenRefreshTimeout_RetriesAndExhausts`** — refresh leg. Same 3-attempt retry shape, plus: connection health **stays healthy** (network_error is transient), 2 retry-warn lines, persisted refresh-token row is byte-preserved (load-bearing — a partial write here would corrupt every future refresh).
- **`TestUpstreamApiTimeout_SurfacedToCaller`** — upstream API leg. Exactly **one** attempt (no retry on the resource leg), no refresh attempted (transport error must not be confused with a 401 → refresh-and-replay), connection state and health unchanged.

Companion `provider_timeouts_test.md` documents the DropConnection-as-timeout rationale, the missing-timeout observation that motivates it, and explicitly out-of-scope cases:

- **Authorize endpoint** — not proxy-driven (the browser hits it), so no proxy code path exists to test.
- **Revocation endpoint** — runs in the background `disconnect_connection` Asynq task; coverage belongs to the P2 disconnect tests in #181.

Closes #177.

## Test plan

- [x] `go test -tags=integration -run 'TestTokenExchangeTimeout_RetriesAndExhausts|TestTokenRefreshTimeout_RetriesAndExhausts|TestUpstreamApiTimeout_SurfacedToCaller' ./oauth2/...` — all 3 pass
- [x] `./scripts/preflight.sh` — passes

### DIFF
--- a/integration_tests/oauth2/provider_timeouts_test.go
+++ b/integration_tests/oauth2/provider_timeouts_test.go
@@ -1,0 +1,264 @@
+//go:build integration
+
+package oauth2
+
+import (
+	"net/url"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/rmorlok/authproxy/integration_tests/helpers"
+	"github.com/rmorlok/authproxy/internal/database"
+	cschema "github.com/rmorlok/authproxy/internal/schema/connectors"
+)
+
+// Scenario 18 from issue #177: the proxy's HTTP client has no configured
+// timeout (httpf.factory uses http.DefaultTransport as-is). The closest
+// observable failure mode is a connection that terminates mid-flight —
+// from the proxy's perspective a server-side timeout, a forced TCP reset,
+// and a process crash all surface identically as a transport error on the
+// pending POST/GET. The test provider's ScriptAction{DropConnection:true}
+// reproduces this by hijacking the response writer and closing the
+// underlying connection.
+//
+// The proxy classifies any transport-layer failure on the token-exchange
+// or refresh leg as `network_error` (transient). The token endpoint and
+// refresh endpoint both retry transport errors through their shared
+// retry helpers (`postTokenExchangeWithRetry`, `postRefreshWithRetry`)
+// up to {tokenExchange,tokenRefresh}MaxAttempts. The upstream-API leg
+// (`ProxyRequest` → `sendProxyRequest`) does NOT retry on transport
+// errors — the 401-retry-after-refresh path is the only retry on the
+// proxy leg and it keys on the HTTP status, which doesn't exist when
+// the connection dies.
+//
+// What is NOT covered here (documented in provider_timeouts_test.md):
+//
+//   - **Authorize endpoint timeouts.** The proxy never POSTs to the
+//     authorize endpoint — the browser hits it directly. A timeout
+//     there manifests as a UI failure, not a proxy failure.
+//   - **Revocation endpoint timeouts.** Revocation runs in the
+//     background via the disconnect_connection Asynq task. Coverage
+//     for the disconnect path lives with the P2 disconnect tests
+//     (#181); folding it in here would require standing up a worker
+//     in the integration env.
+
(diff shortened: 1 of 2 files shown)


integration_tests/oauth2/provider_timeouts_test.go
```diff
@@ -0,0 +1,264 @@
...
+	"github.com/rmorlok/authproxy/internal/database"
+	cschema "github.com/rmorlok/authproxy/internal/schema/connectors"
+)
+
+// Scenario 18 from issue #177: the proxy's HTTP client has no configured
+// timeout (httpf.factory uses http.DefaultTransport as-is). The closest
+// observable failure mode is a connection that terminates mid-flight —
+// from the proxy's perspective a server-side timeout, a forced TCP reset,
+// and a process crash all surface identically as a transport error on the
+// pending POST/GET. The test provider's ScriptAction{DropConnection:true}
```
rmorlok: Are you saying there is a gap in our logic in that we don't have timeouts configured, or are you saying that we inherit timeouts from the incoming request based on context cancellation?

rmorlok: Both, with the gap being the bigger piece. Walked the code:

- `httpf.factory` wraps `http.DefaultTransport` and never sets `Timeout` on the client (`internal/httpf/factory.go:210`).
- The inbound `http.Server` is built with no `ReadTimeout` / `WriteTimeout` / `IdleTimeout` (`internal/schema/config/service_common.go:152-156`).
- Proxy/callback handlers never wrap their `ctx` in `context.WithTimeout` (only `/healthz` does — `internal/service/api/server.go:78`).

So there is no time-budgeted deadline anywhere — if the calling client stays connected, the proxy will wait on a hung provider indefinitely.

What we *do* inherit is ctx cancellation if the *calling client* closes its connection mid-request: the refresh, proxy, and revocation legs all thread `ctx` through `gentleman.UseContext(ctx)` (`proxy.go:302,349`, `revocation.go:77,143`), so the in-flight upstream request aborts in that case. The token-exchange leg is the exception — `postTokenExchangeWithRetry` (`callback.go:259`) does not call `UseContext` on the request itself; only the retry backoff `select` checks `ctx.Done()` between attempts, so an attempt already in flight is not cancellable.

That's why this scenario uses `DropConnection` rather than a server-side sleep: a sleep is unobservable to the proxy under the current configuration. I updated the file header in 89fe3ca to spell this out instead of just saying "no timeout."

rmorlok: Ok, that's reasonable. The expected behavior is the proxy stays live as long as the client initiating the proxy request stays live.

PR #343, opened 2026-05-19
Implement PKCE (RFC 7636) for OAuth2 authorization-code flow

### LINKED ISSUE #335: Implement PKCE (RFC 7636) for OAuth2 authorization-code flow
Parent project: #159
Blocks: #174 (PKCE validation tests)
Spec: RFC 7636

## Goal

Add Proof Key for Code Exchange (PKCE) support to the OAuth2 authorization-code flow in `internal/auth_methods/oauth2/`. Currently the proxy emits no `code_challenge` on authorize and sends no `code_verifier` on token exchange, so any provider that requires PKCE is unconnectable through AuthProxy. Test ticket #174 cannot move until this lands.

### 1. Connector schema

`internal/schema/connectors/auth_oauth2_authorization.go` — add an optional PKCE block:

```go
type AuthOauth2Authorization struct {
    Endpoint       string
    QueryOverrides map[string]string
    PKCE           *AuthOauth2PKCE  // nil = PKCE disabled (current behavior)
}

type AuthOauth2PKCE struct {
    Method string  // \"S256\" (default/required) or \"plain\" (discouraged, behind a flag)
}
```

Validate `Method` in `internal/schema/config` loader — reject anything other than `S256` or `plain`. Default to `S256` when the block is present without a method.

### 2. Verifier generation + persistence

`internal/auth_methods/oauth2/state.go` — extend the state struct with `CodeVerifier string` (and a method enum). Verifier is a per-flow secret with the same lifetime as `state_id` — same Redis TTL, same encryption-at-rest treatment as the rest of the state record.

Verifier generation (new helper): 43 chars from the RFC 7636 unreserved alphabet (`A-Z a-z 0-9 - . _ ~`) via `crypto/rand`. 43 chars of that alphabet is ~256 bits — well past the 128-bit floor RFC 7636 §4.1 sets.

### 3. Authorize URL

`internal/auth_methods/oauth2/auth_url.go:96-103` — when `o.auth.Authorization.PKCE != nil`:

```go
query.Set(\"code_challenge\", challenge)
query.Set(\"code_challenge_method\", method)  // \"S256\" or \"plain\"
```

For `S256`: `challenge = base64url(sha256(verifier))` (no padding, per RFC 7636 §4.2). For `plain`: `challenge = verifier`.

The verifier is generated *before* `saveStateToRedis` and persisted into the state record in the same write. No second round-trip to Redis.

### 4. Token exchange

`internal/auth_methods/oauth2/callback.go` — in the token-exchange path, when the loaded state record has a non-empty verifier, add `code_verifier` to the POST form before calling `postTokenExchangeWithRetry`.

PKCE applies to the **initial** authorization-code exchange only. Refresh-token grants (`task_refresh_oauth_token.go`) and revocation must remain unchanged — RFC 7636 §6 is explicit that PKCE is not used on refresh.

### 5. Error surfaces

The token endpoint can return PKCE-specific errors:
- `invalid_grant` — verifier doesn't match the original challenge (or no verifier sent when one was expected). The existing `tokenExchangeInvalidGrant` category already covers this — we don't need a new category.
- `invalid_request` — `code_verifier` malformed.

Both already map cleanly onto existing categories in `token_exchange_failure.go`. No new dashboards needed.

## Acceptance Criteria

- A connector YAML declaring `pkce: { method: S256 }` produces an authorize URL with `code_challenge` and `code_challenge_method=S256`.
- The token-exchange POST for that connector includes `code_verifier` matching the persisted state record.
- Refresh-token grants and revocation requests do NOT include `code_verifier` or `code_challenge`.
- A connector without a `pkce` block behaves exactly as today (no `code_challenge`, no `code_verifier`) — fully backwards-compatible.
- Schema validator rejects `method: foo`.
- Existing tests (`TestStandardAuthorizationCodeFlow`, etc.) still pass — PKCE is opt-in per connector.
- Unblocks #174 — the test helper's existing `RequirePKCE`/`SkipPKCECheck` knobs on `oauth2_provider.go` already give the surface area #174 needs.

## Out of Scope

- Migrating existing connectors to require PKCE. That's a per-connector decision and follows on issues — this ticket only adds the mechanism.
- PKCE on refresh (explicitly forbidden by RFC 7636 §6).
- Authorization code binding to a specific client beyond what PKCE already provides — orthogonal hardening (e.g. DPoP, mTLS).

## Notes

The choice to make PKCE per-connector (not always-on) is deliberate: not every provider supports PKCE, and turning it on for one that doesn't would silently break the connector with a confusing `invalid_request` from the token endpoint. Once we've audited the connector catalog, we can flip the default to PKCE-on and let connectors opt out explicitly.

Implementation should land in a single PR — the changes are coupled (schema field is unused until authorize emits it; authorize emit is broken until callback sends verifier). Splitting would require flag-gating that's not worth the churn for a feature this contained.

rmorlok:
## Summary
- Adds RFC 7636 PKCE support to the OAuth2 authorization-code flow, gated by a new optional `pkce` block on each connector. Connectors without the block behave exactly as before.
- Mints a 43-char crypto/rand verifier from the RFC 7636 unreserved alphabet (~256 bits of entropy, well past the 128-bit floor) and persists it alongside the rest of the OAuth state record — same TTL, same AES-GCM envelope.
- Emits `code_challenge` and `code_challenge_method` on the authorize URL, and sends `code_verifier` on the **initial** token exchange only. Refresh-token grants and revocation are untouched (RFC 7636 §6).

## Files
- `internal/schema/connectors/auth_oauth2_authorization.go` — new `AuthOauth2PKCE` type and `PKCE` field; `PKCEMethodS256` (default) / `PKCEMethodPlain` constants.
- `internal/schema/connectors/auth_oauth2.go` — `AuthOAuth2.Validate` rejects unknown PKCE methods; `Clone` carries the block.
- `internal/schema/config/reexport.go` — re-exports the new types/constants.
- `internal/auth_methods/oauth2/pkce.go` (new) — verifier generator (uniform rejection-sampling) and `pkceChallengeFor` (S256 / plain).
- `internal/auth_methods/oauth2/state.go` — extended state struct with `PKCECodeVerifier` and `PKCEMethod`; `saveStateToRedis` mints the verifier in the same write when PKCE is configured.
- `internal/auth_methods/oauth2/auth_url.go` — emits the challenge params when PKCE is on.
- `internal/auth_methods/oauth2/callback.go` — adds `code_verifier` to the token-exchange form when state carries one.
- `internal/auth_methods/oauth2/pkce_test.go` (new) — 14 unit tests.

## Test plan
- [x] `go test ./internal/auth_methods/oauth2/... ./internal/schema/... -count=1`
- [x] `./scripts/preflight.sh`
- [x] RFC 7636 Appendix B vector — `pkceChallengeFor(S256, "dBjftJeZ4CVP-mB92K27uhbUJU1p1r_wW1gFWFOEjXk")` produces `"E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM"`.
- [x] Authorize URL emits `code_challenge` / `code_challenge_method=S256` when PKCE is on; omits both when off.
- [x] Token exchange sends `code_verifier` matching the persisted state; omits it when no verifier was persisted.

Closes #335

### DIFF
--- a/internal/schema/connectors/auth_oauth2_authorization.go
+++ b/internal/schema/connectors/auth_oauth2_authorization.go
@@ -1,6 +1,48 @@
 package connectors
 
+// PKCEMethod selects the code-challenge transformation used by the
+// connector's authorize/token-exchange pair. Values follow RFC 7636 §4.2.
+type PKCEMethod string
+
+const (
+	// PKCEMethodS256 is the recommended challenge method:
+	// challenge = base64url(sha256(verifier)). Always use this unless the
+	// 3rd-party explicitly rejects it.
+	PKCEMethodS256 = PKCEMethod("S256")
+	// PKCEMethodPlain is the fallback method: challenge = verifier. RFC 7636
+	// §4.2 marks plain as discouraged — it removes the cryptographic
+	// hop that protects the verifier on a leaky TLS terminator. Configure
+	// only when a provider cannot validate S256.
+	PKCEMethodPlain = PKCEMethod("plain")
+)
+
+// AuthOauth2PKCE turns on RFC 7636 Proof Key for Code Exchange for a
+// connector's authorization-code flow. nil disables PKCE (current default
+// behavior); a non-nil block enables it with the chosen method.
+type AuthOauth2PKCE struct {
+	// Method is the challenge transformation. Defaults to S256 when unset.
+	Method PKCEMethod `json:"method,omitempty" yaml:"method,omitempty"`
+}
+
+// GetMethodOrDefault returns the configured method, defaulting to S256 if
+// the block is present but the method was omitted.
+func (p *AuthOauth2PKCE) GetMethodOrDefault() PKCEMethod {
+	if p == nil || p.Method == "" {
+		return PKCEMethodS256
+	}
+	return p.Method
+}
+
+func (p *AuthOauth2PKCE) Clone() *AuthOauth2PKCE {
+	if p == nil {
+		return nil
+	}
+	clone := *p
+	return &clone
+}
+
 type AuthOauth2Authorization struct {
 	Endpoint       string            `json:"endpoint" yaml:"endpoint"`
 	QueryOverrides map[string]string `json:"query_overrides,omitempty" yaml:"query_overrides,omitempty"`
+	PKCE           *AuthOauth2PKCE   `json:"pkce,omitempty" yaml:"pkce,omitempty"`
 }
--- a/internal/auth_methods/oauth2/auth_url.go
+++ b/internal/auth_methods/oauth2/auth_url.go
@@ -102,6 +102,16 @@ func (o *oAuth2Connection) GenerateAuthUrl(ctx context.Context, actor IActorData
 	query.Set("scope", strings.Join(scopes, " "))
 	query.Set("state", o.state.Id.String())
 
+	if o.auth.Authorization.PKCE != nil {
+		method := o.auth.Authorization.PKCE.GetMethodOrDefault()
+		challenge, err := pkceChallengeFor(method, o.state.PKCECodeVerifier)
+		if err != nil {
+			return "", fmt.Errorf("failed to compute pkce challenge for connector %s: %w", cv.GetId(), err)
+		}
+		query.Set("code_challenge", challenge)
+		query.Set("code_challenge_method", string(method))
+	}
+
 	for k, v := range o.auth.Authorization.QueryOverrides {
 		rendered, err := o.renderMustache(ctx, v)
 		if err != nil {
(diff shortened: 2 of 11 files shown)


internal/schema/connectors/auth_oauth2_authorization.go
```diff
@@ -1,6 +1,48 @@
...
+		return nil
+	}
+	clone := *p
+	return &clone
+}
+
 type AuthOauth2Authorization struct {
 	Endpoint       string            `json:"endpoint" yaml:"endpoint"`
 	QueryOverrides map[string]string `json:"query_overrides,omitempty" yaml:"query_overrides,omitempty"`
+	PKCE           *AuthOauth2PKCE   `json:"pkce,omitempty" yaml:"pkce,omitempty"`
```
rmorlok, line 47: You missed updating the json schema for this change.

rmorlok: Good catch — added the PKCE def (method enum: S256, plain) under internal/schema/connectors/schema-oauth.json and referenced it from Authorization. Also added two test_data fixtures (valid + invalid method) so Test_SchemaAgainstRealData exercises both branches. Fix in 34cb6a8.

PR #344, opened 2026-05-19
[api-key] Unified reauth flow (manual + probe-driven) (#256)

### LINKED ISSUE #256: [api-key] Unified reauth flow (manual + probe-driven)
Part of #243 — see that issue, Part 6.

## Scope

Unified UX for re-authenticating a connection. Generalized across api-key and OAuth2 — same UI button, different backend.

### Triggers

- **Manual**: UI exposes "Re-authenticate" on any `Ready` connection.
- **Automatic**: when a connection's `health_state` flips to `unhealthy`, the same action is surfaced as actionable in the UI.

### api-key path

1. Reset the connection's `setup_step` to the credential-collection step.
2. Form payload is presented **with no prior values pre-filled**. The active `api_key_credentials` row is never decrypted into a form payload, never sent to the UI, and never logged. (This is the "no replay" requirement from #243.)
3. On submit: insert a new `api_key_credentials` row, soft-delete the prior, run verify probes against the new credential.
4. On verify success: flip `health_state` back to `healthy`, record `last_rotated_at` and `rotated_by_actor_id`.

### OAuth2 path

The unified action hooks into the existing OAuth2 reinitiate flow (`service_connection_retry.go` or equivalent). OAuth2 already does not replay credentials (refresh-token replacement is internal). No new no-replay enforcement needed for OAuth2 — but include the UI wiring so the same button works.

### "No replay" enforcement

Add a focused test asserting that after rotation begins, the prior key bytes never appear in any UI/API response or log line for that connection.

## Test plan

- [ ] Manual reauth on a healthy api-key connection inserts a new credential row, soft-deletes prior
- [ ] Auto-reauth surfaces the action when probes flip the connection to `unhealthy`
- [ ] No-replay: form payload contains no prior credential bytes
- [ ] OAuth2 reauth via the unified action triggers the existing OAuth2 path
- [ ] `last_rotated_at` and `rotated_by_actor_id` recorded correctly
- [ ] Successful reauth flips `health_state` from `unhealthy` back to `healthy`

## Dependencies

- Connection-initiate-flow sub-issue
- Proxy sub-issue
- Probe-runtime sub-issue

rmorlok:
Closes #256. Part of #243.

## Summary

Adds a unified re-authentication flow against existing Ready connections — works the same UX-wise for api-key and OAuth2. Used for user-driven credential rotation **and** as the recovery action when probes flip a connection unhealthy.

## Backend

**New endpoint** — `POST /connections/{id}/_reauth`:
- Dispatches by auth type. **api-key** → reset `setup_step` to `credentials:0`, return the credentials form. **OAuth2** → preconnect:0 if defined, else re-initiate the OAuth redirect.
- Distinct from `/_retry` (which requires a terminal-failure state). Reauth applies to any Ready connection regardless of health.
- Connection lifecycle State stays Ready throughout; only `setup_step` is reset and re-driven.
- Permission gate: `ForVerbs("create", "update")`.

**Rotation semantics — no schema changes needed.** `InsertApiKeyCredential` already soft-deletes the active row and inserts the new one in the same transaction. The audit trail (`created_at`, `created_by_actor_id`) on the new row is exactly the rotation event. The submit handler picks up reauth submissions naturally — no new code paths.

**No-replay invariant.** Form payloads are built from the connector schema only. The active `api_key_credentials` row is never decrypted into a UI/API response. Asserted by a focused test.

**Health flip on verify pass.** `onVerifyPassed` now calls `MarkHealthState(healthy, "verify_passed")` after probes succeed. Idempotent on initial setup (already healthy); flips back to healthy on reauth-driven verify after a successful credential rotation.

**API surface** — `ConnectionJson.health_state` exposed (Go + Swagger + JS SDK type + enum).

## UI

**Marketplace** (`ui/marketplace`):
- ConnectionCard surfaces health: state chip turns warning-orange with a warning icon when unhealthy.
- New \"Re-authenticate\" button on Ready connections — filled+orange on unhealthy, subtle text button on healthy.
- Reauth click dispatches `reauthConnectionAsync`; setup-flow rendering reuses the existing form/redirect plumbing — no new dialogs.

**Admin** (`ui/admin`):
- ConnectionDetail surfaces an \"Unhealthy\" chip next to the state chip when relevant.

### Screenshots

**Healthy (subtle CTA):**

marketplace-healthy

**Unhealthy (state chip + prominent CTA):**

marketplace-unhealthy

## Test plan

- [x] Rejects non-Ready connection
- [x] api-key Ready returns credentials form with no prior credential bytes in payload (no-replay)
- [x] Reauth works on unhealthy api-key (same form path)
- [x] Reauth clears any prior setup error
- [x] OAuth2 with preconnect resets to preconnect:0
- [x] no-auth connector rejected
- [x] `onVerifyPassed` flips unhealthy → healthy via `MarkHealthState`
- [x] Existing TestOnVerifyPassed cases unchanged (idempotent when already healthy)
- [x] Marketplace tests + storybook stories updated (`Connected` + `Unhealthy` variants)
- [x] Full preflight + `go test ./...` green on SQLite + Postgres


internal/core/service_connection_reauth.go
```diff
@@ -0,0 +1,76 @@
...
+	switch connector.Auth.Inner().(type) {
+	case *config.AuthApiKey:
+		if !connector.SetupFlow.HasCredentials() {
+			return nil, httperr.InternalServerErrorMsg("api-key connector has no credentials step")
+		}
+		first := cschema.MustNewIndexedSetupStep(cschema.SetupPhaseCredentials, 0)
+		return conn.buildFormResponse(ctx, first, connector.SetupFlow)
+	case *config.AuthOAuth2:
+		if connector.SetupFlow.HasPreconnect() {
+			first := cschema.MustNewIndexedSetupStep(cschema.SetupPhasePreconnect, 0)
```
rmorlok, line 69: Why does the OAuth2 connection type go to the preconnect stage instead of going directly through the auth process? I'm not saying it's wrong, I'm just trying to undersand.

rmorlok: Preconnect steps collect inputs that the OAuth2 endpoints depend on at render time — `authorization.endpoint`, `token.endpoint`, `query_overrides`, and `form_overrides` are Mustache templates whose only allowed variables are preconnect fields (see `ValidateMustacheReferences` in `internal/schema/connectors/auth_oauth2.go`). Typical examples: a tenant subdomain (`https://{{preconnect.subdomain}}.example.com/oauth/authorize`), a region, or an instance hostname.

So if the connector has preconnect, we can't construct the authorization URL without those values. Restarting at `preconnect:0` lets the user confirm or update them — useful if they're reauthing because the tenant/instance moved, but harmless if nothing changed since the fields come pre-populated from the saved configuration.

This also mirrors `RetryConnectionSetup` exactly, so the OAuth2 dispatch is consistent between the two recovery paths. If preconnect isn't declared, we skip straight to `initiateAuthStep` and go directly into the redirect — which is the "directly through the auth process" branch you were thinking of.

PR #347, opened 2026-05-20
CLI streaming proxy: rename raw-proxy, add ap proxy + ap curl (#332)

### LINKED ISSUE #332: Streaming: CLI — rename raw-proxy, add ap proxy + ap curl
Part of #319. Client-side ergonomics for the streaming raw proxy.

Depends on the streaming endpoint sub-ticket.

### 1. Rename \`ap raw-proxy\` → \`ap signing-proxy\`

The existing \`ap raw-proxy\` is a JWT-signing reverse proxy for hitting the AuthProxy admin/api itself (development tool). The name was always misleading; with the new streaming feature the name confusion gets worse. Rename to \`signing-proxy\`. Keep \`raw-proxy\` as a hidden alias for one release, with a deprecation warning printed to stderr.

### 2. New \`ap proxy --connection <id>\`

A local reverse proxy that forwards through the new \`/api/v1/connections/{id}/_proxy_raw\` endpoint. Streams bytes both directions.

Two ways the caller can specify the upstream:

- **Path-based**: launch with \`--upstream-base https://api.openai.com\`. Then any request to \`http://127.0.0.1:9999/v1/chat/completions\` gets the upstream URL derived as \`https://api.openai.com/v1/chat/completions\` and the \`X-AuthProxy-Upstream-URL\` header set automatically.
- **Header-based**: caller sets \`X-AuthProxy-Upstream-URL\` directly on the inbound request; \`ap proxy\` forwards it through unchanged.

Flags:
- \`--connection <id>\` (required)
- \`--upstream-base <url>\` (optional)
- \`--port <n>\` (default 9999)
- \`--server <url>\` for the AuthProxy api endpoint (defaults from \`~/.authproxy.yaml\`)

Uses the same signer path as the existing signing-proxy.

### 3. New \`ap curl <curl args>\`

Same CLI surface as the real \`curl\` binary (URL + flags). Internally:
1. Boots an \`ap proxy\` listener on an ephemeral port.
2. Rewrites the user-supplied URL: keeps the path/query, swaps the host:port for the listener.
3. Either shells out to real \`curl\` against the rewritten URL OR makes the HTTP call itself with Go's \`http.Client\` + streams to stdout.
4. Tears down the listener on exit.

Required flag: \`--connection <id>\` (must come first / be detectable so we don't confuse it with curl's own flags).

\`ap curl\` is the one-shot path; \`ap proxy\` is the long-running path. Same machinery underneath.

## Acceptance

- \`ap signing-proxy\` works exactly like the old \`ap raw-proxy\` did.
- \`ap raw-proxy\` still works but prints a one-line deprecation notice to stderr.
- \`ap proxy --connection conn_abc --upstream-base https://example.com\` runs as a local reverse proxy; \`curl http://127.0.0.1:9999/foo\` proxies through to \`https://example.com/foo\`.
- \`ap proxy --connection conn_abc\` (no \`--upstream-base\`) requires the caller to set \`X-AuthProxy-Upstream-URL\` on the inbound request.
- \`ap curl --connection conn_abc https://api.openai.com/v1/chat/completions\` returns the streamed response on stdout, byte-for-byte.
- Body streaming verified manually: a chunked response from a local httptest server is visible on stdout as it arrives, not after the whole thing is collected.

rmorlok:
Closes #332. Part of #319.

Renames `ap raw-proxy` to `ap signing-proxy` (the name was always misleading — it's a JWT-signing reverse proxy for the AuthProxy server itself, not a raw passthrough). The old name keeps working for one release as a hidden alias that prints a one-line deprecation warning to stderr.

Adds one new connection-scoped command (`ap proxy`) with three modes, discriminated by the first positional arg:

- **Long-running listener:** `ap proxy --connection cxn_x [--upstream-base https://...]`
  Boots a local reverse proxy on `--port` (default 9999). Each inbound request is forwarded through `/_proxy_raw` with bodies streaming both directions. Hop-by-hop headers (RFC 7230 §6.1) are stripped on both legs. Either `--upstream-base` derives `X-AuthProxy-Upstream-URL` from the inbound path+query, or the caller supplies the header directly.

- **One-shot through curl:** `ap proxy --connection cxn_x curl https://api.example.com/v1/things -X POST -d @body.json`

- **One-shot through wget:** `ap proxy --connection cxn_x wget https://api.example.com/files/big.bin -O out.bin`

  Both one-shot modes boot an ephemeral-port listener, derive `--upstream-base` from the URL's scheme+host, rewrite the URL to point at the listener, and shell out to the named tool. All `ap proxy` flags must come before the literal `curl`/`wget`; everything after is forwarded to the tool verbatim. Cobra's `SetInterspersed(false)` keeps each tool's flag surface (including clashes like curl's `--config`) from leaking into our parser.

Adds [docs/cli.md](https://github.com/rmorlok/authproxy/blob/cli-streaming-proxy/docs/cli.md) — full CLI reference covering config file shape, RS256 vs HS256 signing keys, every command's flags and intent, plus recipes. Linked from the README TOC + Documentation table and the `docs/` index.

## Tests
- `cmd/cli/proxy_test.go` covers: upstream derivation from base, header-supplied upstream, missing-upstream 502, hop-by-hop stripping, SSE-shaped chunked-streaming round-trip (fails if buffering is introduced), URL detection across positional / `--url <v>` / `--url=v`, and cobra flag-boundary checks for both `curl <args>` and `wget <args>` (proves `--connection cxn_x curl -X POST …` and `--connection cxn_x wget -O out https://… --user-agent ap` parse correctly).

## Test plan
- [x] `go test ./cmd/cli/...` passes
- [x] `./scripts/preflight.sh` passes
- [x] `go run ./cmd/cli --help` shows `signing-proxy`, `proxy`; `raw-proxy` is hidden
- [x] `go run ./cmd/cli proxy --help` documents long-running + curl + wget modes
- [x] `go run ./cmd/cli raw-proxy --proxyTo=api` still runs and prints the deprecation notice
- [ ] Manual smoke against a running dev server: long-running, `… curl …`, `… wget …`


cmd/cli/proxy_test.go
```diff
@@ -0,0 +1,268 @@
...
+	var ran []string
+	cmd := cmdProxy()
+	cmd.RunE = func(c *cobra.Command, args []string) error { ran = args; return nil }
+	cmd.SetArgs([]string{"--connection", "cxn_x", "--upstream-base", "https://api.example.com"})
+	require.NoError(t, cmd.Execute())
+	assert.Empty(t, ran, "no positional args means listener mode")
+}
+
+// cmdRawProxyAlias must reuse cmdSigningProxy's flag surface so the old
+// command keeps working byte-for-byte. The deprecation warning lives in
```
rmorlok: We don't need to be backwards compatible.

cmd/cli/main.go
```diff
@@ -17,7 +17,9 @@ func main() {
 	rootCmd.AddCommand(cmdList())
 	rootCmd.AddCommand(cmdSignJwt())
 	rootCmd.AddCommand(cmdVerifyJwt())
-	rootCmd.AddCommand(cmdRawProxy())
+	rootCmd.AddCommand(cmdSigningProxy())
+	rootCmd.AddCommand(cmdRawProxyAlias())
```
rmorlok: We don't need to be backwards compatible, let's just get rid of this now

cmd/cli/proxy_test.go
rmorlok: Done — alias removed in 444653f. cmdRawProxyAlias factory, the test, and the docs mention all gone.

cmd/cli/main.go
rmorlok: Done — removed in 444653f. cmdRawProxyAlias registration dropped from main.go, the factory deleted from signing_proxy.go, alias test and docs reference cleaned up.

Issue #351, opened 2026-05-23
Implement OAuth2 token-endpoint client auth method support (basic, post, none)


rmorlok:
Parent project: #159
Blocks: #178 (auth method compatibility tests)
Spec: RFC 6749 §2.3.1 (client_secret_basic / client_secret_post), RFC 6749 §2.1 / RFC 7591 §2 (public clients — `none`)

## Goal

Add explicit token-endpoint client authentication method support to the OAuth2 connector. Today the proxy hard-codes `client_secret_post` style: every token-endpoint POST (initial code-for-token and refresh) carries `client_id` + `client_secret` in the form body, with no path that uses HTTP Basic auth and no path that omits the secret entirely. This means:

- Connectors against providers that require / strongly prefer `client_secret_basic` (e.g., some Google APIs, Salesforce, Okta) cannot be used through the proxy. RFC 6749 §2.3.1 lists Basic as the canonical method ("The authorization server MUST support the HTTP Basic authentication scheme") and many providers reject the post-body form.
- Public clients (mobile / SPA / first-party native apps that legitimately have no `client_secret`, paired with PKCE for proof-of-possession) cannot be modeled. RFC 7591 §2 defines `none` for this case; the proxy's `ClientSecret.GetValue` always fires, so a connector without a secret fails before the form is built.

Test ticket #178 (token-endpoint auth method compatibility) cannot move on the `client_secret_basic` and `none` cases until this lands.

### 1. Connector schema

`internal/schema/connectors/auth_oauth2.go` — add an optional method selector on the top-level `AuthOAuth2` struct:

```go
type AuthOAuth2 struct {
    Type                    AuthType
    TokenEndpointAuthMethod TokenEndpointAuthMethod  // new — defaults to "post"
    ClientId                *common.StringValue
    ClientSecret            *common.StringValue
    // ...rest unchanged
}

type TokenEndpointAuthMethod string

const (
    TokenEndpointAuthClientSecretBasic = TokenEndpointAuthMethod("client_secret_basic")
    TokenEndpointAuthClientSecretPost  = TokenEndpointAuthMethod("client_secret_post")
    TokenEndpointAuthNone              = TokenEndpointAuthMethod("none")  // public client (PKCE-only)
)
```

Method name follows RFC 7591 §2 (`token_endpoint_auth_method`). Default-when-unset is `client_secret_post` to preserve current behavior for every existing connector.

Validator (`AuthOAuth2.Validate`):
- Reject unknown methods.
[...]

Issue #178, opened 2026-04-27
OAuth tests P1: provider auth method compatibility


rmorlok:
Parent: #159
Phase: 2 (P1)
Spec: #159 scenario 19
Blocked by: #162

## Cases

Test provider token-endpoint authentication styles:

- Client secret basic auth.
- Client credentials in POST body.
- Public client with PKCE.
- Invalid auth method for provider.

## Verify

- Configured auth method succeeds.
- Incorrect auth method fails clearly.
- Client secrets are not leaked.

rmorlok:
## Test approach: go-oauth2-server `/test/*` (preferred)

The primary goal of this scenario is non-interactive (token-endpoint behavior, refresh, revocation, timing, response shape, etc.) — there is no user-flow leg under test. Use the go-oauth2-server `/test/*` control plane to seed state and short-circuit the authorize step rather than booting chromedp.

Reach for chromedp only if the scenario grows to depend on a real consent/login/redirect leg.

PR #355, opened 2026-05-23
Implement OAuth2 token-endpoint client auth method support (#351)

rmorlok:
## Summary

Implements #351 — adds the RFC 7591 §2 `token_endpoint_auth_method` selector to the OAuth2 connector. Three methods are supported:

- **`client_secret_post`** (default; today's behaviour): `client_id` and `client_secret` in the form body.
- **`client_secret_basic`**: `Authorization: Basic <base64(qs(id):qs(secret))>` per RFC 6749 §2.3.1, with credentials removed from the form body.
- **`none`**: public clients (RFC 7591 §2) — `client_id` only, no secret, paired with mandatory PKCE so the exchange retains proof-of-possession.

A single helper (`applyTokenEndpointClientAuth`) renders the body+header shape per method; callback (initial code exchange) and refresh both call through it, so the wire format cannot drift between the two grants.

Validator enforces:
- Unknown methods rejected.
- `client_secret` required for `post`/`basic`.
- `client_secret` forbidden + PKCE required for `none`.

Default behaviour is unchanged — every existing connector takes the `client_secret_post` branch.

Unblocks #178.

## Note on redaction

The ticket asks to verify the `Authorization: Basic` header is redacted symmetrically with the `client_secret` form field. The request-log already protects credential material **structurally**, not via a field-level allowlist:

- `LogRecord` (the searchable / indexed record) carries no headers or body — only safe metadata. This is what shows up in dashboards and search results.
- `FullLog` (the full-payload capture, optionally written to blob storage) carries the full request including all headers and body verbatim.

For `client_secret_post`, the secret lives in `FullLog.Request.Body`. For `client_secret_basic`, it lives in `FullLog.Request.Headers["Authorization"]`. Both have identical exposure (FullStore only); neither leaks via `LogRecord`. No code change required.

## Test plan

- [x] Unit tests for `applyTokenEndpointClientAuth` covering all three methods, default-when-empty, special-char encoding in basic auth, and unknown-method rejection.
- [x] Wire-format tests for the callback path with each method (gock-captured body + Authorization header).
- [x] Validator tests for each invariant: basic-needs-secret, none-rejects-secret, none-requires-PKCE, none-with-PKCE accepted, unknown-method rejected.
- [x] JSON schema fixtures: `valid-oauth-connector-basic.yaml`, `valid-oauth-connector-public.yaml` (PKCE+none), `invalid-oauth-connector-auth-method.yaml`.
- [x] Full `./scripts/preflight.sh` clean.
- [x] Full `go test ./internal/...` clean — no regressions in callback / refresh / retry suites.

### DIFF
--- a/internal/auth_methods/oauth2/client_auth.go
+++ b/internal/auth_methods/oauth2/client_auth.go
@@ -1,0 +1,87 @@
+package oauth2
+
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+	"net/url"
+
+	sconfig "github.com/rmorlok/authproxy/internal/schema/config"
+)
+
+// resolveClientCredentials reads the connector's client_id and (when present)
+// client_secret. For token_endpoint_auth_method == "none" the secret is
+// optional — public clients have no secret to send — so its absence is not
+// an error.
+func (o *oAuth2Connection) resolveClientCredentials(ctx context.Context) (string, string, error) {
+	clientId, err := o.auth.ClientId.GetValue(ctx)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to get client id for connector: %w", err)
+	}
+
+	if o.auth.GetTokenEndpointAuthMethodOrDefault() == sconfig.TokenEndpointAuthNone {
+		return clientId, "", nil
+	}
+
+	clientSecret, err := o.auth.ClientSecret.GetValue(ctx)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to get client secret for connector: %w", err)
+	}
+	return clientId, clientSecret, nil
+}
+
+// applyTokenEndpointClientAuth attaches client credentials to the
+// token-endpoint request per the connector's configured
+// token_endpoint_auth_method (RFC 6749 §2.3.1, RFC 7591 §2).
+//
+// Returns the (possibly augmented) form values that should be sent in the
+// request body, and the value of the Authorization header to set on the
+// request (empty string when no header is required).
+//
+// Method semantics:
+//   - client_secret_post (default): client_id + client_secret go in the form
+//     body, no Authorization header.
+//   - client_secret_basic: Authorization: Basic base64(qs(id):qs(secret)) is
+//     set; neither client_id nor client_secret goes in the form body.
+//     RFC 6749 §2.3.1 forbids sending the credentials in two places.
+//   - none: client_id only in the form body, no client_secret, no header.
+//     The token endpoint relies on PKCE for proof-of-possession.
+//
+// The values map is treated as input-only and not mutated; the returned
+// url.Values is a copy with the appropriate fields set.
+func applyTokenEndpointClientAuth(
+	method sconfig.TokenEndpointAuthMethod,
+	clientId, clientSecret string,
+	values url.Values,
+) (url.Values, string, error) {
+	// Effective default — empty method behaves as client_secret_post for
+	// backwards compatibility.
+	if method == "" {
+		method = sconfig.TokenEndpointAuthClientSecretPost
+	}
+
+	out := make(url.Values, len(values)+2)
+	for k, v := range values {
+		out[k] = v
+	}
+
+	switch method {
+	case sconfig.TokenEndpointAuthClientSecretPost:
+		out.Set("client_id", clientId)
+		out.Set("client_secret", clientSecret)
+		return out, "", nil
+	case sconfig.TokenEndpointAuthClientSecretBasic:
+		// RFC 6749 §2.3.1: client_id and client_secret are
+		// application/x-www-form-urlencoded before being colon-joined and
+		// base64-encoded. The gentleman/std basic-auth helpers do not do
+		// this form-encoding step, so we build the header by hand.
+		userPass := url.QueryEscape(clientId) + ":" + url.QueryEscape(clientSecret)
+		encoded := base64.StdEncoding.EncodeToString([]byte(userPass))
+		return out, "Basic " + encoded, nil
+	case sconfig.TokenEndpointAuthNone:
+		out.Set("client_id", clientId)
+		return out, "", nil
+	default:
+		return nil, "", fmt.Errorf("unsupported token_endpoint_auth_method %q", method)
+	}
+}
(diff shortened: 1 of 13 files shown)


internal/schema/connectors/auth_oauth2.go
```diff
@@ -5,14 +5,53 @@ import (
...
-	ClientSecret  *common.StringValue     `json:"client_secret" yaml:"client_secret"`
-	Scopes        []Scope                 `json:"scopes" yaml:"scopes"`
-	Authorization AuthOauth2Authorization `json:"authorization" yaml:"authorization"`
-	Token         AuthOauth2Token         `json:"token" yaml:"token"`
-	Revocation    *AuthOauth2Revocation   `json:"revocation,omitempty" yaml:"revocation,omitempty"`
+	Type AuthType `json:"type" yaml:"type"`
+	// TokenEndpointAuthMethod selects how client credentials are presented
+	// to the token endpoint. When empty, defaults to client_secret_post —
+	// matches the proxy's behavior before this field existed.
+	TokenEndpointAuthMethod TokenEndpointAuthMethod `json:"token_endpoint_auth_method,omitempty" yaml:"token_endpoint_auth_method,omitempty"`
```
rmorlok: Change this to `TokenEndpointAuthMethod *TokenEndpointAuthMethod` where nil represents the default value that falls back to `TokenEndpointAuthClientSecretPost` as in `GetTokenEndpointAuthMethodOrDefault` below. Other places that depend on `TokenEndpointAuthMethod` should not apply a default for empty string. That is not a valid value.

internal/schema/connectors/auth_oauth2.go
```diff
@@ -109,6 +148,36 @@ func (a *AuthOAuth2) Validate(vc *common.ValidationContext) error {
 		}
 	}
 
+	hasSecret := a.ClientSecret != nil && a.ClientSecret.InnerVal != nil
+	switch a.TokenEndpointAuthMethod {
+	case "", TokenEndpointAuthClientSecretPost, TokenEndpointAuthClientSecretBasic:
```
rmorlok: See other note about not accepting empty string at this point.

internal/auth_methods/oauth2/client_auth.go
```diff
@@ -0,0 +1,87 @@
...
+// The values map is treated as input-only and not mutated; the returned
+// url.Values is a copy with the appropriate fields set.
+func applyTokenEndpointClientAuth(
+	method sconfig.TokenEndpointAuthMethod,
+	clientId, clientSecret string,
+	values url.Values,
+) (url.Values, string, error) {
+	// Effective default — empty method behaves as client_secret_post for
+	// backwards compatibility.
+	if method == "" {
```
rmorlok: See other notes about not accepting defaults at this point.

internal/auth_methods/oauth2/client_auth.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	out := make(url.Values, len(values)+2)
+	for k, v := range values {
+		out[k] = v
+	}
+
+	switch method {
+	case sconfig.TokenEndpointAuthClientSecretPost:
+		out.Set("client_id", clientId)
```
rmorlok, line 64: Is there an RFC that we are following at this phase? If so add a comment like TokenEndpointAuthClientSecretBasic below.

internal/auth_methods/oauth2/client_auth.go
```diff
@@ -0,0 +1,87 @@
...
+	case sconfig.TokenEndpointAuthClientSecretBasic:
+		// RFC 6749 §2.3.1: client_id and client_secret are
+		// application/x-www-form-urlencoded before being colon-joined and
+		// base64-encoded. The gentleman/std basic-auth helpers do not do
+		// this form-encoding step, so we build the header by hand.
+		userPass := url.QueryEscape(clientId) + ":" + url.QueryEscape(clientSecret)
+		encoded := base64.StdEncoding.EncodeToString([]byte(userPass))
+		return out, "Basic " + encoded, nil
+	case sconfig.TokenEndpointAuthNone:
+		out.Set("client_id", clientId)
```
rmorlok, line 80: Same note here, add a comment about the RFC section if there is one for this.

internal/schema/connectors/auth_oauth2.go
rmorlok: Done in 23a3c7e — field is now `*TokenEndpointAuthMethod` with nil meaning "use the default". `GetTokenEndpointAuthMethodOrDefault` is the only place the nil → client_secret_post collapse happens. Added a `NewTokenEndpointAuthMethod` ptr constructor for typed-string-const call sites.

internal/schema/connectors/auth_oauth2.go
rmorlok: Done in 23a3c7e — the validator now explicitly rejects an empty-string value (`token_endpoint_auth_method: must not be empty`) instead of silently treating it as the default. Only nil/omitted falls through to the default. New test `TestAuthOAuth2_Validate_RejectsExplicitEmptyTokenEndpointAuthMethod` covers it.

internal/auth_methods/oauth2/client_auth.go
rmorlok: Done in 23a3c7e — removed the `if method == "" { method = post }` shim. Callers MUST pass a resolved method (typically via `AuthOAuth2.GetTokenEndpointAuthMethodOrDefault`); empty string now returns an error from the switch's default case. New test `TestApplyTokenEndpointClientAuth_RejectsEmptyMethod` guards the contract.

internal/auth_methods/oauth2/client_auth.go
rmorlok: Done in 23a3c7e — added `RFC 6749 §2.3.1 / RFC 7591 §2` reference on the `client_secret_post` branch. (§2.3.1 marks it NOT RECOMMENDED relative to Basic; §7591 §2 is where the method name itself is registered.)

internal/auth_methods/oauth2/client_auth.go
rmorlok: Done in 23a3c7e — added `RFC 7591 §2` reference on the `none` branch, plus a note that proof-of-possession comes from PKCE (RFC 7636), which the connector validator already requires when this method is selected.

Issue #257, opened 2026-05-10
[api-key] Proxy-triggered probe acceleration on 401/403


rmorlok:
Part of #243 — see that issue, Part 5 (Optional optimization).

## Scope

When the proxy receives a `401` or `403` from the upstream, enqueue a one-shot probe-now task for that connection's probes. Reduces detection lag from "next probe interval" (potentially many minutes) to ~immediate.

This is a **layered optimization** on top of the probe-driven health model — it is *not* a separate reauth-trigger mechanism. The probe outcomes still feed the existing counters and `health_state` transitions defined in the probe-runtime sub-issue.

### Behavior

- 401/403 from upstream → enqueue probe-now (Asynq one-shot task) for each probe on the connection.
- Throttle to avoid storms: e.g. max one enqueue per probe per connection per minute.
- Probe-now outcomes update counters identically to scheduled runs.

## Test plan

- [ ] 401 from upstream enqueues probe-now task
- [ ] 403 from upstream enqueues probe-now task
- [ ] Throttle prevents repeated enqueues within the configured window
- [ ] Probe-now outcome updates counters identically to a scheduled run
- [ ] Other 4xx (e.g. 400, 404, 429) do not enqueue

## Dependencies

- Proxy sub-issue
- Probe-runtime sub-issue

## Note

Marked optional in #243. Land if the probe-runtime sub-issue ships with bandwidth to spare; otherwise this can wait — base detection still works, just at probe-interval granularity.

PR #356, opened 2026-05-23
[api-key] Integration test harness + cross-cutting probe tests (#258)

### LINKED ISSUE #258: [api-key] Integration test harness + cross-cutting probe tests
Part of #243 — see that issue, Part 8.

## Scope

End-to-end integration tests for api-key auth and the generalized probe-driven health model.

### `integration_tests/api_key/`

Mirror the structure of `integration_tests/oauth2/harness_smoke_test.go`.

Test provider: stub upstream that requires the configured credential and returns `401` if absent or wrong. Should support all four placements (bearer / header / query / basic).

Coverage:

- [ ] Connection initiate → form → submit → verify → ready (one test per placement)
- [ ] Manual key rotation: submit new credential, prior soft-deleted, verify succeeds
- [ ] Auto-reauth: stub upstream rotates its accepted secret → probes flip connection to `unhealthy` → user submits new key → connection returns to `healthy`
- [ ] No-replay assertion: prior key bytes never appear in any UI / API response or log line after rotation begins

### `integration_tests/probes/` (or extend existing OAuth2 harness)

Cross-cutting test proving the probe-driven health logic generalizes:

- [ ] OAuth2 connection: forced probe failures flip `health_state` to `unhealthy`
- [ ] OAuth2 connection: subsequent successes flip `health_state` back to `healthy`

## Dependencies

- Connection-initiate-flow sub-issue
- Proxy sub-issue
- Probe-runtime sub-issue
- Reauth-flow sub-issue

rmorlok:
## Summary

Adds end-to-end coverage for the api-key auth method (per-placement initiate→ready, manual rotation, auto-reauth via probe-driven health) and a cross-cutting OAuth2 probe-driven health test that proves the probe runtime generalizes across auth types.

Closes #258 — the validation pass for the #243 api-key project.

## Probe runtime fix

`internal/core/probe_http.go` silently treated non-2xx upstream responses as successful probes. Without this fix, an upstream that 401s the proxied request would never flip the connection unhealthy — which broke the probe-driven health story end-to-end. Both probe branches (proxy and raw) now treat status codes outside `[200,300)` as a failure, surfacing the status code in the recorded error.

## iface.C surface additions

Two synchronous entry points used by the integration tests, and reusable for future operational paths (e.g. #257's probe-now optimization):

- `RunProbe(ctx, connectionId, probeId)` — inline counterpart to the periodic asynq probe task. Invokes a probe and records the outcome against health-state counters.
- `RunVerifyConnection(ctx, connectionId)` — inline counterpart to the verify_connection asynq task. Runs every probe and advances `setup_step` on the outcome.

Both share the existing task-handler body via small refactors; behavior under asynq is unchanged.

## Test coverage

**``integration_tests/api_key/``**
- `harness_smoke_test.go` — `ApiKeyStubUpstream` sanity for bearer.
- `initiate_flow_test.go` — `initiate → form → submit → verify → ready` for bearer / header / query / basic, asserting on the placement-specific upstream request shape.
- `rotation_test.go` — manual rotation via `/_reauth`, active credential row id changes, **no-replay** invariant: old key bytes never appear in any upstream request that arrives after rotation begins.
- `auto_reauth_test.go` — provider-side rotation invalidates the stored key → probe fails → health flips unhealthy → user submits new key → probe succeeds → health flips back to healthy. `failure_threshold=1` and `recovery_threshold=1` for test determinism.

**``integration_tests/probes/``**
- `oauth2_probe_health_test.go` — same failure / recovery transitions against an OAuth2 connection, using the test provider's resource endpoint as the probe target and scripted 401s to force the failure window. Proves the probe-driven health logic is auth-method-agnostic.

## Test plan

- [x] `go test -tags integration ./api_key/...`
- [x] `go test -tags integration ./probes/...`
- [x] Existing OAuth2 non-chromedp tests still pass
- [x] `./scripts/preflight.sh` clean

Note: two pre-existing chromedp tests (`TestStandardAuthorizationCodeFlow`, `TestUserDenialFlow`) fail in this environment with `unknown IPAddressSpace value: Loopback` chromedp/Chrome version mismatch. Verified to fail on bare main as well — unrelated to this PR.

### DIFF
--- a/internal/core/probe_http.go
+++ b/internal/core/probe_http.go
@@ -3,6 +3,7 @@ package core
 import (
 	"bytes"
 	"context"
+	"fmt"
 
 	"github.com/rmorlok/authproxy/internal/core/iface"
 	"github.com/rmorlok/authproxy/internal/httpf"
@@ -12,6 +13,12 @@ type probeHttp struct {
 	probeBase
 }
 
+// Invoke runs the probe and reports either success or error. Both branches
+// (proxy and raw) treat any non-2xx upstream response as a probe failure —
+// without this, an upstream that 401s the proxied request would silently be
+// considered a success and the probe-driven health signal would never flip
+// to unhealthy. The recorded error carries the status code so operators can
+// see what the upstream actually returned.
 func (p *probeHttp) Invoke(ctx context.Context) (string, error) {
 	return p.recordInvokeOutcome(ctx, func(ctx context.Context) (string, error) {
 		if p.cfg.ProxyHttp != nil {
@@ -30,13 +37,17 @@ func (p *probeHttp) Invoke(ctx context.Context) (string, error) {
 				BodyJson: h.BodyJson,
 			}
 
-			// TODO: translate result
-			_, err = proxy.ProxyRequest(ctx, httpf.RequestTypeProbe, &req)
+			resp, err := proxy.ProxyRequest(ctx, httpf.RequestTypeProbe, &req)
 			if err != nil {
 				return ProbeOutcomeError, err
-			} else {
-				return ProbeOutcomeSuccess, nil
 			}
+			if resp == nil {
+				return ProbeOutcomeError, fmt.Errorf("probe upstream returned no response")
+			}
+			if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+				return ProbeOutcomeError, fmt.Errorf("probe upstream returned status %d", resp.StatusCode)
+			}
+			return ProbeOutcomeSuccess, nil
 		} else {
 			req := p.s.httpf.
 				ForConnection(p.c).
@@ -62,12 +73,17 @@ func (p *probeHttp) Invoke(ctx context.Context) (string, error) {
 				req.BodyString(h.Body)
 			}
 
-			_, err := req.Do()
+			resp, err := req.Do()
 			if err != nil {
 				return ProbeOutcomeError, err
-			} else {
-				return ProbeOutcomeSuccess, nil
 			}
+			if resp == nil {
+				return ProbeOutcomeError, fmt.Errorf("probe upstream returned no response")
+			}
+			if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+				return ProbeOutcomeError, fmt.Errorf("probe upstream returned status %d", resp.StatusCode)
+			}
+			return ProbeOutcomeSuccess, nil
 		}
 	})
 }
--- a/internal/core/task_probe.go
+++ b/internal/core/task_probe.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log/slog"
 
 	"github.com/hibiken/asynq"
 	"github.com/rmorlok/authproxy/internal/apid"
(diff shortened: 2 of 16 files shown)


internal/core/probe_http.go
```diff
@@ -30,13 +37,17 @@ func (p *probeHttp) Invoke(ctx context.Context) (string, error) {
...
 
-			// TODO: translate result
-			_, err = proxy.ProxyRequest(ctx, httpf.RequestTypeProbe, &req)
+			resp, err := proxy.ProxyRequest(ctx, httpf.RequestTypeProbe, &req)
 			if err != nil {
 				return ProbeOutcomeError, err
-			} else {
-				return ProbeOutcomeSuccess, nil
 			}
+			if resp == nil {
```
rmorlok, line 44: I see probe_http.go and probe_no_op.go don't have unit tests. Please add tests to both of the files -- be sure to cover the scenario where this response code situation would have been caught.

internal/core/task_verify_connection.go
```diff
@@ -48,11 +48,36 @@ func (s *service) verifyConnection(ctx context.Context, t *asynq.Task) error {
...
+	return s.runVerifyConnection(ctx, connectionId)
+}
+
+// runVerifyConnection is the shared body of the verify_connection task and
+// the inline RunVerifyConnection entry. It loads the connection, runs every
+// probe in order, and advances the setup step based on the outcome. Returns
+// asynq.SkipRetry for non-recoverable shapes (not-found, wrong phase) so the
+// task handler can return without retrying; wraps onVerifyFailed errors
+// unchanged so they remain retryable.
+func (s *service) runVerifyConnection(ctx context.Context, connectionId apid.ID) error {
```
rmorlok, line 74: I see this task is also missing a test. Please add one.

internal/core/task_probe.go
```diff
@@ -62,25 +63,39 @@ func (s *service) runProbeForConnection(ctx context.Context, t *asynq.Task) erro
...
+	return invokeErr
+}
+
+// runProbeInternal is the shared body executed by both the periodic asynq task
+// handler and the inline RunProbe entry point. It loads the connection, looks
+// up the probe by id, invokes it, and records the outcome against the
+// connection's health-state counters. Returns the probe (when found) and any
+// error from looking it up or invoking it; health-recording errors are logged
+// but not returned because they don't invalidate the probe outcome itself.
+func (s *service) runProbeInternal(ctx context.Context, logger *slog.Logger, connectionId apid.ID, probeId string) (iface.Probe, error) {
```
rmorlok, line 79: This task is missing a unit test. Please add one.

internal/core/probe_http.go
rmorlok: Added in `internal/core/probe_http_test.go` and `internal/core/probe_no_op_test.go` (commit f708835).

`probe_http_test.go` covers both branches:
- `ProxyHttp_Success` (200), `ProxyHttp_Non2xxIsFailure` (401/403/404/500/503 each surface as `ProbeOutcomeError` with the status code in the error — the exact regression this PR's fix prevents), `ProxyHttp_2xxRangeIsSuccess` (200/201/204/299), `ProxyHttp_TransportErrorIsFailure`.
- `RawHttp_Success` / `RawHttp_Non2xxIsFailure` for the gentleman/httpf path via gentleman-mock.

`probe_no_op_test.go` covers `TestProbeNoOp_Invoke` — `NewProbe` with neither Http nor ProxyHttp returns the no-op variant, which always reports success across repeated calls.

internal/core/task_verify_connection.go
rmorlok: Added in `internal/core/task_verify_connection_test.go` (commit f708835). Cases:

- `NoProbes_AdvancesToReady` — connector without probes runs the `onVerifyPassed` path: clears `setup_step`, transitions to `Ready`.
- `StalePhase_SkipsCleanly` — connection already past verify is a no-op (matches the asynq stale-task guard).
- `ConnectionNotFound` — returns `asynq.SkipRetry`.
- `ProbeFailure` — non-2xx probe → `onVerifyFailed` → `setup_error` populated, `setup_step=verify_failed`, error wraps `SkipRetry`.

internal/core/task_probe.go
rmorlok: Added in `internal/core/task_probe_test.go` (commit f708835). Cases:

- `Success` — periodic outcome row inserted, no health transition when already healthy.
- `FailureBelowThreshold` — outcome inserted, no transition fires (threshold=3, one observed failure).
- `FailureCrossesThreshold` — threshold=1 + one failure triggers `SetConnectionHealthState(unhealthy)`.
- `ProbeNotFound` / `ConnectionNotFound` — both return `asynq.SkipRetry`.

While writing the not-found tests I caught a pre-existing argument-order bug in `runProbeInternal`: `errors.Is(database.ErrNotFound, err)` was reversed (correct order is `errors.Is(err, target)`). Because `s.getConnection` wraps `database.ErrNotFound` inside `iface.ErrConnectionNotFound`, the reversed check never matched, so the `SkipRetry` branch silently fell through and asynq would retry probe tasks against a deleted connection. Fixed both checks (connection-not-found and probe-not-found); `TestRunProbe_ConnectionNotFound` guards against the regression.

Issue #366, opened 2026-05-23
Auth-method factories emit setup steps; core dispatches via factory; drop SetupFlow.Credentials + Connector.Normalize


rmorlok:
Part of #360. Depends on #361 (factory standardization), #362 (manifest types), #365 (ID-based step encoding).

## Goal

Make auth methods own their setup-step emission. Replace core's per-auth-type type-asserts with a uniform "ask the factory for its steps and dispatch" pattern. Drop the now-unused ``SetupFlow.Credentials`` field, ``Connector.Normalize``, and the schema-side ``SynthesizeApiKeyCredentialsStep``.

## Today

After the prior subtickets land:

- ``auth_methods.Factory`` is a real interface that all three methods implement.
- ``core/iface.ManifestSetupStep`` / ``ManifestSetupFlow`` are defined but unused.
- ``setup_step_id`` is the column; ``apxy:`` is the reserved prefix.

But the dispatch hot paths still type-assert on ``connector.Auth.Inner()``:

- ``service_connection_initiate.go:134``: ``if _, ok := connector.Auth.Inner().(*config.AuthOAuth2); ok { ... }`` — direct OAuth2 redirect logic.
- ``service_connection_submit.go::submitCredentialsStep``: ``switch auth := connector.Auth.Inner().(type) { case *config.AuthApiKey: ... }``.
- ``service_connection_submit.go::initiateAuthStep``: ``if _, ok := connector.Auth.Inner().(*config.AuthOAuth2); !ok { return error }``.
- ``service_connection_reauth.go``: branches on ``case *config.AuthApiKey:`` vs ``case *config.AuthOAuth2:``.

And the schema still has ``SetupFlow.Credentials`` + the synthesized step bolt-on.

### 1. Extend ``auth_methods.Factory``

```go
type Factory interface {
    NewAuthenticator(connection coreIface.Connection) Authenticator
    // ManifestSetupSteps returns the ordered auth-phase steps this method
    // contributes for the given connection. Called when building the
    // ManifestSetupFlow. May be empty (no_auth contributes nothing).
    ManifestSetupSteps(connection coreIface.Connection) []coreIface.ManifestSetupStep
}
```

### 2. Per-method implementations

- ``internal/auth_methods/no_auth`` — ``ManifestSetupSteps`` returns ``nil``.
- ``internal/auth_methods/api_key`` — ``ManifestSetupSteps`` returns one step:
[...]

Issue #360, opened 2026-05-23
Refactor connector setup flow: schema-only schema, runtime in iface + auth methods


rmorlok:
## Motivation

Today the connector setup flow has two leaky abstractions:

1. **The schema package describes things it shouldn't know about.** `internal/schema/connectors/setup_flow.go` carries a `Credentials` phase whose presence/shape depends entirely on the auth method (`Connector.Normalize` synthesizes it for `AuthApiKey`; OAuth2 never uses it). The schema package is supposed to describe what the user authors — credentials are not user-authored.

2. **`internal/core` reasons directly about auth type.** `service_connection_initiate.go:134`, `service_connection_submit.go::submitCredentialsStep`, `service_connection_reauth.go`, etc. all type-assert on `connector.Auth.Inner()` to make dispatch decisions. Adding a new auth type (e.g. #352 client-credentials) or a new grant within an existing type means editing each of those sites.

## Goal

- `internal/schema/connectors` describes only user-authored configuration.
- A new manifest type in `internal/core/iface` describes the runtime, fully-materialized setup flow (preconnect from schema + auth-method-emitted steps + configure from schema + implicit failure pseudo-steps).
- Each auth method (`internal/auth_methods/{oauth2,api_key,no_auth}`) owns its own setup-step emission and credential persistence. Core dispatches via a uniform factory.
- ID-based step identifiers replace the current ``phase:index`` encoding.
- ``ConnectionState`` values renamed to better reflect that the setup-step + health-state pair (not the lifecycle state alone) determines whether a connection is usable.
- New ``redirect`` step type with signed ``RETURN_ADVANCE`` / ``RETURN_ABORT`` tokens for connectors whose 3rd party can bounce the user back to an arbitrary URL.

### 1. ``auth_methods.Factory`` interface

New interface in top-level ``internal/auth_methods``:

```go
type Factory interface {
    NewAuthenticator(connection coreIface.Connection) Authenticator
    ManifestSetupSteps(connection coreIface.Connection) []coreIface.ManifestSetupStep
}
```

Each per-method package (``oauth2``, ``api_key``, ``no_auth``) implements it. Per-method extras (e.g. ``oauth2.Factory.NewOAuth2``, ``oauth2.Factory.GetOAuth2State``) stay on the concrete type. ``no_auth`` gains a real factory for symmetry (today it has only a package-level ``NewAuthenticator()``).

[...]

PR #373, opened 2026-05-23
Standardize auth_methods.Factory interface + uniform factory registry (#361)

### LINKED ISSUE #361: Standardize auth_methods.Factory interface + uniform factory registry in core
Part of #360.

## Goal

Lift a common ``auth_methods.Factory`` interface so each ``internal/auth_methods/*`` package exposes a uniform factory, and replace core's per-method accessors with a single resolver keyed by auth type.

## Today

Inconsistent factory surfaces:

- ``internal/auth_methods/oauth2.Factory`` — ``NewOAuth2(connection) OAuth2Connection``, ``NewAuthenticator(connection) auth_methods.Authenticator``, ``GetOAuth2State(ctx, actor, stateId) (OAuth2Connection, error)``.
- ``internal/auth_methods/api_key.Factory`` — ``NewAuthenticator(connection) auth_methods.Authenticator``.
- ``internal/auth_methods/no_auth`` — no factory; package-level ``NewAuthenticator() auth_methods.Authenticator``.

Core wires each separately:

- ``service.o2Factory`` constructed lazily in ``getOAuth2Factory()`` (``service_oauth.go``).
- ``service.apiKeyFactory`` constructed lazily in ``getApiKeyFactory()`` (``service_api_key.go``).
- ``no_auth`` invoked directly via the top-level function (``connection_proxy.go``).

Every call site that needs to act on an auth method does an auth-type type-assert and then calls the right factory.

### 1. Top-level interface

Add to ``internal/auth_methods/factory.go``:

```go
type Factory interface {
    NewAuthenticator(connection coreIface.Connection) Authenticator
}
```

(``ManifestSetupSteps`` is added in a later subticket — keep this PR focused on the bare interface.)

### 2. Per-method conformance

- ``internal/auth_methods/oauth2/factory.go`` — declare ``var _ auth_methods.Factory = (*factory)(nil)``. ``NewOAuth2`` and ``GetOAuth2State`` remain on the concrete type / package-specific ``oauth2.Factory`` interface.
- ``internal/auth_methods/api_key/factory.go`` — already shape-conformant; add the var-assertion.
- ``internal/auth_methods/no_auth/`` — introduce a real ``factory`` struct + ``NewFactory()`` + ``var _ auth_methods.Factory = (*factory)(nil)``. Keep the existing package-level ``NewAuthenticator()`` deprecated-but-working during the transition, or delete it if no external callers remain.

### 3. Core wiring

- ``internal/core/service.go`` — replace ``o2FactoryOnce/o2Factory`` and ``apiKeyFactoryOnce/apiKeyFactory`` with a single registry populated at ``NewCoreService`` time:

```go
authMethodFactories map[reflect.Type]auth_methods.Factory
```

  (Or keyed by a small enum derived from ``config.AuthImpl``; reflect.Type works fine because the registry is built once at startup and read often.)

- Add ``service.getAuthMethodFactory(connector *cschema.Connector) auth_methods.Factory`` that maps ``connector.Auth.Inner()`` to the registered factory. Returns an error if no factory is registered for the auth type.

- ``service_oauth.go::getOAuth2Factory()`` and ``service_api_key.go::getApiKeyFactory()`` become thin pass-throughs (still needed for the OAuth2-specific extras like ``NewOAuth2`` / ``GetOAuth2State``). Mark them with comments tying them to the generic resolver.

- ``connection_proxy.go::getAuthenticator()`` switches from a type-switch over ``connector.Auth.Inner()`` to ``getAuthMethodFactory(connector).NewAuthenticator(c)``.

### 4. Tests

Existing unit tests should keep passing. Add a small test in ``internal/core`` that ``getAuthMethodFactory`` returns the expected factory type for each of the three auth methods.

## Out of scope

- ``ManifestSetupSteps`` method on ``Factory`` (separate subticket).
- Removing the OAuth2-specific extras on its concrete factory (they stay).
- Any schema or DB changes.

## Acceptance

- All three auth methods implement ``auth_methods.Factory``.
- ``connection_proxy.go``'s authenticator resolution has no ``connector.Auth.Inner().(*config.AuthFoo)`` switch.
- Existing tests pass.

rmorlok:
Closes #361. Part of #360.

## Summary

- New ``auth_methods.Factory`` interface in ``internal/auth_methods/factory.go`` (just ``NewAuthenticator`` for now; ``ManifestSetupSteps`` lands in #366).
- ``no_auth`` gains a real factory for symmetry with the other auth methods.
- Core service holds a ``map[cschema.AuthType]auth_methods.Factory`` populated eagerly at ``NewCoreService`` (replaces the per-method ``sync.Once`` lazy accessors). New resolver ``getAuthMethodFactory(connector)``.
- ``connection_proxy.go::resolveAuthenticator`` no longer type-switches over ``connector.Auth.Inner()`` — it dispatches via the registry.
- ``getOAuth2Factory`` / ``getApiKeyFactory`` remain as thin pass-throughs so OAuth2-specific extras (``NewOAuth2``, ``GetOAuth2State``) keep their typed surface.

## Test plan

- [x] ``go test ./...`` passes (full repo, 30+ packages).
- [x] ``./scripts/preflight.sh`` passes.
- [x] New unit tests in ``internal/core/service_auth_methods_test.go`` cover the resolver for each registered auth type and the nil/unknown cases.


internal/core/service_api_key.go
```diff
@@ -1,13 +1,16 @@
...
+// auth_methods.Factory surface, but later subtickets (#364 credential
+// persistence move) will add api-key-specific methods that this typed
+// accessor exposes. Generic call sites should use getAuthMethodFactory
+// instead.
 func (s *service) getApiKeyFactory() api_key.Factory {
-	s.apiKeyFactoryOnce.Do(func() {
-		s.apiKeyFactory = api_key.NewFactory(s.db, s.encrypt, s.httpf, s.logger)
-	})
-	return s.apiKeyFactory
+	return s.authMethodFactories[cschema.AuthTypeAPIKey].(api_key.Factory)
```
rmorlok: Can this method be removed in favor of just using the general factory system that we've introduced at the callsite?

internal/core/service_oauth.go
```diff
@@ -1,15 +1,16 @@
...
-	s.o2FactoryOnce.Do(func() {
-		var opts []oauth2.FactoryOption
-		if s.telProviders != nil {
-			opts = append(opts, oauth2.WithTelemetry(s.telProviders, s.telCfg))
-		}
-		s.o2Factory = oauth2.NewFactory(s.cfg, s.db, s.r, s, s.httpf, s.encrypt, s.logger, opts...)
-	})
-
-	return s.o2Factory
+	return s.authMethodFactories[cschema.AuthTypeOAuth2].(oauth2.Factory)
```
rmorlok: Same question here? Can we just update callsites to use the generic factory system?

rmorlok: Done in 0101729 — deleted `getOAuth2Factory()` and updated the three call sites (`service_connection_initiate.go`, `service_connection_submit.go`, `connection_revoke_credentials.go`) to use `getAuthMethodFactory(connector).(oauth2.Factory).NewOAuth2(...)`. Each site already type-asserts on `*config.AuthOAuth2` first, so the cast is safe; those type-asserts go away in #366.

internal/core/service_api_key.go
rmorlok: Done in 0101729 — `getApiKeyFactory()` had no remaining callers so I just deleted the file.

internal/core/connection_revoke_credentials.go
```diff
@@ -15,8 +16,7 @@ func (c *connection) getRevokeCredentialsOperations() []operation {
 	auth := def.Auth
 
 	if _, ok := auth.Inner().(*config.AuthOAuth2); ok {
-		o2f := c.s.getOAuth2Factory()
-		o2 := o2f.NewOAuth2(c)
+		o2 := c.s.getAuthMethodFactory(def).(oauth2.Factory).NewOAuth2(c)
```
rmorlok: Change this so it doesn't require an explicit cast. Instead update the Authenticator interface to have a  SupportsRevoke method to indicate that a particular connection can support revoke. If it does support revoke, confirm that the authenticator supports a s Remove (currently listed as RevokeTokens) interface and apply that.  Basically make this generic across auth types. Don't make it assume OAuth2.

rmorlok: Done in acc33fc — extended `auth_methods.Authenticator` with `SupportsRevoke() bool` and `Revoke(ctx) error`. OAuth2 delegates to its existing `SupportsRevokeTokens` / `RevokeTokens` (kept on `OAuth2Connection` for the typed callers); api-key and no-auth return false / no-op. `connection_revoke_credentials.go` is now auth-type agnostic: ask factory for the Authenticator, gate on `SupportsRevoke`, return `auth.Revoke` as the generic operation. Disconnect test mocks the new Authenticator surface via a generated mock.

internal/auth_methods/oauth2/revocation.go
```diff
@@ -14,12 +14,24 @@ func (o *oAuth2Connection) SupportsRevokeTokens() bool {
...
 }
 
+// SupportsRevoke is the generic auth_methods.Authenticator surface for
+// SupportsRevokeTokens. The OAuth2-specific name remains on OAuth2Connection
+// for callers that already work against the typed interface.
+func (o *oAuth2Connection) SupportsRevoke() bool {
+	return o.SupportsRevokeTokens()
+}
+
 func (o *oAuth2Connection) RevokeTokens(ctx context.Context) error {
```
rmorlok: Let's remove this method in favor of just having the revoke method below. This method doesn't add anything given that we have a revokeTokensInner(...).

rmorlok: Done — and applied the same logic to `SupportsRevokeTokens` since it was the same shape (one-line delegate). Inlined the span wrap from `RevokeTokens` directly into `Revoke` (calling `revokeTokensInner`), moved the `SupportsRevokeTokens` body into `SupportsRevoke`, and dropped both from the `OAuth2Connection` interface. Tests + mock updated.

PR #416, opened 2026-05-27
Centralize connector API DTOs

### LINKED ISSUE #383: Schema project: migrate connector API contracts to centralized schema models
Parent: #96
Depends on: #381 and the `internal/schema/api` foundation issue.

## Goal
Make connector API request/response contracts use centralized schema models instead of route-local and swagger-only DTO definitions.

## Scope
- Move connector API DTOs from `internal/routes/connectors.go` and `internal/routes/swagger_models.go` into `internal/schema/api`.
- Reuse `internal/schema/resources/connectors` for connector resource shape instead of duplicating connector fields.
- Update connector routes, tests, and swagger annotations to reference centralized models.
- Add/update JSON schema definitions, tests, and sample payloads for connector API endpoints.
- Add or update README documentation for connector resource/API model boundaries.

## Acceptance Criteria
- Connector API models have a single canonical definition under schema packages.
- Route-local connector DTOs and swagger-only connector duplicates are removed or reduced to aliases during transition.
- Swagger docs regenerate without unexpected contract drift.

rmorlok:
## Summary
- move connector API request/response DTOs into internal/schema/api and keep routes on aliases
- add OpenAPI-only connector adapters for schema-driven definition fields
- add connector API JSON schema definitions, validation samples, and regenerated Swagger docs

Closes #383.

## Tests
- go test ./internal/schema/...
- go test ./internal/routes -run TestConnectors
- go test ./internal/... -run '^$'
- (cd integration_tests && go test ./... -run '^$')
- git diff --check
- ./scripts/preflight.sh


internal/schema/api/schema.json
```diff
@@ -235,6 +239,125 @@
       },
       "required": ["items"],
       "additionalProperties": false
+    },
+    "Connector": {
```
rmorlok, line 243: Does it make sense to import this from the `internal/schema/resources/connectors` JSON schema? Or is there some sort of difference?

rmorlok: Good callout. I kept this as an API-owned schema rather than importing the resource connector schema because this object is the connector summary response projection, not the connector definition syntax. It flattens some definition fields for API consumers (for example logo is a URL string) and adds runtime/API fields such as state, has_configure, timestamps, version counts, and state summaries. The canonical resource connector schema is still referenced for the definition fields on create/update/version detail. I added an explicit description to the API schema and refreshed Swagger in 4aeab16.

Issue #404, opened 2026-05-25
App metrics: add resource snapshot storage


rmorlok:
Parent: #395

## Goal
Add durable app metrics storage for resource snapshots so historical resource counts can be queried independently from the main database.

## Scope
- Add resource sample tables for connection and actor metrics.
- Store sample timestamp, resource type, resource id, namespace, labels, lifecycle fields, and resource-specific dimensions.
- Preserve deleted resources historically by storing point-in-time samples rather than querying live tables.
- Add migrations for SQLite, Postgres, and ClickHouse.
- Add provider methods for writing and querying resource samples.

## Acceptance Criteria
- Resource samples can be inserted idempotently for a snapshot timestamp.
- Connection samples include state, health_state, connector_id, and connector_version.
- Actor samples include namespace and labels at minimum.
- Label selectors and namespace matcher-compatible fields are stored in queryable form.

PR #432, opened 2026-05-29
Add app metrics resource snapshot worker

### LINKED ISSUE #407: App metrics: add resource snapshot worker
Parent: #395

## Goal
Add the periodic worker that snapshots live AuthProxy resources into app_metrics storage on a configurable cadence.

## Scope
- Add an Asynq task and cron registrar for resource metric snapshots.
- Default cadence should be 15 minutes via app_metrics.resource_snapshot_interval.
- Snapshot connections and actors from the main database.
- Include resources that are live for the sampled time slice; deleted resources should remain represented by older samples but not appear in later samples.
- Add structured logging and retry behavior appropriate for periodic background work.

## Acceptance Criteria
- Worker registers the periodic snapshot task when app metrics are enabled/configured.
- Snapshot cadence is configurable and defaults to 15m.
- Tests cover connection and actor sample creation, deleted-resource behavior, and idempotent reruns for the same bucket.
- Task failures do not break request-event logging.

rmorlok:
## Summary
- add an app metrics resource snapshot task and periodic registrar
- snapshot live connections and actors into app metrics samples on the configured cadence
- wire the task into the worker scheduler and asynq mux
- cover sample creation, deleted-resource behavior, idempotent reruns, and cadence registration

Closes #407

## Tests
- `env GOCACHE=/private/tmp/authproxy-go-cache AUTH_PROXY_TEST_DATABASE_PROVIDER=sqlite AUTH_PROXY_APP_METRICS_TEST_DATABASE_PROVIDER=sqlite go test ./internal/app_metrics ./internal/service ./internal/service/worker`
- `./scripts/preflight.sh`

Note: `go test ./...` was attempted locally with SQLite selectors, but this sandbox blocks local listener creation and home-directory writes used by unrelated packages (httptest/miniredis/OTLP tests).

### DIFF
--- a/internal/app_metrics/task_resource_snapshot.go
+++ b/internal/app_metrics/task_resource_snapshot.go
@@ -1,0 +1,183 @@
+package app_metrics
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"time"
+
+	"github.com/hibiken/asynq"
+	"github.com/rmorlok/authproxy/internal/apctx"
+	"github.com/rmorlok/authproxy/internal/database"
+	sconfig "github.com/rmorlok/authproxy/internal/schema/config"
+	"github.com/rmorlok/authproxy/internal/util/pagination"
+)
+
+const (
+	taskTypeResourceSnapshot  = "app_metrics:resource_snapshot"
+	resourceSnapshotBatchSize = 1000
+)
+
+// ResourceSnapshotTaskHandler snapshots live resources into app_metrics storage.
+type ResourceSnapshotTaskHandler struct {
+	db     database.DB
+	store  ResourceSampleStore
+	cfg    *sconfig.AppMetrics
+	logger *slog.Logger
+}
+
+func NewResourceSnapshotTaskHandler(
+	db database.DB,
+	store ResourceSampleStore,
+	cfg *sconfig.AppMetrics,
+	logger *slog.Logger,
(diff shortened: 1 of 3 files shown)


internal/app_metrics/task_resource_snapshot.go
```diff
@@ -0,0 +1,183 @@
...
+	logger.Info(
+		"completed app metrics resource snapshot",
+		"connections", connectionCount,
+		"actors", actorCount,
+		"duration", time.Since(start),
+	)
+	return nil
+}
+
+func (h *ResourceSnapshotTaskHandler) snapshotConnections(ctx context.Context, sampledAt time.Time) (int, error) {
```
rmorlok, line 118: This task only snapshots connections and actors -- what about connectors, namespaces, rate limiters? Are those in follow up tasks?

If it isn't already covered in follow up tasks, and if there isn't data store definitions for those, please create a follow up task to add that support end-to-end.

rmorlok: Good catch. This PR intentionally follows the current storage surface from #404, which only introduced actor and connection sample tables, so the worker is limited to those two resource types for now. I created follow-up #435 to add connectors/connector versions, namespaces, and rate limiters end-to-end across storage, snapshots, query execution, API/docs, and tests.

PR #437, opened 2026-05-29
SetupFlowStep.Type (form/redirect) + setup_token + /setup/{advance,abort} endpoints (#367)

### LINKED ISSUE #367: Add SetupFlowStep.Type (form/redirect) + RETURN_ADVANCE/ABORT signed tokens + setup advance/abort endpoints
Part of #360. Depends on #365 (id-based step encoding).

## Goal

Generalize ``SetupFlowStep`` to support redirect steps in addition to form steps. Add a signed, one-time-use token mechanism (``RETURN_ADVANCE`` / ``RETURN_ABORT``) that 3rd parties can use to bounce the user back to the marketplace with the connection setup advanced or aborted.

## Today

``SetupFlowStep`` only describes form steps (``JsonSchema`` + ``UiSchema``). The OAuth2 redirect is hardcoded in core (``service_connection_initiate.go::initiateAuthStep`` calls ``SetStateAndGeneratePublicUrl``); it is not a first-class step type. There is no generic mechanism for a non-OAuth2 connector that needs to redirect the user to a 3rd party which then bounces back.

### 1. Schema extension

``internal/schema/connectors/setup_flow.go``:

```go
type SetupFlowStepType string

const (
    SetupFlowStepTypeForm     SetupFlowStepType = "form"
    SetupFlowStepTypeRedirect SetupFlowStepType = "redirect"
)

type SetupFlowStep struct {
    Id          string             ``json:"id" yaml:"id"``
    Type        SetupFlowStepType  ``json:"type,omitempty" yaml:"type,omitempty"`` // default: form
    Title       string             ``...``
    Description string             ``...``

    // Form-only fields
    JsonSchema  common.RawJSON     ``json:"json_schema,omitempty" yaml:"json_schema,omitempty"``
    UiSchema    common.RawJSON     ``json:"ui_schema,omitempty" yaml:"ui_schema,omitempty"``
    DataSources map[string]DataSourceDef ``...``

    // Redirect-only fields
    Redirect    *SetupFlowStepRedirect ``json:"redirect,omitempty" yaml:"redirect,omitempty"``
}

type SetupFlowStepRedirect struct {
    // URL is a mustache template. Supports placeholders for connection cfg
    // ({{cfg.X}}, as elsewhere) plus two synthetic placeholders:
    //   - {{RETURN_ADVANCE}}: replaced with a signed one-time-use token-bearing
    //     URL the 3rd party should redirect the user to after the user
    //     completes the off-platform step.
    //   - {{RETURN_ABORT}}: same shape, used when the user cancels.
    URL string `` ``
}
```

Validation:

- ``Type == "form"`` requires ``JsonSchema``; rejects ``Redirect``.
- ``Type == "redirect"`` requires ``Redirect.URL``; rejects ``JsonSchema``/``UiSchema``/``DataSources``.
- Defaults to ``form`` if absent (preserves backward compat for existing YAMLs).
- ``Redirect.URL`` mustache references are validated against the available config fields at that point in the flow (same machinery as ``DataSourceProxyRequest.ValidateMustacheReferences``).
- Plus existing ``apxy:`` reserved-prefix rule from #365.

### 2. Signed-token machinery

New package ``internal/setup_token`` (or under ``internal/apauth/jwt`` if that fits better):

```go
type Intent string

const (
    IntentAdvance Intent = "advance"
    IntentAbort   Intent = "abort"
)

type Claims struct {
    ConnectionId apid.ID
    StepId       string  // the step the token is anchored to
    Intent       Intent
    Jti          string  // one-time-use nonce, redis-tracked
}

// Mint signs a one-time-use JWT for the given intent + connection + step.
// nonceStore: apredis.Client. ttl: short (1-15 min) — long enough to survive
// the user finishing the 3rd-party flow, short enough that a leaked token
// expires before it can be abused.
func Mint(ctx context.Context, signer jwt.Signer, store apredis.Client, claims Claims, ttl time.Duration) (string, error)

// VerifyAndConsume returns the verified claims and atomically marks the jti
// as consumed (SET NX in redis). Replays return ErrAlreadyConsumed.
func VerifyAndConsume(ctx context.Context, verifier jwt.Verifier, store apredis.Client, token string) (Claims, error)
```

Signing key: reuse ``system_auth.jwt_signing_key`` per existing convention.

### 3. Public endpoints

In ``internal/service/public`` (mirrors the OAuth2 callback's home):

- ``POST /public/connections/{id}/setup/advance`` — takes ``?token=...`` (or body), verifies + consumes, then calls the connection's flow to advance past the current step. Redirects to the connection's ``return_to_url`` (with ``setup=pending`` if more steps remain, matching the OAuth2 callback's convention).
- ``POST /public/connections/{id}/setup/abort`` — same shape but transitions the connection through the abort path (same as ``AbortConnection``).

Both endpoints are unauthenticated to the AuthProxy session — the token *is* the authorization.

### 4. Redirect step rendering

When core builds a ``ManifestSetupStep`` for a schema-defined redirect step, its ``RenderRedirect`` resolves the URL template:

- ``{{cfg.X}}`` — standard mustache, against the connection's mustache context.
- ``{{RETURN_ADVANCE}}`` — minted on the fly via ``setup_token.Mint(IntentAdvance, …)``.
- ``{{RETURN_ABORT}}`` — minted on the fly via ``setup_token.Mint(IntentAbort, …)``.

The minted tokens are bound to a URL like ``{public_base}/connections/{id}/setup/advance?token=…`` — the 3rd party redirects the user to that URL when they're done.

### 5. Tests

- Unit tests for ``setup_token.Mint`` / ``VerifyAndConsume`` (happy path, replay, expired, wrong intent, signature mismatch).
- Schema validation tests for the new step type.
- Integration test (placeholder fixture connector with a redirect step) end-to-end: initiate → renders redirect URL with tokens → POST advance → connection transitions to next step. Reuse the chromedp test harness if practical, or use a programmatic POST against the public server.

## Out of scope

- Wiring OAuth2 through this mechanism (#368).
- UI changes for redirect steps in the marketplace beyond what the existing redirect-handling code does (the existing ``ConnectionSetupRedirect`` response type carries the URL — the UI just needs to navigate to it).

## Acceptance

- A connector YAML with a ``type: redirect`` step validates and runs end-to-end against the new advance/abort endpoints.
- Token replay returns 403; expired tokens return 401.
- ``./scripts/preflight.sh`` clean.

rmorlok:
Closes #367. Part of #360.

## Summary

Connectors can now declare redirect-type setup steps that send the user to a 3rd party for an off-platform interaction (e.g. a vendor-specific install flow that doesn't speak OAuth2). The 3rd party bounces the user back through ``/public/connections/{id}/setup/advance`` or ``.../abort`` with a signed one-time-use token that authorizes the transition.

## Schema

- ``SetupFlowStepType`` (``form`` | ``redirect``). Defaults to ``form`` when empty.
- ``SetupFlowStepRedirect{URL}`` — the mustache-templated redirect destination.
- ``SetupFlowStep`` gains ``Type`` + ``Redirect`` fields.
- Validation enforces required/forbidden fields per kind: ``form`` requires ``json_schema`` and rejects ``Redirect``; ``redirect`` requires ``Redirect.URL`` and rejects ``json_schema`` / ``ui_schema`` / ``data_sources``.

## setup_token package

New ``internal/setup_token``:
- ``Mint(ctx, redis, encrypt, MintInput, ttl)`` returns a fresh token bound to ``{connection_id, step_id, intent, return_to_url, jti}``. Stores the claims AES-GCM-encrypted under ``setup_token:<jti>`` in Redis with the supplied TTL.
- ``VerifyAndConsume`` atomically reads + deletes the Redis row (``GETDEL``) and decrypts the claims. Replays return ``ErrNotFound``; tampering returns ``ErrTampered``.
- New apid prefix ``stk_`` for token ids.

## Public endpoints

New ``internal/routes/public_setup.go``:
- ``POST/GET /public/connections/{id}/setup/advance`` — ``VerifyAndConsume``, validate intent + connection-id binding + stale-step check, then call ``HandleCredentialsEstablished`` on the connection. Redirects to the claims' ``ReturnToUrl`` with ``connection_id`` (+ ``setup=pending`` when more steps remain).
- ``POST/GET /public/connections/{id}/setup/abort`` — same validation path, then ``AbortConnection``. Redirects to the return URL.
- No session required; the token IS the authorization.

Wired into the public server alongside the OAuth2 callback routes.

## Redirect step rendering

``internal/core/connection_setup_flow.go``:
- ``newSchemaStep`` dispatches by ``spec.Type`` — form steps reuse the existing merge-into-config closure; redirect steps get a new closure that renders the URL template with cfg mustache + freshly-minted ``{{RETURN_ADVANCE}}`` / ``{{RETURN_ABORT}}`` substitutions.
- ``mintReturnURL`` builds the ``/public/connections/{id}/setup/{advance,abort}?token=...`` URL from the public service's base URL plus the minted jti.
- 15-minute token TTL.

## Test plan

- [x] ``go test ./...`` clean.
- [x] ``./scripts/preflight.sh`` clean.
- [x] New tests in ``internal/setup_token/token_test.go`` cover round-trip, validation, replay / tampered / expired / empty rejection paths.
- [x] New schema tests in ``internal/schema/resources/connectors/setup_flow_test.go`` cover per-kind required/forbidden field validation + the empty-type defaults-to-form path.
- [x] Integration: ``./api_key/`` and ``./probes/`` still pass.

### DIFF
--- a/internal/core/setup_token/token.go
+++ b/internal/core/setup_token/token.go
@@ -1,7 +1,7 @@
 // Package setup_token implements signed, one-time-use redirect tokens used
 // by the schema-defined redirect-step machinery. A token represents an
 // authorization for a 3rd party to bounce a user back through one of the
-// /public/connections/{id}/setup/{advance,abort} endpoints to transition a
+// /setup/connections/{id}/{advance,abort} endpoints to transition a
 // connection's setup_step.
 //
 // Wire format: opaque base64-encoded UUID (the jti). The claim payload is
--- a/internal/core/setup_token/token.go
+++ b/internal/core/setup_token/token.go
@@ -41,10 +41,11 @@ const (
 func (i Intent) IsValid() bool { return i == IntentAdvance || i == IntentAbort }
 
 // Claims is the verified payload a /setup/advance or /setup/abort handler
-// reads after VerifyAndConsume succeeds. ConnectionId + StepId are pinned
-// at mint time so a token can only act on the specific connection/step
-// pair it was issued for; ReturnToUrl carries the marketplace's eventual
-// destination through the off-platform bounce.
+// reads after VerifyAndConsume succeeds. ConnectionId + StepId + ActorId
+// are all pinned at mint time so a token can only act on the specific
+// connection/step pair it was issued for and only when consumed by the
+// same actor that initiated the flow; ReturnToUrl carries the
+// marketplace's eventual destination through the off-platform bounce.
 type Claims struct {
 	// ConnectionId is the connection the token authorizes a transition on.
 	ConnectionId apid.ID `json:"connection_id"`
@@ -54,6 +55,13 @@ type Claims struct {
 	// from this id between mint and consume, defending against a stale
 	// token unexpectedly advancing past an unrelated step.
 	StepId string `json:"step_id"`
+
+	// ActorId is the actor that initiated the redirect step. The /advance
+	// and /abort endpoints require a session and reject when the
+	// session's actor doesn't match this claim — defends against a
+	// leaked token being used by a different user even before its TTL
+	// expires.
+	ActorId apid.ID `json:"actor_id"`
 
 	// Intent gates which endpoint can consume the token.
 	Intent Intent `json:"intent"`
@@ -83,6 +91,7 @@ var ErrTampered = errors.New("setup_token: tampered payload")
 type MintInput struct {
 	ConnectionId apid.ID
 	StepId       string
+	ActorId      apid.ID
 	Intent       Intent
 	ReturnToUrl  string
 }
@@ -94,6 +103,9 @@ func (m MintInput) Validate() error {
 	}
 	if m.StepId == "" {
 		return errors.New("setup_token: step_id is required")
+	}
+	if m.ActorId == apid.Nil {
+		return errors.New("setup_token: actor_id is required")
 	}
 	if !m.Intent.IsValid() {
 		return fmt.Errorf("setup_token: invalid intent %q", m.Intent)
@@ -118,6 +130,7 @@ func Mint(ctx context.Context, r apredis.Client, e encrypt.E, in MintInput, ttl 
 	claims := Claims{
 		ConnectionId: in.ConnectionId,
 		StepId:       in.StepId,
+		ActorId:      in.ActorId,
 		Intent:       in.Intent,
 		ReturnToUrl:  in.ReturnToUrl,
 		Jti:          jti,
--- a/internal/routes/public_setup.go
+++ b/internal/routes/public_setup.go
@@ -1,0 +1,220 @@
+package routes
+
+import (
+	"errors"
+	"fmt"
+	"log/slog"
+	"net/http"
+	"net/url"
+
+	"github.com/gin-gonic/gin"
+
+	"github.com/rmorlok/authproxy/internal/apid"
+	"github.com/rmorlok/authproxy/internal/apredis"
+	"github.com/rmorlok/authproxy/internal/config"
+	"github.com/rmorlok/authproxy/internal/core/iface"
+	"github.com/rmorlok/authproxy/internal/encrypt"
+	sconfig "github.com/rmorlok/authproxy/internal/schema/config"
+	"github.com/rmorlok/authproxy/internal/setup_token"
+)
+
+// PublicSetupRoutes exposes the /public/connections/{id}/setup/advance and
+// .../abort endpoints. The endpoints are unauthenticated to the AuthProxy
+// session — the setup token (carried as ?token=...) is the authorization.
+//
+// Flow: a connector YAML declares a redirect step whose URL template
+// includes {{RETURN_ADVANCE}} / {{RETURN_ABORT}}. At redirect-step render
+// time core mints a token per placeholder, embedding the connection id +
+// the current setup_step id + the original _initiate return-to URL. The
+// 3rd party redirects the user back through one of these endpoints when
+// they're done; we verify + consume the token, advance/abort the
+// connection, and bounce the user to the marketplace return URL.
+type PublicSetupRoutes struct {
+	cfg     config.C
+	core    iface.C
+	r       apredis.Client
+	encrypt encrypt.E
+	logger  *slog.Logger
+}
+
+// NewPublicSetupRoutes constructs the public setup-transition routes. All
+// dependencies come from the DependencyManager — no per-instance state.
+func NewPublicSetupRoutes(
+	cfg config.C,
+	core iface.C,
+	r apredis.Client,
+	encrypt encrypt.E,
+	logger *slog.Logger,
+) *PublicSetupRoutes {
+	return &PublicSetupRoutes{
+		cfg:     cfg,
+		core:    core,
+		r:       r,
+		encrypt: encrypt,
+		logger:  logger,
+	}
+}
+
+// Register binds the two endpoints to the gin engine. Both accept GET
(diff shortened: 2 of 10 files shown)


internal/core/setup_token/token.go
```diff
@@ -0,0 +1,174 @@
...
+// /public/connections/{id}/setup/{advance,abort} endpoints to transition a
+// connection's setup_step.
+//
+// Wire format: opaque base64-encoded UUID (the jti). The claim payload is
+// stored encrypted in Redis under that jti — verifying = looking up the
+// jti, decrypting the payload, and atomically deleting the row so the
+// token can only be used once. The redis-as-source-of-truth shape mirrors
+// the OAuth2 state machinery and gets us authentication (via encrypt.E's
+// AES-GCM) plus one-time-use for free.
+package setup_token
```
rmorlok, line 13: Could we please move this inside the core package? So `internal/core/setup_token`?

internal/core/setup_token/token.go
```diff
@@ -0,0 +1,174 @@
...
+
+// IsValid reports whether i is a known intent.
+func (i Intent) IsValid() bool { return i == IntentAdvance || i == IntentAbort }
+
+// Claims is the verified payload a /setup/advance or /setup/abort handler
+// reads after VerifyAndConsume succeeds. ConnectionId + StepId are pinned
+// at mint time so a token can only act on the specific connection/step
+// pair it was issued for; ReturnToUrl carries the marketplace's eventual
+// destination through the off-platform bounce.
+type Claims struct {
```
rmorlok, line 49: Please make a particular claim bound to an actor. So this would require a session on the endpoint that consumes the advance/abort and the claim to store that value and then be compared to the claimed actor.

I'm nervous about these tokens being used in attack flows event though they are short lived.

rmorlok:
Addressed both review comments in f5026a5:

- Moved `internal/setup_token` → `internal/core/setup_token` (tightly coupled to core's setup flow; nothing outside core needs to import it directly).
- Bound tokens to an actor: Claims gained `ActorId`; Mint requires it; the redirect-step closure reads the initiating actor from request context and embeds it. The `/setup/advance` and `/setup/abort` endpoints now require a session (matches the OAuth2 callback shape: unauthenticated browsers bounce through session-initiate and return) and reject when the session actor mismatches the claim. A leaked token used by a different signed-in user fails the actor check before its TTL expires.

rmorlok (commented):
In your description of the PR:

> No session required; the token IS the authorization.

I don't think this should be correct. While we have the 15 minute limit on these tokens, if they do provide authorize

internal/routes/public_setup.go
```diff
@@ -0,0 +1,259 @@
...
+	mw := h.authService.NewRequiredBuilder().
+		ForResource("connections").
+		ForVerb("create").
+		WithRedirectOnUnauthenticated(h.sessionInitiateUrlGenerator).
+		Build()
+
+	g.GET("/public/connections/:id/setup/advance", mw, h.advance)
+	g.POST("/public/connections/:id/setup/advance", mw, h.advance)
+	g.GET("/public/connections/:id/setup/abort", mw, h.abort)
+	g.POST("/public/connections/:id/setup/abort", mw, h.abort)
```
rmorlok: Why is the path here `/public/...`? The host implies public. Let's do `/setup/connections/:id/{abort,advance}` instead.

rmorlok: Done in ee06d5e — paths are now `/setup/connections/{id}/advance` and `/setup/connections/{id}/abort`.

PR #438, opened 2026-05-29
Execute resource app metrics queries

### LINKED ISSUE #409: App metrics: execute resource metrics queries
Parent: #395

## Goal
Implement metrics query execution over resource snapshot samples.

## Scope
- Support resources.connections.count.
- Support resources.actors.count.
- Support time bucketing over the requested range/step.
- Support namespace matchers and existing label selector syntax.
- Support initial connection group_by dimensions: state, health_state, connector_id, connector_version.
- Support actor grouping by namespace initially.
- Ensure query semantics use stored samples, not current live DB state.

## Acceptance Criteria
- Query responses return time series suitable for Admin UI charts.
- Historical buckets continue to show resources that existed then even if they were later deleted.
- Namespace and label filtering are covered by tests.
- Multiple group_by series are returned with stable labels and ref_ids.

rmorlok:
## Summary
- adds resource metrics query execution for connection and actor counts over app metrics samples
- wires /metrics/query to dispatch request-event and resource metric refs through the shared API response shape
- covers namespace/label filters, stable grouped series, distinct resource counting, and historical buckets from stored samples

Closes #409

## Tests
- env GOCACHE=/private/tmp/authproxy-go-cache AUTH_PROXY_TEST_DATABASE_PROVIDER=sqlite AUTH_PROXY_APP_METRICS_TEST_DATABASE_PROVIDER=sqlite go test ./internal/app_metrics ./internal/routes
- ./scripts/preflight.sh

### DIFF
--- a/internal/routes/request_events.go
+++ b/internal/routes/request_events.go
@@ -275,23 +275,10 @@ func metricsQueryToResourceQuery(
 	groupBy := make([]app_metrics.ResourceGroupBy, 0, len(ref.GroupBy))
 	for _, raw := range ref.GroupBy {
 		gb := app_metrics.ResourceGroupBy(raw)
-		switch metric {
-		case app_metrics.ResourceMetricConnectionsCount:
-			switch gb {
-			case app_metrics.ResourceGroupByState,
-				app_metrics.ResourceGroupByHealthState,
-				app_metrics.ResourceGroupByConnectorID,
-				app_metrics.ResourceGroupByConnectorVersion:
-				groupBy = append(groupBy, gb)
-			default:
-				return app_metrics.ResourceMetricsQuery{}, httperr.BadRequestf("invalid group_by %q", raw)
-			}
-		case app_metrics.ResourceMetricActorsCount:
-			if gb != app_metrics.ResourceGroupByNamespace {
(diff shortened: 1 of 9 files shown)


internal/routes/request_events.go
```diff
@@ -260,35 +261,161 @@ func requestEventMetricFromAPI(metric, aggregation string) (app_metrics.RequestE
...
 
-	out := sapi.MetricsQueryResponseJson{
-		Series: make([]sapi.MetricsSeriesJson, 0, len(series)),
+	groupBy := make([]app_metrics.ResourceGroupBy, 0, len(ref.GroupBy))
+	for _, raw := range ref.GroupBy {
+		gb := app_metrics.ResourceGroupBy(raw)
+		switch metric {
+		case app_metrics.ResourceMetricConnectionsCount:
+			switch gb {
+			case app_metrics.ResourceGroupByState,
```
rmorlok: Is there an opportunity here to simplify this by using `isValidResourceGroupBy` and returning an error in that case? I think you could call that, return error if found, and then just do groupBy = append(groupBy, gb).

rmorlok: Done in aca85a3. I exposed the resource group-by validator from app_metrics and simplified the route conversion to validate with IsValidResourceGroupBy before appending the dimension.

PR #459, opened 2026-05-30
Support OAuth2 client credentials grant

### LINKED ISSUE #352: Support OAuth2 client-credentials grant (RFC 6749 §4.4)
WIP

Parent project: #159
Spec: RFC 6749 §4.4 (Client Credentials Grant), §4.4.3 (response — no refresh token)
Related: #351 (token-endpoint auth method — required by this flow)

## Goal

Add support for the **Client Credentials Grant** (RFC 6749 §4.4) as a new OAuth2 flow. Today the proxy implements only the authorization_code grant: every OAuth2 connection routes through an end-user authorize redirect, a callback with `code`, and a code-for-token exchange. The client-credentials grant has no end-user — the client_id+client_secret are themselves the identity, exchanged directly at the token endpoint for an access token used per request.

This is a first-class need for service-to-service integrations where the customer's backend talks to a provider's API on its own behalf (no end-user involved). Today these integrations have no path through the proxy.

## Design choice: grant selector vs. new auth type

Two viable structures:

1. **Grant selector inside `AuthOAuth2`** — add `grant_type: authorization_code | client_credentials` (default `authorization_code`). Reuses `client_id`/`client_secret`/`token`/`scopes`/`revocation`/`token_endpoint_auth_method` fields. Validator gates which sub-fields are required vs. forbidden per grant.

2. **New top-level `AuthImpl`** — `AuthOAuth2ClientCredentials` alongside `AuthOAuth2`/`AuthApiKey`/`AuthNoAuth`. Cleaner field surface (no irrelevant Authorization/PKCE/refresh fields) at the cost of duplicating token-endpoint plumbing.

**Use option (1).** The wire-level mechanics (token endpoint POST, client-auth method from #351, scope encoding, access-token storage, expiry-driven re-acquisition, revocation) are identical between the two grants. The differences (no authorize URL, no callback, no refresh_token, no PKCE) are best expressed as conditional validation, not as a parallel type. Token refresh especially benefits from sharing: a client_credentials access token nearing expiry is renewed by re-running the same grant against the same token endpoint with the same client auth — that's a one-line difference from the refresh_token grant we already have.

### 1. Connector schema

`internal/schema/connectors/auth_oauth2.go` — add a grant selector on `AuthOAuth2`:

```go
type OAuth2GrantType string

const (
    OAuth2GrantAuthorizationCode = OAuth2GrantType(\"authorization_code\")
    OAuth2GrantClientCredentials = OAuth2GrantType(\"client_credentials\")
)

type AuthOAuth2 struct {
    Type      AuthType
    GrantType *OAuth2GrantType  // new — nil defaults to authorization_code
    // ...
}
```

Validator (`AuthOAuth2.Validate`):
- Reject unknown grants.
- For `client_credentials`:
  - `client_id` and `client_secret` MUST be asent - must be supplied as credentials.
  - `Authorization` block (endpoint, PKCE, query_overrides) MUST be absent — no authorize URL is used.
  - `Revocation` allowed (RFC 7009 still applies).
  - `Token` endpoint required (same as today).
  - Refresh-token-specific config (`refresh_tokens_in_background`, `refresh_tokens_cron_schedule`, etc.) allowed but interpreted as *re-grant* schedule rather than refresh-token grant.
- For `authorization_code` (current default): `Authorization` required, no change from today.

### 2. Setup flow

`internal/schema/connectors/setup_flow.go` — the `auth` phase today means "OAuth2 authorize redirect, wait for callback". That phase has no meaning for client_credentials. Two options:

- **(a)** Reuse the `auth` phase but special-case its semantics by grant: for client_credentials, instead of building a redirect URL it performs the synchronous token-endpoint POST inside `initiateAuthStep`. The transition out is identical (HandleCredentialsEstablished → verify → configure).
- **(b)** Add a new `token_exchange` phase that runs instead of `auth` when grant is client_credentials.

Recommend (a). The phase model already proved (with API-key) that a phase doesn't need to literally redirect — it just needs to materialize credentials. The auth-method-specific code in `service_connection_submit.go::initiateAuthStep` is already a type switch on the inner `AuthImpl`, so an `auth.GrantType` check there has a natural home.

`NextSetupStep`'s preconnect→auth→verify transition stays unchanged.

### 3. New auth-method entry points

`internal/auth_methods/oauth2/` — currently the package is wired around `SetStateAndGeneratePublicUrl` + `HandleAuthCallback`. Add a synchronous code path:

```go
// ExchangeClientCredentials performs the RFC 6749 §4.4 token-endpoint POST,
// persists the resulting access token (no refresh_token), and triggers
// HandleCredentialsEstablished. Returns the same setup response shape that
// service_connection_submit.go expects.
func (o *OAuth2) ExchangeClientCredentials(ctx context.Context) (Outcome, error)
```

Implementation reuses the existing `postTokenExchangeWithRetry`, `createDbTokenFromResponse`, `classifyTokenEndpointStatus`, `emitAndRecordExchangeFailure` helpers. Body shape:

```
grant_type=client_credentials
scope=<configured scopes, joined with scope_separator>
```

Client auth is applied via the same method-branching helper that #351 introduces — Basic header / form body / none. No `redirect_uri`, no `code`, no `code_verifier`.

Per RFC 6749 §4.4.3, the response MUST NOT include a refresh_token. If the provider sends one, log a warning and discard it — never persist. (Belt-and-suspenders, since some providers do this in violation of the spec.)

### 4. Token refresh (re-grant)

`internal/auth_methods/oauth2/proxy.go` (the refresh path) — branch on grant:

- `authorization_code` → today's behavior: grant_type=refresh_token, send stored refresh_token.
- `client_credentials` → grant_type=client_credentials, no refresh_token, identical to the initial exchange.

`internal/core/iface` already has refresh entry points (background refresh job, on-demand refresh in the proxy roundtripper); both need to flow through the grant-aware refresh helper. The connection's stored token model doesn't change — still `oauth2_tokens` row with `access_token` and `expires_at`. The `encrypted_refresh_token` column stays nullable; for client_credentials it's always NULL.

The \"no refresh_token, just re-grant\" branch is functionally the same as today's \"refresh_token is required for refresh; fail unhealthy if absent\" branch — except for client_credentials it isn't unhealthy, it just re-runs the grant. The connection-state machine needs to know the difference. Cleanest place to express it: a `(*OAuth2).supportsRefreshToken()` boolean used at both the failure-classification site (\"no refresh_token present\") and the refresh-mechanics site.

### 5. Token revocation

Existing revocation flow works as-is — client_credentials tokens are revoked at the same revocation endpoint with the same Authorization header. No code changes needed beyond verifying the existing revocation suite covers \"connector with no Authorization block\" (i.e., a client_credentials connector with revocation configured).

### 6. Background refresh scheduling

The existing `refresh_tokens_cron_schedule` / `refresh_tokens_in_background` config drives the periodic refresh task. The task today filters connections that have a refresh_token — for client_credentials connections it needs to instead filter on \"grant_type = client_credentials AND access_token is near expiry\". One filter swap in the cron task, no schema change.

### 7. Error surfaces

The provider-side failures for client_credentials map to existing token-exchange categories:
- 401 invalid_client → `tokenExchangeInvalidClient` (existing).
- 400 invalid_scope → existing category.
- 400 unauthorized_client → existing category covers \"client not allowed to use this grant\".
- 5xx → existing `tokenExchangeServerError`.

No new categories.

### 8. Probes

Probes (`HasProbes`) already work against the stored access token via the proxy roundtripper. No change.

### 9. Tests

- Unit tests for `ExchangeClientCredentials` happy path (S256 stub provider via gentleman/gock).
- Unit tests for re-grant on expiry (mirror of `proxy_test.go`'s refresh tests).
- Validator tests rejecting `authorization` block when grant is client_credentials, requiring `client_id`/`client_secret` (or `none` method), etc.
- Integration test in `integration_tests/oauth2/` against `go-oauth2-server` (extends the existing test infra; if the test server doesn't expose client_credentials, file the gap per #159's process).
- Schema fixture YAMLs under `internal/schema/connectors/test_data/` exercising valid+invalid configurations.

## Acceptance Criteria

- Connector YAML declaring `grant_type: client_credentials` with `client_id`+`client_secret` produces a connection that hits the token endpoint once on create (no authorize URL, no callback), stores the resulting access token, and uses it on per-request proxying.
- No `refresh_token` is persisted even if the provider returns one (warning logged).
- Token-near-expiry triggers a re-grant against the token endpoint using `grant_type=client_credentials`, applying the connector's configured token-endpoint auth method from #351.
- Validator rejects: unknown grant types; `client_credentials` connectors that include an `authorization` block; `client_credentials` connectors missing `client_id`/`client_secret` (unless token-endpoint method is `none`).
- Existing `authorization_code` connectors are unaffected — `grant_type` defaults to authorization_code, all current tests pass.
- Probes succeed against a client_credentials connection.
- Revocation works against a client_credentials connection that declares a revocation endpoint.

## Out of Scope

- Resource Owner Password Credentials grant (RFC 6749 §4.3) — deprecated by OAuth 2.1, no plans to support.
- Device Authorization Grant (RFC 8628) — separate ticket.
- JWT-bearer / SAML-bearer assertion grants (RFC 7521/7522/7523) — separate ticket.
- Per-connection (rather than per-connector) client_id/client_secret storage — out of scope per the BYO-via-connector-API model above.

## Notes

- Depends on #351 for proper client-auth method support. Without #351 this ticket would inherit the same `client_secret_post`-only limitation; some client_credentials providers (notably Microsoft Graph) require Basic.
- The two big architectural decisions in this ticket (grant selector vs. new AuthImpl; reuse `auth` phase vs. new `token_exchange` phase) are worth a quick conversation before implementation begins — what's recommended above isn't the only defensible read.

rmorlok:
## Summary
- Add OAuth2 grant_type selection with client_credentials validation and schema support
- Add synchronous client-credentials setup exchange and re-grant refresh path
- Discard refresh_token values returned for client_credentials responses

## Tests
- go test ./internal/schema/resources/connectors ./internal/auth_methods/oauth2 ./internal/core
- ./scripts/preflight.sh

Closes #352

### DIFF
--- a/internal/auth_methods/oauth2/callback.go
+++ b/internal/auth_methods/oauth2/callback.go
@@ -85,11 +85,15 @@ func (o *oAuth2Connection) CallbackFrom3rdParty(ctx context.Context, query url.V
 // dashboards and alerts can key off `category` rather than parsing error strings.
 func (o *oAuth2Connection) exchangeCodeAndAdvance(ctx context.Context, query url.Values) (string, error) {
 	var returnURL string
-	err := o.tel.withSpan(ctx, "token_exchange", o.connectorIDForTelemetry(), func(ctx context.Context) error {
-		var err error
-		returnURL, err = o.exchangeCodeAndAdvanceInner(ctx, query)
-		return err
-	})
+	err := o.tel.withSpan(
+		ctx,
+		"token_exchange",
+		o.connectorIDForTelemetry(),
+		func(ctx context.Context) error {
+			var err error
+			returnURL, err = o.exchangeCodeAndAdvanceInner(ctx, query)
+			return err
+		})
 	return returnURL, err
 }
 
@@ -97,7 +101,11 @@ func (o *oAuth2Connection) exchangeCodeAndAdvance(ctx context.Context, query url
 // failure-counter bump for token-exchange failures. Mirrors the refresh-side
 // classifyAndRecordRefreshFailure: every callsite uses the same shape so the
 // "category" string lands consistently on logs, metrics, and span errors.
-func (o *oAuth2Connection) emitAndRecordExchangeFailure(ctx context.Context, category tokenExchangeCategory, attrs tokenExchangeAttrs) {
+func (o *oAuth2Connection) emitAndRecordExchangeFailure(
+	ctx context.Context,
+	category tokenExchangeCategory,
+	attrs tokenExchangeAttrs,
+) {
 	emitTokenExchangeFailure(ctx, o.logger, category, attrs)
 	o.tel.recordTokenExchangeFailure(ctx, string(category), o.connectionLabelsForTelemetry())
 }
@@ -231,9 +239,13 @@ func (o *oAuth2Connection) exchangeCodeAndAdvanceInner(ctx context.Context, quer
 // endpoint exchange for service-to-service connectors. There is no authorize
 // redirect or callback; the client credentials are the grant.
 func (o *oAuth2Connection) ExchangeClientCredentials(ctx context.Context) error {
-	return o.tel.withSpan(ctx, "token_exchange", o.connectorIDForTelemetry(), func(ctx context.Context) error {
-		return o.exchangeClientCredentialsInner(ctx)
-	})
+	return o.tel.withSpan(
+		ctx,
+		"token_exchange",
+		o.connectorIDForTelemetry(),
+		func(ctx context.Context) error {
+			return o.exchangeClientCredentialsInner(ctx)
+		})
 }
 
 func (o *oAuth2Connection) exchangeClientCredentialsInner(ctx context.Context) error {
@@ -299,12 +311,6 @@ func (o *oAuth2Connection) exchangeClientCredentialsInner(ctx context.Context) e
 	if err != nil {
 		err = fmt.Errorf("failed to create db token from response: %w", err)
 		o.emitAndRecordExchangeFailure(ctx, tokenExchangeMalformedResponse, o.tokenExchangeAttrsFromConn(err))
-		return err
-	}
-
-	if _, err := o.connection.HandleCredentialsEstablished(ctx); err != nil {
-		err = fmt.Errorf("failed to handle post-auth state transition: %w", err)
-		o.emitAndRecordExchangeFailure(ctx, tokenExchangeInternalError, o.tokenExchangeAttrsFromConn(err))
 		return err
 	}
 
--- a/internal/auth_methods/oauth2/callback.go
+++ b/internal/auth_methods/oauth2/callback.go
@@ -227,6 +227,91 @@ func (o *oAuth2Connection) exchangeCodeAndAdvanceInner(ctx context.Context, quer
 	return o.state.ReturnToUrl, nil
 }
 
+// ExchangeClientCredentials performs RFC 6749 §4.4's synchronous token
+// endpoint exchange for service-to-service connectors. There is no authorize
+// redirect or callback; the client credentials are the grant.
+func (o *oAuth2Connection) ExchangeClientCredentials(ctx context.Context) error {
+	return o.tel.withSpan(ctx, "token_exchange", o.connectorIDForTelemetry(), func(ctx context.Context) error {
+		return o.exchangeClientCredentialsInner(ctx)
+	})
+}
+
+func (o *oAuth2Connection) exchangeClientCredentialsInner(ctx context.Context) error {
+	c := o.httpf.
+		ForRequestType(httpf.RequestTypeOAuth).
+		ForConnection(o.connection).
+		New().
+		UseContext(ctx)
+
+	clientId, clientSecret, err := o.resolveClientCredentials(ctx)
+	if err != nil {
+		o.emitAndRecordExchangeFailure(ctx, tokenExchangeInternalError, o.tokenExchangeAttrsFromConn(err))
+		return err
+	}
+
+	tokenEndpoint, err := o.renderMustache(ctx, o.auth.Token.Endpoint)
+	if err != nil {
+		err = fmt.Errorf("failed to render token endpoint template: %w", err)
+		o.emitAndRecordExchangeFailure(ctx, tokenExchangeInternalError, o.tokenExchangeAttrsFromConn(err))
+		return err
+	}
+
+	values := url.Values{
+		"grant_type": {"client_credentials"},
+	}
+	if scopes := JoinScopes(o.auth.Scopes); scopes != "" {
+		values.Set("scope", scopes)
+	}
+
+	values, authHeader, err := applyTokenEndpointClientAuth(
+		o.auth.GetTokenEndpointAuthMethodOrDefault(), clientId, clientSecret, values,
+	)
+	if err != nil {
+		o.emitAndRecordExchangeFailure(ctx, tokenExchangeInternalError, o.tokenExchangeAttrsFromConn(err))
+		return err
+	}
+
+	for k, v := range o.auth.Token.FormOverrides {
+		values.Set(k, v)
+	}
+
+	resp, attempts, err := o.postTokenExchangeWithRetry(ctx, c, tokenEndpoint, values, authHeader)
+	if err != nil {
+		err = fmt.Errorf("failed to post client credentials token exchange: %w", err)
+		attrs := o.tokenExchangeAttrsFromConn(err)
+		attrs.Attempts = attempts
+		o.emitAndRecordExchangeFailure(ctx, tokenExchangeNetworkError, attrs)
+		return err
+	}
+
+	if resp.StatusCode != 200 {
+		category, providerErr := classifyTokenEndpointStatus(resp.StatusCode, resp.Bytes())
+		err := fmt.Errorf("received status code %d from client credentials token exchange", resp.StatusCode)
+		attrs := o.tokenExchangeAttrsFromConn(err)
+		attrs.ProviderStatusCode = resp.StatusCode
+		attrs.ProviderError = providerErr
+		attrs.Attempts = attempts
+		o.emitAndRecordExchangeFailure(ctx, category, attrs)
+		return err
+	}
+
+	_, err = o.createDbTokenFromResponseWithOptions(ctx, resp, nil, tokenPersistOptions{PersistRefreshToken: false})
+	if err != nil {
+		err = fmt.Errorf("failed to create db token from response: %w", err)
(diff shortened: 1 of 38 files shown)


internal/core/connection_setup_dispatch.go
```diff
@@ -34,6 +35,24 @@ func (c *connection) advanceToStep(
 		return c.completeFlow(ctx)
 	}
 
+	if next.Id() == oauth2.OAuth2ClientCredentialsStepId {
```
rmorlok: The core logic should not be reasoning about internal details about OAuth. OAuth should be rendering steps as appropriate so that core can handle the generic types.

rmorlok: Addressed in a4bf979: core no longer checks the OAuth client-credentials step id or type-asserts the OAuth factory. OAuth now emits a generic immediate manifest step with its own OnEnter handler, and core only dispatches the generic immediate step type before rendering the resulting setup state.

internal/core/iface/manifest_setup_step.go
```diff
@@ -10,13 +10,14 @@ import (
...
+// (api-key emits form steps for credential collection; OAuth2 emits redirect
+// or immediate steps depending on grant type).
 type ManifestStepType string
 
 const (
-	ManifestStepTypeForm     ManifestStepType = "form"
-	ManifestStepTypeRedirect ManifestStepType = "redirect"
+	ManifestStepTypeForm      ManifestStepType = "form"
+	ManifestStepTypeRedirect  ManifestStepType = "redirect"
+	ManifestStepTypeImmediate ManifestStepType = "immediate"
```
rmorlok: Please add comments on the `ManifestStepType` values to describe what those steps do (from a UI perspective, possibly with examples for how those would be used by oauth/api key/etc.

rmorlok: Addressed in ada32aa: added doc comments for form, redirect, and immediate manifest step types describing what each means to the UI and how auth methods/connectors use them (API-key form collection, OAuth authorization redirect, OAuth client_credentials immediate exchange).

internal/core/iface/manifest_setup_step.go
```diff
@@ -10,13 +10,32 @@ import (
...
+	// authorization-code setup uses this for the provider authorize URL, and
+	// connector-authored redirect steps can use it for external setup tasks.
 	ManifestStepTypeRedirect ManifestStepType = "redirect"
+
+	// ManifestStepTypeImmediate is a server-side step that runs as soon as
+	// the setup flow enters it. There is no form or redirect for the UI to
+	// render; the handler performs work and advances the connection to the
+	// next visible state. OAuth2 client_credentials uses this to exchange at
+	// the token endpoint without involving an end user.
+	ManifestStepTypeImmediate ManifestStepType = "immediate"
```
rmorlok: Instead of calling this immediate, let's call this `lambda` to imply that invocation is happening rather than UI visible action.

internal/auth_methods/oauth2/callback.go
```diff
@@ -227,6 +227,91 @@ func (o *oAuth2Connection) exchangeCodeAndAdvanceInner(ctx context.Context, quer
...
+}
+
+func (o *oAuth2Connection) exchangeClientCredentialsInner(ctx context.Context) error {
+	c := o.httpf.
+		ForRequestType(httpf.RequestTypeOAuth).
+		ForConnection(o.connection).
+		New().
+		UseContext(ctx)
+
+	clientId, clientSecret, err := o.resolveClientCredentials(ctx)
```
rmorlok, line 258: How is this working? Isn't this pulling the data from the configuration for the connector? For the client credentials auth type, the user is supposed to be prompted to enter the client id and secret as a form, then we exchange those internally. I don't see anything about pulling them from a stored form in this method.

My assumption is that as we received this data from the form, we would store the data in a credentials table, likely extending / renaming the table already used for API keys, because the client id/secret is very similar to an api key in this case.

internal/core/iface/manifest_setup_step.go
rmorlok: Addressed in c1d959d by removing the invoked setup-step type entirely. With the revised flow, client_credentials is just a regular form step, so core only handles the existing generic form/redirect/verify types.

internal/auth_methods/oauth2/callback.go
rmorlok: Addressed in c1d959d: client_credentials now renders a credentials form, persists the submitted client_id/client_secret into the encrypted credential row, and performs the token exchange from that stored credential. The connector schema now rejects connector-level client_id/client_secret for client_credentials to avoid silently reading static connector config.

internal/database/api_key_credential.go
rmorlok: Let's rename this table to `connection_credentials` to reflect the fact that it has cross-cutting credential types, not just API keys.

rmorlok: Addressed in 30cc0de: renamed the migration/table/indexes from api_key_credentials to connection_credentials and updated the DB table constant, encrypted-field registration, SQL tests, and comments that describe the shared credential storage.

internal/auth_methods/oauth2/client_credentials.go
```diff
@@ -0,0 +1,122 @@
...
+	apauthcore "github.com/rmorlok/authproxy/internal/apauth/core"
+	coreIface "github.com/rmorlok/authproxy/internal/core/iface"
+	"github.com/rmorlok/authproxy/internal/database"
+	"github.com/rmorlok/authproxy/internal/httperr"
+	"github.com/rmorlok/authproxy/internal/schema/common"
+	cschema "github.com/rmorlok/authproxy/internal/schema/resources/connectors"
+)
+
+func synthesizeClientCredentialsStep(method cschema.TokenEndpointAuthMethod) *cschema.SetupFlowStep {
+	type schemaProp struct {
```
rmorlok: These schema props, ui control, etc structs are to allow us to define JSON schema and UI schema to communicate them as part of the setup steps.

Let's move those to the internal/schema/common as two different subpackages: json_schema and ui_schema. That should make the naming of the individual structs clean in terms of naming. Please be sure to add readmes to each of those packages describing where the contract is sourced from.

rmorlok: Addressed in aff2956: moved the setup-form JSON Schema and JSON Forms UI Schema helper structs into internal/schema/common/json_schema and internal/schema/common/ui_schema, added README files documenting the source contracts for each package, added schema contract tests, and updated both OAuth2 client_credentials and API-key synthesized credential forms to use the shared helpers.

PR #517, opened 2026-06-06
Add actor-templated permission namespaces

### LINKED ISSUE #516: Support actor-templated permission namespaces
## Summary

Add support for templating actor values into permission namespace matchers so shared actor permission configuration can scope each actor to its own namespace.

## Desired behavior

Permission namespace values should support these template references:

- `{{external_id}}`
- `{{labels.<label>}}`
- `{{annotations.<annotation>}}`

Example:

```yaml
permissions:
  - namespace: root.{{external_id}}.**
    resources: ["*"]
    verbs: ["*"]
```

When evaluating permissions for an actor with `external_id: acme`, this should match `root.acme` and descendants.

If a permission references a label or annotation that the actor does not have, that permission namespace should never match. Missing template values must not collapse to an empty namespace or broaden access.

## Notes

This is especially useful for externally sourced actors where many actors share one permission template, but each should only be granted access under a namespace derived from their actor identity or metadata.

rmorlok:
## Summary

- Adds actor-based rendering for permission namespace matchers using `{{external_id}}`, `{{labels.<label>}}`, and `{{annotations.<annotation>}}`.
- Denies permission matches when referenced labels or annotations are missing, or when rendered values do not form a valid namespace matcher.
- Updates auth permission schema/docs and adds tests for matching, query namespace derivation, and schema validation.

Closes #516.

## Tests

- `go test ./internal/apauth/core`
- `go test ./internal/schema/auth`
- `go test ./internal/apauth/service`
- `go test ./internal/schema/config`
- `go test ./internal/apauth/...`
- `go test ./internal/schema/...`
- `./scripts/preflight.sh`

### DIFF
--- a/internal/apauth/core/actor.go
+++ b/internal/apauth/core/actor.go
@@ -13,6 +13,10 @@ type IActorData interface {
 	GetLabels() map[string]string
 }
 
+type actorAnnotationsData interface {
+	GetAnnotations() map[string]string
+}
+
 // Actor is the information that identifies who is making a request. This can be an actor in the calling
 // system, an admin from the calling system, a devops admin from the cli, etc.
 type Actor struct {
@@ -23,6 +27,7 @@ type Actor struct {
 	ExternalId  string               `json:"external_id"`
 	Namespace   string               `json:"namespace,omitempty"`
 	Labels      map[string]string    `json:"labels,omitempty"`
+	Annotations map[string]string    `json:"-"`
 	Permissions []aschema.Permission `json:"permissions"`
 }
 
@@ -44,14 +49,19 @@ func (a *Actor) GetNamespace() string {
 
 func (a *Actor) GetLabels() map[string]string { return a.Labels }
 
-// GetAnnotations always returns nil for JWT-derived actors. Annotations are
-// server-side metadata and do not propagate via JWT, so an upsert from a JWT
-// claim must not overwrite annotations on the stored actor.
-func (a *Actor) GetAnnotations() map[string]string { return nil }
+// GetAnnotations returns server-side actor annotations when this actor was
+// built from database state. The field is not deserialized from JWT actor
+// claims, so upserts from claims preserve existing annotations.
+func (a *Actor) GetAnnotations() map[string]string { return a.Annotations }
 
 func CreateActor(data IActorData) *Actor {
 	if a, ok := data.(*Actor); ok {
 		return a
+	}
+
+	var annotations map[string]string
+	if a, ok := data.(actorAnnotationsData); ok {
+		annotations = a.GetAnnotations()
 	}
 
 	return &Actor{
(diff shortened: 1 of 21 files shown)


internal/apauth/core/actor.go
```diff
@@ -23,6 +27,7 @@ type Actor struct {
 	ExternalId  string               `json:"external_id"`
 	Namespace   string               `json:"namespace,omitempty"`
 	Labels      map[string]string    `json:"labels,omitempty"`
+	Annotations map[string]string    `json:"-"`
```
rmorlok: Please have this pull from the JWT data the same as labels. Update any other places to make sure that upserted actors from JWT also carry this data. Add appropriate tests for it.

Not having annotation support here was a miss on the previous PR.

internal/apauth/core/permissions.go
```diff
@@ -1,11 +1,15 @@
...
 
 import (
+	"regexp"
 	"slices"
+	"strings"
 
 	aschema "github.com/rmorlok/authproxy/internal/schema/auth"
 )
 
+var permissionNamespaceTemplateRegex = regexp.MustCompile(`{{\s*([^{}]+?)\s*}}`)
```
rmorlok: Why is this templated separately? This should be using aptmpl.

internal/apauth/core/actor.go
rmorlok: Addressed in 4dcf889. Actor annotations now deserialize from JWT actor data via `json:"annotations,omitempty"`, the JWT claims/token builders have `WithAnnotations` / `WithAnnotation` support mirroring labels, and `establishAuthFromRequest` now upserts JWT-supplied annotations into the actor row. Added claims/token builder tests plus a service test covering annotation update through JWT auth.

internal/apauth/core/permissions.go
rmorlok: Addressed in 4dcf889. Permission namespace rendering now uses `aptmpl.ExtractVariables` and `aptmpl.RenderMustache`; the permission layer only pre-validates that referenced variables are one of `external_id`, `labels.*`, or `annotations.*` and that label/annotation keys exist before rendering, so missing metadata still fails closed.

internal/apauth/core/actor.go
```diff
@@ -13,6 +13,10 @@ type IActorData interface {
 	GetLabels() map[string]string
 }
 
+type actorAnnotationsData interface {
+	GetAnnotations() map[string]string
```
rmorlok, line 15: Why isn't this inlined into IActorData?

rmorlok: Addressed in 5f6bad8. `GetAnnotations` is now inlined into `core.IActorData`; the redundant side interface is gone. I also updated the OAuth actor interface, test stubs, and generated mock so the JWT/OAuth call path still compiles with the broader actor contract.

PR #538, opened 2026-06-06
Migrate connection disconnect to workflow

### LINKED ISSUE #523: Migrate connection disconnect from Asynq task to workflow
Parent: #519

## Goal

Replace the `core:disconnect_connection` Asynq task with a go-workflows workflow while preserving existing disconnect behavior.

## Scope

- Add `core.connection.disconnect.v1` workflow.
- Move revocation/finalization side effects into versioned activities, e.g. `core.connection.disconnect.revoke_credentials.v1` and `core.connection.disconnect.finalize.v1`.
- Preserve retry behavior for credential revocation and the existing policy that local disconnect proceeds after non-context revocation failures.
- Remove the old Asynq disconnect task path in the same PR; AuthProxy is not production deployed.

## Acceptance criteria

- `DisconnectConnection` creates a workflow-backed operation and returns a tracking token.
- Existing unit and OAuth integration coverage for disconnect/revocation passes through the workflow path.
- The old `core:disconnect_connection` task is no longer registered.

rmorlok:
## Summary

- replaces the `core:disconnect_connection` Asynq handler with the versioned `core.connection.disconnect.v1` workflow
- registers disconnect workflow activities on the worker service and starts disconnect workflows from `DisconnectConnection`
- extends encrypted task tokens and `/tasks/{encryptedTaskInfo}` polling to support workflow-backed operations
- isolates go-workflows migrations from AuthProxy main DB migrations with a dedicated migration table while keeping workflow state in the primary DB
- updates the OAuth disconnect revocation integration test to drain the workflow worker path

Closes #523.

## Tests

- `go test ./internal/workflows ./internal/core ./internal/routes ./internal/tasks ./internal/service/worker ./internal/service/api ./internal/service/public`
- `go test -tags integration ./oauth2 -run DisconnectRevocation -count=1` from `integration_tests`
- `./scripts/preflight.sh`

### DIFF
--- a/internal/service/worker/server.go
+++ b/internal/service/worker/server.go
@@ -39,8 +39,12 @@ func Serve(cfg config.C) {
 	defer dm.ShutdownTelemetry()
 
 	workerConfig := cfg.GetRoot().Worker
-	router := apgin.ForService(&workerConfig, logger, cfg.IsDebugMode(),
-		apgin.WithTelemetry(dm.GetTelemetry(), dm.GetConfigRoot().Telemetry, dm.GetServiceId()))
+	router := apgin.ForService(
+		&workerConfig,
+		logger,
+		cfg.IsDebugMode(),
+		apgin.WithTelemetry(dm.GetTelemetry(), dm.GetConfigRoot().Telemetry, dm.GetServiceId()),
+	)
 
 	router.GET("/ping", func(c *gin.Context) {
 		c.PureJSON(http.StatusOK, gin.H{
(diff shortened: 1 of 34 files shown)


internal/service/worker/server.go
```diff
@@ -139,17 +143,25 @@ func Serve(cfg config.C) {
...
-		WorkflowWorkerOptions: workflowworker.WorkflowWorkerOptions{
-			MaxParallelWorkflowTasks: workerConfig.GetConcurrency(context.Background()),
-		},
-		ActivityWorkerOptions: workflowworker.ActivityWorkerOptions{
-			MaxParallelActivityTasks: workerConfig.GetConcurrency(context.Background()),
+	workflowWorker, err := apworkflows.NewWorker(
+		workflowRuntime,
+		&workflowworker.Options{
+			WorkflowWorkerOptions: workflowworker.WorkflowWorkerOptions{
+				WorkflowPollers:          2,
```
rmorlok: Please add Activity/WorkflowPollers to the config for the worker service, along with MaxParallelWorkflowTasks/MaxParallelActivityTasks and WorkflowHeartbeatInterval.

Use string values (or int equivalents) for the ones that are integers and a human duration for the later. These should be optional and default to the values defaulted in the library.

rmorlok: Addressed in 0da1421: added optional worker config fields for workflow_pollers, activity_pollers, max_parallel_workflow_tasks, max_parallel_activity_tasks, and workflow_heartbeat_interval; the worker now starts from go-workflows DefaultOptions and only overrides configured values. I also added schema/Helm support plus tests for defaults and overrides.

Issue #539, opened 2026-06-06
Add go-workflows migration table configuration upstream


rmorlok:
Parent project: #519

## Context

AuthProxy stores go-workflows state in the primary application database for both SQLite and PostgreSQL. AuthProxy also uses `golang-migrate` for its own database migrations, which creates and owns the default `schema_migrations` table.

`github.com/cschleiden/go-workflows` v1.4.2 also uses `golang-migrate` internally for its SQL backend migrations. For the Postgres and SQLite backends, its migration code uses the default migration table. When go-workflows is pointed at the same database as AuthProxy, it reads AuthProxy's migration version as if it were the go-workflows migration version. In practice this failed once AuthProxy's DB was at version 10:

```text
no migration found for version 10: read down for version 10 db/migrations: file does not exist
```

To unblock the workflow disconnect migration, AuthProxy currently embeds a copy of the go-workflows SQL migrations in `internal/workflows/migrations` and runs them with a dedicated migration table, `authproxy_workflows_schema_migrations`. That keeps workflow state in the primary DB while avoiding migration-version collisions, but it means AuthProxy must manually track upstream go-workflows SQL migrations.

## Desired upstream/fork change

Add a backend option to `github.com/cschleiden/go-workflows` that lets callers configure the `golang-migrate` migrations table used by SQL backends. The runtime workflow tables should remain unchanged; only the migration bookkeeping table should be configurable.

Suggested API shape:

```go
postgres.WithMigrationsTable("authproxy_workflows_schema_migrations")
sqlite.WithMigrationsTable("authproxy_workflows_schema_migrations")
```

or a more generic option name, if it fits the library style better:

```go
postgres.WithMigrationConfig(postgres.MigrationConfig{MigrationsTable: "..."})
sqlite.WithMigrationConfig(sqlite.MigrationConfig{MigrationsTable: "..."})
```

Implementation sketch:

- Add a `MigrationsTable string` field to the Postgres and SQLite backend options structs.
- Default it to empty so existing behavior remains unchanged.
- In Postgres `Migrate()`, pass it through to `github.com/golang-migrate/migrate/v4/database/postgres.Config{MigrationsTable: ...}`.
- In SQLite `Migrate()`, pass it through to `github.com/golang-migrate/migrate/v4/database/sqlite.Config{MigrationsTable: ...}`.
- Add backend tests proving two migrators can coexist in one database when they use different migration tables.
[...]

PR #543, opened 2026-06-06
Use forked go-workflows migration table option

rmorlok:
## Summary
- Temporarily replace github.com/cschleiden/go-workflows with rmorlok/go-workflows at v0.0.0-20260606194005-c8be9ef9bb4e
- Remove AuthProxy's copied go-workflows SQL migrations and custom golang-migrate runner
- Configure go-workflows sqlite/postgres backends with authproxy_workflows_schema_migrations via WithMigrationsTable
- Add a SQLite regression test proving an existing main schema_migrations table at version 10 no longer collides with workflow migrations

## Draft status
This is intentionally draft-only while the go-workflows migration-table option lives in the rmorlok fork. Once the upstream repo has the option, this PR should drop the replace and move back to the upstream module/version.

Refs #539.

## Verification
- go test ./internal/workflows ./internal/core ./internal/routes ./internal/tasks ./internal/service/worker ./internal/service/api ./internal/service/public
- go test -tags integration ./oauth2 -run DisconnectRevocation -count=1 (from integration_tests)
- ./scripts/preflight.sh
- go test ./...
- push hook preflight passed


rmorlok:
Blocked on https://github.com/cschleiden/go-workflows/pull/485 merging.

PR #558, opened 2026-06-08
Share Postgres pool with workflow runtime

### LINKED ISSUE #533: Consolidate Postgres database connection pools
Parent project: #519

## Context

The initial go-workflows infrastructure opens a separate Postgres `database/sql` pool for the workflow backend while the core application database code owns its own pool. This is acceptable for the first integration slice, but it means connection limits, backpressure, and pool telemetry are managed independently.

## Goal

Consolidate Postgres database ownership so go-workflows can share the primary application database connection pool, or introduce a central pool manager that applies consistent limits and observability across core database access, workflows, and adjacent storage services.

## Tasks

- Audit current database pool creation paths for core database access, workflows, app metrics, and migrations.
- Decide whether workflow code should receive a shared `*sql.DB`, a narrow database handle abstraction, or a pool manager from `DependencyManager`.
- Add configurable pool limits for Postgres if they do not already exist.
- Ensure OpenTelemetry DB stats and query instrumentation remain consistent after consolidation.
- Add regression tests or focused integration coverage for startup/shutdown ownership so shared pools are closed exactly once.

## Notes

The workflow runtime should continue to support SQLite and Postgres. Avoid coupling workflow code directly to `internal/core`; keep the dependency at the database/runtime infrastructure layer.

rmorlok:
## Summary

- move primary app SQL pool ownership into DependencyManager for both SQLite and Postgres
- construct internal/database over an already-open *sql.DB plus database dialect config
- let workflow runtime and workflow migrations borrow the DependencyManager-owned Postgres pool instead of opening their own
- add optional Postgres database pool settings and apply them when DependencyManager opens the pool
- shut down workflow runtime and the owned SQL pool from API, public, admin, and worker services

Closes #533

## Tests

- go test ./internal/schema/config ./internal/database ./internal/workflows ./internal/service/...
- ./scripts/preflight.sh

### DIFF
--- a/internal/service/dependency_manager.go
+++ b/internal/service/dependency_manager.go
@@ -2,12 +2,18 @@ package service
 
 import (
 	"context"
+	"database/sql"
 	"fmt"
 	"log/slog"
+	"math"
+	"os"
 	"sync"
 	"time"
 
 	"github.com/hibiken/asynq"
+	_ "github.com/jackc/pgx/v5/stdlib"
+	_ "github.com/mattn/go-sqlite3"
+	"github.com/mitchellh/go-homedir"
 	"github.com/rmorlok/authproxy/internal/apasynq"
 	"github.com/rmorlok/authproxy/internal/apauth/tasks"
 	authSync "github.com/rmorlok/authproxy/internal/apauth/tasks"
@@ -23,6 +29,7 @@ import (
 	"github.com/rmorlok/authproxy/internal/httpf"
 	"github.com/rmorlok/authproxy/internal/ratelimit"
 	sconfig "github.com/rmorlok/authproxy/internal/schema/config"
+	"github.com/rmorlok/authproxy/internal/sqlh"
 	"github.com/rmorlok/authproxy/internal/util/pagination"
 	"github.com/rmorlok/authproxy/internal/workflows"
 )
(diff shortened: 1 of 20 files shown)


internal/service/dependency_manager.go
```diff
@@ -228,6 +235,129 @@ func (dm *DependencyManager) GetDatabase() database.DB {
...
+	switch cfg := root.Database.InnerVal.(type) {
+	case *sconfig.DatabaseSqlite:
+		if err := ensureSqliteDatabaseFile(cfg.Path); err != nil {
+			return nil, err
+		}
+		return database.OpenInstrumentedSQL(
+			"sqlite3",
+			cfg.GetDsn(),
+			database.DBSystemSQLite,
+			database.WithTelemetry(dm.GetTelemetry(), root.Telemetry),
```
rmorlok: I think since we are moving all of the sqldb related functionality to this package, the telemetry around that should move here as well.

rmorlok: Moved the reusable database/sql telemetry helpers out of internal/database and into internal/sqlh. DependencyManager now uses sqlh.OpenInstrumentedSQL/sqlh.WithTelemetry for the primary pool, and app_metrics uses the same neutral helper package for its independently-owned SQL stores without creating a service import cycle.

PR #587, opened 2026-06-10
Define connector lifecycle API contract

### LINKED ISSUE #570: Define connector lifecycle operation API contract and permissions
Parent: #44
Label: `project:connector-lifecycle`

## Goal
Define the API, schema, SDK, and permission contract for connector-level lifecycle operations before the workflow implementations are wired to the UI.

## Proposed endpoint contract
- `POST /connectors/{id}/_disconnect_all`
- `POST /connectors/{id}/_archive`
- Both accept an optional timebox request body, likely `{ "timeout_seconds": 600 }` with validation for positive values and a documented default.
- Both return a task-polling response using the existing encrypted task token pattern, likely `{ "task_id": "...", "connector_id": "..." }`.
- Polling remains `GET /tasks/{task_id}`, which already supports workflow-backed tasks.

## Permissions
Add separate connector verbs so operators can grant read, disconnect, and archive independently:
- `connectors:disconnect_all`
- `connectors:archive`

The routes should still validate the loaded connector/connector-version namespace/resource id through the existing resource permission validator.

## Scope
- Add request/response structs under `internal/schema/api` and matching JSON schema fixtures/tests.
- Add OpenAPI doc wrapper structs where needed.
- Add JS SDK methods and exported types in `sdks/js/src/connectors.ts`.
- Add route-level tests for auth/permission failures and request validation; routes can initially rely on mocked core methods once the interface is extended.
- Regenerate Swagger through preflight.

## Acceptance Criteria
- API schema validates the new request/response types.
- Swagger includes both endpoints and their permission-protected behavior.
- SDK exposes typed helpers for both operations.
- Tests cover missing/invalid connector id, invalid timebox, permission denial, and successful task response shape.

rmorlok:
Closes #570.

## Summary

- adds connector lifecycle contract endpoints for `POST /connectors/{id}/_disconnect_all` and `POST /connectors/{id}/_archive`
- introduces dedicated permissions via the route resource checks: `connectors:disconnect_all` and `connectors:archive`
- adds the shared `{ "timeout_seconds": 600 }` request payload, default 10 minute timeout handling, and `{ "task_id", "connector_id" }` task response contract
- wires the contract through OpenAPI/schema generation and the JS SDK connector helpers

The core methods intentionally return `501 Not Implemented` until the workflow execution work lands in the follow-up issues. Route tests use a fake lifecycle core so this PR validates the API/permission/task-token contract without starting placeholder workflow instances.

## Tests

- `go test ./internal/httperr ./internal/schema/api ./internal/core ./internal/routes ./internal/service/...`
- `go test ./...`
- `yarn workspace @authproxy/api typecheck`
- `./scripts/preflight.sh`

### DIFF
--- a/internal/routes/connectors.go
+++ b/internal/routes/connectors.go
@@ -150,6 +150,22 @@ func parseConnectorID(gctx *gin.Context) (apid.ID, *httperr.Error) {
 	return id, nil
 }
 
+func parseConnectorVersionID(gctx *gin.Context) (connectorVersionID, *httperr.Error) {
+	id, herr := parseConnectorID(gctx)
+	if herr != nil {
+		return connectorVersionID{}, herr
+	}
+	versionStr := gctx.Param("version")
+	if versionStr == "" {
+		return connectorVersionID{}, httperr.BadRequest("version is required")
+	}
+	version, err := strconv.ParseUint(versionStr, 10, 64)
+	if err != nil {
+		return connectorVersionID{}, httperr.BadRequest("failed to parse version as an integer")
+	}
+	return connectorVersionID{ConnectorID: id, Version: version}, nil
+}
+
 func connectorVersionStatesToAPI(states database.ConnectorVersionStates) schemaapi.ConnectorVersionStates {
 	if states == nil {
 		return nil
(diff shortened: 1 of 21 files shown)


internal/routes/connectors.go
```diff
@@ -119,12 +126,30 @@ type ConnectorsRoutes struct {
...
+	encrypt              encrypt.E
 	labelsAdapter        key_value.Adapter[apid.ID]
 	annotsAdapter        key_value.Adapter[apid.ID]
 	versionLabelsAdapter key_value.Adapter[connectorVersionID]
 	versionAnnotsAdapter key_value.Adapter[connectorVersionID]
 }
 
+const defaultConnectorLifecycleTimeout = 10 * time.Minute
+
+func parseConnectorID(gctx *gin.Context) (apid.ID, *httperr.Error) {
```
rmorlok, line 138: if we are going to introduce this new helper method, apply it to the other routes in this file so they share the same logic

rmorlok: Also add a test directly for this

internal/routes/connectors.go
```diff
@@ -1196,6 +1221,194 @@ func (r *ConnectorsRoutes) forceVersionState(gctx *gin.Context) {
 	gctx.PureJSON(http.StatusOK, ConnectorVersionToJson(cv))
 }
 
+func (r *ConnectorsRoutes) loadConnectorByID(ctx context.Context, connectorId apid.ID) (connIface.Connector, error) {
```
rmorlok, line 1093: same statement here, if we are adding this method, add tests for it and use it consistently across the code in this file.

internal/routes/connectors.go
rmorlok: Addressed in cf23959. `parseConnectorID` is now used by the existing connector routes as well as the lifecycle routes, and I added direct tests for the parser success and failure cases.

internal/routes/connectors.go
rmorlok: Addressed in cf23959. The connector loader is now shared by the existing get/create-version/lifecycle paths and the key-value adapters use it via the route helper; I also added direct tests for found and not-found loader behavior.

PR #599, opened 2026-06-13
Fix archive workflow activity logging

rmorlok:
## Summary
- inject a logger into the archive workflow activity tests that call service methods requiring `service.logger`
- keep the production archive workflow activity behavior unchanged

## Main-branch failure addressed
- `Golang Codebase` failed on `main` in run `27473805850` because `TestArchiveConnectorVersionActivitiesTransitionStates` built `&service{db: db}` and then called an activity that uses `s.logger.With(...)`.
- The separate `Build Image` failure timed out while pulling BuildKit from Docker Hub, which prevented the demo-shell image tag and caused the later manual demo deploy to fail image-tag verification. No repo code had executed in that matrix leg.

## Validation
- `env AUTH_PROXY_TEST_DATABASE_PROVIDER=sqlite go test ./internal/core -run TestArchiveConnectorVersionActivities -count=1`
- `env AUTH_PROXY_TEST_DATABASE_PROVIDER=sqlite go test ./internal/core -count=1`
- `./scripts/preflight.sh`

### DIFF
--- a/internal/core/workflow_archive_connector_test.go
+++ b/internal/core/workflow_archive_connector_test.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"context"
+	"log/slog"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


rmorlok:
Closing as superseded by latest main: 73ec67d (#597) already injects test_utils.NewTestLogger() into the archive workflow activity tests, keeping production behavior unchanged.

PR #650, opened 2026-06-16
Add connector predicate schema support

### LINKED ISSUE #646: Schema support for conditional OAuth scopes and probes
Part of #645.

Implement the connector schema surface for conditional OAuth scopes and probes.

Tasks:
- Add a reusable predicate schema shape for { javascript: string }.
- Add if to OAuth scope schema.
- Change OAuth scope required to accept the existing boolean form or a predicate object.
- Add if to probe schema.
- Validate non-empty JavaScript consistently with setup-step predicates.
- Update committed JSON schema artifacts and schema fixtures.

Acceptance:
- Connector YAML accepts scope if.javascript.
- Connector YAML accepts probe if.javascript.
- Scope required supports both required: false and required.javascript.
- Blank JavaScript is rejected.
- Focused connector schema/unit tests pass.

rmorlok:
## Summary
- add a shared connector Predicate schema/type for JavaScript boolean conditions
- add OAuth scope if.javascript and dynamic required.javascript while preserving required: true/false
- add probe if.javascript schema support and validation fixtures

## Verification
- go test ./internal/schema/resources/connectors
- go test ./internal/schema/config
- go test ./internal/auth_methods/oauth2
- go test ./internal/schema/resources/connectors ./internal/schema/config ./internal/auth_methods/oauth2
- go test ./internal/...
- ./scripts/preflight.sh

Closes #646

### DIFF
--- a/internal/schema/common/predicate.go
+++ b/internal/schema/common/predicate.go
@@ -1,0 +1,30 @@
+package common
+
+import (
+	"strings"
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/rmorlok/authproxy/internal/apjs"
+)
+
+// Predicate defines a JavaScript condition evaluated by runtime with a
+// caller-defined set of variables in scope. Runtime requires the fragment to
+// return a boolean.
+type Predicate struct {
+	Javascript string `json:"javascript" yaml:"javascript"`
+}
+
+func (p *Predicate) Validate(vc *ValidationContext, vars map[string]any) error {
+	if p == nil {
+		return nil
+	}
+	result := &multierror.Error{}
+	if strings.TrimSpace(p.Javascript) == "" {
+		result = multierror.Append(result, vc.NewErrorfForField("javascript", "javascript is required"))
+		return result.ErrorOrNil()
+	}
+	if _, err := apjs.EvaluateBoolean(p.Javascript, vars); err != nil {
+		result = multierror.Append(result, vc.NewErrorfForField("javascript", "javascript must evaluate to a boolean: %v", err))
+	}
+	return result.ErrorOrNil()
+}
--- a/internal/schema/common/predicate_test.go
+++ b/internal/schema/common/predicate_test.go
@@ -1,0 +1,41 @@
+package common
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestPredicateValidate(t *testing.T) {
+	vars := map[string]any{
+		"cfg":         map[string]any{},
+		"labels":      map[string]string{},
+		"annotations": map[string]string{},
+	}
+
+	t.Run("accepts boolean result", func(t *testing.T) {
+		p := &Predicate{Javascript: `cfg.enabled === true`}
+		require.NoError(t, p.Validate(&ValidationContext{Path: "predicate"}, vars))
+	})
+
+	t.Run("rejects blank javascript", func(t *testing.T) {
+		p := &Predicate{Javascript: " \n\t "}
+		err := p.Validate(&ValidationContext{Path: "predicate"}, vars)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "predicate.javascript")
+	})
+
+	t.Run("rejects syntax errors", func(t *testing.T) {
+		p := &Predicate{Javascript: `cfg.enabled ===`}
+		err := p.Validate(&ValidationContext{Path: "predicate"}, vars)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "must evaluate to a boolean")
+	})
+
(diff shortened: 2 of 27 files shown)


internal/schema/resources/connectors/predicate.go
```diff
@@ -0,0 +1,30 @@
...
+	"strings"
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/rmorlok/authproxy/internal/schema/common"
+)
+
+// Predicate defines a connector-authored JavaScript condition evaluated by
+// runtime with cfg, labels, and annotations variables in scope. Runtime
+// requires the fragment to return a boolean.
+type Predicate struct {
```
rmorlok: Can we please move predict to `internal/schema/common` to be potentially shared in other places? This would obviously include the json schema pieces.

internal/schema/resources/connectors/predicate.go
```diff
@@ -0,0 +1,30 @@
...
+	Javascript string `json:"javascript" yaml:"javascript"`
+}
+
+func (p *Predicate) Validate(vc *common.ValidationContext) error {
```
rmorlok: Please extend validation to make sure the javascript is valid. The validate method would take map of objects that will be provided to the js and the validate should invoke the js to make sure it returns a boolean value. If it errors or returns an invalid type, that's an error.

rmorlok (commented):
See comments inline

internal/schema/resources/connectors/predicate.go
rmorlok: Moved the predicate type into internal/schema/common, moved the JSON Schema definition to common/schema.json, and changed connector schemas to reference the shared common Predicate definition. The connector package now only keeps aliases for compatibility/domain naming.

internal/schema/resources/connectors/predicate.go
rmorlok: Extended Predicate.Validate to take the provided JS variable map and execute the fragment through the shared boolean evaluator. Validation now rejects syntax/runtime errors and non-boolean/null/undefined results; connector schema validation passes cfg, labels, and annotations placeholders, with common-level tests covering the behavior.

internal/schema/common/predicate.go
```diff
@@ -0,0 +1,59 @@
...
+		result = multierror.Append(result, vc.NewErrorfForField("javascript", "javascript must evaluate to a boolean: %v", err))
+	}
+	return result.ErrorOrNil()
+}
+
+// EvaluateBooleanJavascript runs a JavaScript expression with the supplied
+// variables in scope and returns its boolean result. Undefined, null, and
+// other result types are rejected so callers can fail closed.
+func EvaluateBooleanJavascript(expression string, vars map[string]any) (bool, error) {
+	vm := goja.New()
```
rmorlok: Please move the implementation of this logic back into internal/apjs. The predicate can simple call that function and use the error output for the validate step. The logic does not need to be in the schema package.

internal/schema/common/predicate_test.go
```diff
@@ -0,0 +1,41 @@
+package common
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestPredicateValidate(t *testing.T) {
```
rmorlok, line 9: Tests can be duplicated here as they are testing the behavior of predicate.

internal/schema/resources/connectors/predicate.go
```diff
@@ -0,0 +1,17 @@
+package connectors
+
+import "github.com/rmorlok/authproxy/internal/schema/common"
+
+type Predicate = common.Predicate
```
rmorlok: Do not alias this here. Just use the predicate from common.

internal/schema/common/predicate.go
rmorlok: Addressed in 69c987d: the JavaScript boolean evaluator is back in internal/apjs as apjs.EvaluateBoolean, and common.Predicate.Validate now delegates to it and surfaces the evaluator error in schema validation.

internal/schema/common/predicate_test.go
rmorlok: Kept this coverage in internal/schema/common/predicate_test.go. Those tests still exercise Predicate.Validate behavior directly, even though execution now delegates to internal/apjs.

internal/schema/resources/connectors/predicate.go
rmorlok: Addressed in 69c987d: removed the connector-local Predicate and SetupFlowStepIf aliases. Setup-flow steps, OAuth scopes/required predicates, probes, and tests now use common.Predicate directly.

PR #651, opened 2026-06-16
Add AWS KMS support under unified key model

### LINKED ISSUE #617: Add AWS KMS support under the unified key model
Parent: #605

## Scope

Add AWS KMS support under the unified key/DEK model.

## Acceptance criteria

- AWS KMS provider supports wrapping, unwrapping, provider identity/version metadata, and native DEK generation where appropriate.
- Config/schema/API can express AWS KMS key reference, region, credentials, optional endpoint override, and cache behavior.
- Unit tests cover provider mapping, generation, wrap/unwrap, errors, and context cancellation.
- Integration tests exercise namespace key creation, DEK generation, encrypt/decrypt, provider rotation or metadata advancement, DEK rewrap sync, re-encryption, and decrypt after restart.
- Integration tests are gated with opt-in AWS env vars and documented run commands.

rmorlok:
## Summary
- add an aws_kms KeyData provider that uses AWS KMS GenerateDataKey, Encrypt, Decrypt, and DescribeKey for DEK lifecycle
- persist provider metadata on newly generated DEKs so sync can compare KMS wrapping state without immediate redundant rewraps
- add AWS KMS config schema/docs plus an opt-in integration test path with optional metadata-advancement rewrap coverage

## Tests
- go test ./internal/schema/config ./internal/encrypt -count=1
- go test -tags "integration,aws" ./encrypt -run TestAwsKMSKeySyncAndReencrypt -count=1 (from integration_tests; skips without AWS env)
- ./scripts/preflight.sh

Closes #617

### DIFF
--- a/internal/schema/config/key_version_info.go
+++ b/internal/schema/config/key_version_info.go
@@ -18,6 +18,7 @@ const (
 	ProviderTypeFile           ProviderType = "file"
 	ProviderTypeRandom         ProviderType = "random"
 	ProviderTypeAws            ProviderType = "aws"
+	ProviderTypeAwsKMS         ProviderType = "aws_kms"
 	ProviderTypeGcp            ProviderType = "gcp"
 	ProviderTypeHashicorpVault ProviderType = "hashicorpvault"
 	ProviderTypeRaw            ProviderType = "raw"
--- a/internal/schema/config/key_version_info.go
+++ b/internal/schema/config/key_version_info.go
@@ -11,17 +11,17 @@ import (
 type ProviderType string
 
 const (
-	ProviderTypeValue          ProviderType = "value"
-	ProviderTypeBase64         ProviderType = "base64"
-	ProviderTypeEnvVar         ProviderType = "env_var"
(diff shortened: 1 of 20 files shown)


internal/schema/config/key_version_info.go
```diff
@@ -18,6 +18,7 @@ const (
 	ProviderTypeFile           ProviderType = "file"
 	ProviderTypeRandom         ProviderType = "random"
 	ProviderTypeAws            ProviderType = "aws"
+	ProviderTypeAwsKMS         ProviderType = "aws_kms"
```
rmorlok: Can we rename the `aws` one above to something that directly implies that it is aws secret manager? My hunch is the kms option will become the default, and I don't want it to see like the default aws option in secret manager.

rmorlok: Renamed the existing AWS Secrets Manager provider from `ProviderTypeAws = "aws"` to `ProviderTypeAwsSecretsManager = "aws_secrets_manager"` in 17b8a46, and updated the Secrets Manager/KMS tests to use the clearer provider identity.

PR #676, opened 2026-06-20
Add connector-level JavaScript schema and runtime context

### LINKED ISSUE #670: connectors schema: accept and validate connector-level javascript
Parent: #667

Add top-level javascript to connector definitions and schemas. Validate connector-level JavaScript at definition validation time, then pass the compiled/validated library into setup-step, probe, and OAuth scope predicate validation while preserving standalone validator compatibility.

### LINKED ISSUE #671: core: evaluate data-source transforms through connector JavaScript context
Parent: #667

Route setup-flow data-source transforms through the connector JavaScript context so transforms can call connector-level helpers and can receive data, cfg, labels, and annotations. Preserve existing transform shape validation and error handling.

### LINKED ISSUE #672: core: evaluate predicates through connector JavaScript context
Parent: #667

Add cached connector-version JavaScript library access, replace GetPredicateVars with a JavaScript context, and migrate setup-step eligibility, probe enablement, and OAuth effective-scope evaluation to use connector-level helpers.

rmorlok:
Closes #670
Closes #671
Closes #672

## Summary
- Add top-level connector `javascript` field and JSON schema support.
- Validate connector JavaScript at definition validation time and pass the compiled library into connector predicate validation.
- Replace connection predicate vars with `GetJavascriptContext`, backed by a cached connector-version JS library.
- Evaluate setup-step, probe, OAuth scope, and data-source transform JavaScript through the shared context with `cfg`, `labels`, `annotations`, and `data` where applicable.

## Tests
- `go test ./internal/schema/common ./internal/schema/resources/connectors ./internal/core ./internal/auth_methods/oauth2`
- `./scripts/preflight.sh`

### DIFF
--- a/internal/schema/common/predicate.go
+++ b/internal/schema/common/predicate.go
@@ -15,6 +15,10 @@ type Predicate struct {
 }
 
 func (p *Predicate) Validate(vc *ValidationContext, vars map[string]any) error {
+	return p.ValidateWithContext(vc, apjs.NewContext(nil, vars))
+}
+
+func (p *Predicate) ValidateWithContext(vc *ValidationContext, jsctx apjs.Context) error {
 	if p == nil {
 		return nil
 	}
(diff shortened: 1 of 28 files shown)


internal/schema/common/predicate.go
```diff
@@ -15,6 +15,10 @@ type Predicate struct {
 }
 
 func (p *Predicate) Validate(vc *ValidationContext, vars map[string]any) error {
```
rmorlok: Wouldn't validate always need context passed in? How are we validating that the functions being called from the javascript exist if the actually exists in the surrounding javascript? Wouldn't that error on validation?

e.g.

```
javascript: |
  function foo() { return true; }
if:
  javascript: foo()
```

rmorlok: Updated in 92e9cbf: `common.Predicate.Validate` now requires an `apjs.Context` directly, and the connector validators pass the compiled connector JS library through that context. I also added a common predicate test for the `function foo() { ... }` style helper case so validation proves connector-level helpers are visible.

PR #701, opened 2026-06-30
Move key schema code to key resource package

rmorlok:
## Summary
- move key-related schema code from `internal/schema/config` into `internal/schema/resources/key`
- add key resource JSON schema, schema ID, README, and schema tests
- move shared AWS credential schema/types into `internal/schema/common` so key resources do not depend on config
- keep config compatibility re-exports and update config schema refs to compose the new key resource schema

## Testing
- `go test ./internal/schema/common ./internal/schema/resources/key ./internal/schema/config -run 'TestSchema|TestSchemaId|Test_Schema'`
- `go test ./... -run '^$'`
- `./scripts/preflight.sh`


internal/schema/resources/key/key_data_serialization_yaml.go
```diff
@@ -42,6 +43,8 @@ func (kd *KeyData) UnmarshalYAML(value *yaml.Node) error {
 		keyData = &KeyDataFile{}
 	case keys["random"] || keys["num_bytes"]:
 		keyData = &KeyDataRandomBytes{}
+	case keys["num_bytes"]:
```
rmorlok: Isn't this duplicative of the case above?

rmorlok: Removed the duplicate num_bytes branch in 4130203c; the preceding random || num_bytes case already covers it.

Issue #708, opened 2026-07-03
Migration analysis: hooks, config, OAuth, scopes, probes, rollback


rmorlok:
Parent: #703

Implement the behavior inside the connector version migration workflow.

Scope:
- Run stepwise up/down hooks across intermediate versions.
- Apply setup-field defaults and detect missing required preconnect/configure fields.
- Detect OAuth auth input changes and required/optional scope changes.
- Trigger target-version OAuth refresh and classify permanent vs transient failures.
- Run newly introduced target probes when migration is otherwise unblocked.
- Resolve migration-owned notifications during rollback/downgrade when conditions no longer apply.

Acceptance:
- Hook failure leaves the connection unchanged.
- Required reauth remains configured + unhealthy with actionable notification.
- Required configure sets setup_step with actionable notification.
- Downgrade can restore readiness through successful target-version refresh.

rmorlok:
Implemented on main; closing stale project issue. Evidence: candidate construction runs ordered up/down hooks and targets all target-version probes in internal/core/workflow_migrate_connection_version_candidate.go; analysis handles setup defaults, missing configure/preconnect fields, OAuth changes, and required-action notification priority in internal/core/workflow_migrate_connection_version_analysis.go; the activity handles refresh, probe execution, atomic update, and notification resolution in internal/core/workflow_migrate_connection_version_activity.go; focused tests cover these components.

Issue #707, opened 2026-07-03
Notification persistence, service, cache, and API


rmorlok:
Parent: #703

Implement centralized actor-visible notifications used by connector version migrations and future resource workflows.

Scope:
- Add notifications and notification_views tables for SQLite and Postgres.
- Add database/service methods for deterministic upsert, resolve, actor viewed state, and actor-filtered listing.
- Store view permissions and action permissions separately, then compute can_action per request.
- Add GET /notifications and POST /notifications/{id}/_viewed.
- Add Redis-backed actor notification caching with safe invalidation.

Acceptance:
- Actor listing filters by stored permissions.
- Mark-viewed is actor-specific.
- Resolved notifications stop appearing in active lists.
- Cache invalidates on write/resolve/viewed changes.

Issue #706, opened 2026-07-03
Marketplace, API client, and documentation polish for migration notifications


rmorlok:
Parent: #703

Expose the migration and notification experience cleanly to marketplace users and host applications.

Scope:
- Update generated API client/types after the backend API changes.
- Add notification UI surfaces or minimum actionable notification handling in marketplace/admin flows.
- Ensure configured connections with pending setup_step can resume required migration configuration.
- Add docs for connector migration hooks, notification behavior, rollback, and host-orchestrated bulk rollout.
- Include required UI screenshots in the PR description if UI code changes.

Acceptance:
- Users can see and mark relevant notifications viewed.
- Action URLs route actors to reauth/configure flows when can_action is true.
- Docs include up/down ordering and downgrade behavior.

Issue #705, opened 2026-07-03
Connection migration endpoint, workflow, and DB update primitive


rmorlok:
Parent: #703

Add the single-connection migration API and durable workflow shell for connector version changes.

Scope:
- Add POST /connections/{id}/_migrate_version with connections:update permission.
- Add request/response API DTOs and OpenAPI aliases.
- Register workflow core.connection.migrate_version.v1 with per-connection instance ids.
- Replace the existing Asynq shell task path with the workflow-backed implementation.
- Add an atomic database primitive for switching connection connector_version while updating encrypted config, labels, annotations, setup step/error, and health state.

Acceptance:
- Reject no-op, cross-connector, draft/archived target, and concurrent migrations.
- Response returns a task token bound to the actor.
- DB update recomputes connector-version and namespace carry-forward labels.

rmorlok:
Implemented on main; closing stale project issue. Evidence: POST /connections/{id}/_migrate_version is registered in internal/routes/connections.go; request and response DTOs are in internal/schema/api/connections.go; durable workflow core.connection.migrate_version.v1 and per-connection instance IDs are in internal/core/workflow_migrate_connection_version.go; service startup is in internal/core/service_connection_migrate_version.go; the atomic database update primitive lives in internal/database/connection.go; focused workflow, route, and database coverage is present.

PR #709, opened 2026-07-03
Add connector version migration workflow

rmorlok:
## Summary
- Add connector-authored migration hook schema and JS evaluation for up/down version moves.
- Add POST /connections/{id}/_migrate_version backed by a durable workflow and atomic connection-version update path.
- Add notification persistence/API with viewed/action state plus migration analysis for setup-field diffs, OAuth refresh, probes, rollback-owned notification cleanup, and generated schemas/Swagger.

## Related Issues
- Refs #703
- Refs #704
- Refs #705
- Refs #707
- Advances #708
- Leaves #706 for marketplace/API-client UI polish follow-up.

## Tests
- go test ./...
- env AUTH_PROXY_TEST_DATABASE_PROVIDER=sqlite go test ./internal/database
- ./scripts/preflight.sh

## Screenshots
- Not applicable: no UI files changed.


internal/apjs/context.go
```diff
@@ -68,6 +68,26 @@ func (c Context) TransformJSON(expression string, data any) ([]DataSourceOption,
 	return c.WithVar("data", data).TransformOptions(expression)
 }
 
+// EvaluateObject runs a JavaScript expression and exports its result as a
+// JSON-like object. Undefined/null are treated as an empty object so migration
+// hooks can no-op by returning nothing.
```
rmorlok: > so migration hooks can no-op by returning nothing.

Please don't reference migration here. Only talk about what this package's functionality is, as it should be general.

internal/apjs/library.go
```diff
@@ -53,6 +53,20 @@ func CompileAndValidateLibrary(source string) (*Library, error) {
 	return library, nil
 }
 
+// ValidateExpressionSyntax parses a connector-authored expression without
+// executing it. This is used for migration hooks, which often depend on
```
rmorlok: Same thing here. Don't mention migrations, but you can give general scenarios where this would be useful.

internal/core/workflow_migrate_connection_version.go
```diff
@@ -0,0 +1,728 @@
...
+}
+
+type migrationHookPatch struct {
+	Config        migrationAnyPatch          `json:"config"`
+	Labels        migrationStringPatch       `json:"labels"`
+	Annotations   migrationStringPatch       `json:"annotations"`
+	Notifications []migrationNotificationDef `json:"notifications"`
+}
+
+type migrationAnyPatch struct {
```
rmorlok: Please add comments on these data structures for what they are used for.

internal/database/connection.go
```diff
@@ -532,6 +532,143 @@ func (s *service) SetConnectionEncryptedConfiguration(ctx context.Context, id ap
 	return nil
 }
 
+type ConnectionVersionMigrationUpdate struct {
```
rmorlok, line 539: Add a description comment for what this data represents

internal/core/workflow_migrate_connection_version.go
```diff
@@ -0,0 +1,728 @@
+package core
```
rmorlok, line 1: This file is too long to digest easily. Please split it into multiple files that are constrained to a purpose, e.g.

workflow_migrate_connection_version_datatypes_datatypes.go
workflow_migrate_connection_version_datatypes_activity_apply.go
workflow_migrate_connection_version_datatypes_helpers_build_candidate.go

You don't have to use my exact names, but it will make it easier for me to navigate the code.

internal/apjs/context.go
rmorlok: Addressed in 822eff6: generalized the EvaluateObject comment so internal/apjs describes package behavior without referencing migration hooks.

internal/apjs/library.go
rmorlok: Addressed in 822eff6: generalized the ValidateExpressionSyntax comment to describe runtime-variable expressions without naming migration hooks.

internal/database/connection.go
rmorlok: Addressed in 822eff6: added a description comment for ConnectionVersionMigrationUpdate explaining that it is the atomic connection-row replacement applied during version migration.

internal/core/workflow_migrate_connection_version.go
rmorlok: Addressed in 822eff6: added purpose comments for the migration patch, map patch, notification definition, and candidate state structures in workflow_migrate_connection_version_types.go.

internal/core/workflow_migrate_connection_version.go
rmorlok: Addressed in 822eff6: split the large workflow file into focused files for workflow registration, activity application, candidate building, migration analysis, hook patch handling, and data types.

internal/routes/notifications.go
```diff
@@ -0,0 +1,233 @@
...
+// @Param			limit			query		int		false	"Maximum number of notifications to return"
+// @Param			include_viewed	query		bool	false	"Include notifications the actor has already viewed"
+// @Param			state			query		string	false	"Notification state; defaults to active"
+// @Success		200				{object}	OpenAPIListNotificationsResponseJson
+// @Failure		400				{object}	ErrorResponse
+// @Failure		401				{object}	ErrorResponse
+// @Failure		500				{object}	ErrorResponse
+// @Security		BearerAuth
+// @Router			/notifications [get]
+func (r *NotificationsRoutes) list(gctx *gin.Context) {
```
rmorlok, line 53: Notifications should be filterable to a namespace and to a label selector. The label selector means we will need to denormalize labels from the resource in question into the notification.

rmorlok: Addressed in 516efb7: notifications now denormalize resource labels, ListNotifications supports namespace matchers and label_selector filters, GET /notifications exposes those query params, and migration notifications populate labels from the updated connection row. Added SQLite/Postgres-covered DB tests and regenerated Swagger.

internal/database/notification.go
```diff
@@ -0,0 +1,583 @@
...
+			*m = nil
+			return nil
+		}
+		return json.Unmarshal(v, m)
+	default:
+		return fmt.Errorf("cannot convert %T to NotificationMetadata", value)
+	}
+}
+
+type Notification struct {
```
rmorlok, line 142: Please add explanations on the meanings of the fields. Where possible, give examples of what values would be. Especially for fields like key, which are not obvious what that is.

internal/core/workflow_migrate_connection_version_analysis.go
```diff
@@ -0,0 +1,228 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/rmorlok/authproxy/internal/database"
+	aschema "github.com/rmorlok/authproxy/internal/schema/auth"
+	cschema "github.com/rmorlok/authproxy/internal/schema/resources/connectors"
+)
+
+func (s *service) applyProbeMigrationAnalysis(candidate *connectionMigrationCandidate) {
```
rmorlok: I don't understand why this method is needed. Actors only should get notifications for things that require their action (reauth/reconfigure for new features/etc). Actors are likely users of the host application and generally wouldn't have knowledge about probes. They wouldn't need to be notified that probes are going to run.

internal/database/notification.go
rmorlok: Addressed in 5e99a3a: added field-level comments to Notification with concrete meanings and examples, including how Key is used as the deterministic dedupe/upsert key.

internal/core/workflow_migrate_connection_version_analysis.go
rmorlok: Addressed in 5e99a3a: removed the user-facing notification for newly introduced probes. The workflow still queues target-only probes to run after migration when appropriate, but actors only get notifications for actionable or user-relevant conditions.

PR #721, opened 2026-07-11
Fix demo request detail blob handling

rmorlok:
## Summary
- configure the demo app-metrics full-log store to use the mounted filesystem blob path instead of the implicit in-memory fallback
- return and log full-log store failures so missing blobs are visible instead of silently treated as successful writes
- fall back to summary-only request detail data when a full-log blob is missing, avoiding 500s in the admin UI

## Verification
- env AUTH_PROXY_TEST_DATABASE_PROVIDER=sqlite go test ./internal/app_metrics -count=1
- env AUTH_PROXY_TEST_DATABASE_PROVIDER=sqlite go test ./internal/routes -count=1
- kubectl kustomize deploy/kustomize/authproxy-demo/overlays/demo
- ./scripts/preflight.sh


deploy/kustomize/authproxy-demo/overlays/demo/authproxy-config.yaml
```diff
@@ -40,6 +40,9 @@ app_metrics:
   database:
     provider: sqlite
     path: /tmp/authproxy-app-metrics.db
+  blob_storage:
+    provider: filesystem
+    path: /tmp/authproxy-blobs
```
rmorlok: In th demo environment, we are supposed to be using minio to simulate s3

internal/app_metrics/storage_service.go
```diff
@@ -113,7 +114,16 @@ func (ss *StorageService) GetFullLog(ctx context.Context, id apid.ID) (*FullLog,
 		return nil, err
 	}
 
-	return ss.fullStore.GetFullLog(ctx, log.Namespace, id)
+	fullLog, err := ss.fullStore.GetFullLog(ctx, log.Namespace, id)
+	if err != nil {
+		if errors.Is(err, apblob.ErrBlobNotFound) {
+			ss.logger.Warn("full HTTP log blob missing; returning summary-only request event", "record_id", id.String(), "namespace", log.Namespace)
+			return NewFullLogFromRecord(log), nil
```
rmorlok: Instead of faking this out here, if the full log should return not found from the endpoint and the UI should handle that case, falling back to the data it has from the log record.

rmorlok: Changed the missing-blob path so StorageService returns not found, the endpoint surfaces that as 404, and the admin request drawer falls back to the list-row metadata when the detail fetch 404s.

PR #745, opened 2026-07-22
Add API key version migration scenarios

### LINKED ISSUE #739: API key connector version migration integration scenarios
Parent: #737

## Summary
Add API-key end-to-end integration scenarios for connector-version migration.

## Work
- Configure-change scenario: v1 API-key connection reaches configured/healthy, v2 adds a required configure field, migration sets setup_step and a single setup_required notification, user submits configure data, setup clears, notification resolves, and connection remains healthy.
- Preconnect-change scenario: v1 API-key connection reaches configured/healthy, v2 adds a required preconnect field, migration marks configured connection unhealthy with one auth_required notification, user reauths through preconnect plus API-key credential form, verify succeeds, health is healthy, and notification resolves.
- Add one descriptive Markdown scenario file per Go test.

## Acceptance
- Scenarios use public APIs for connection creation, version publishing, migration, setup submission, and reauth.
- DB reads are limited to final state/config/credential assertions.
- Notifications are high-level only: no per-field keys or assertions.

rmorlok:
## Summary
- add API-key connector version migration integration scenarios for required configure and preconnect changes
- document each scenario with fixture, flow, and assertions markdown
- resolve high-level setup/auth required notifications when setup completes, auth credentials are established without verify, or verify succeeds

Closes #739

## Tests
- go test ./internal/core -run 'Test.*Notification|Test.*Verify|Test.*Setup|Test.*MigrateConnectionVersion|TestSubmitForm|TestApiKeySubmit|TestHandleCredentialsEstablished|TestNotificationKeysToResolve'
- go test ./internal/core -run 'TestGetCurrentSetupStepResponse|TestCancelSetup|TestReconfigure|TestReauthConnection|TestRetryConnectionSetup'
- cd integration_tests && go test -tags integration -run '^$' ./version_migration/... ./api_key/...
- ./scripts/preflight.sh

## Not run
- cd integration_tests && go test -tags integration -v ./version_migration/... (blocked: Docker daemon is not running, so the compose-backed Postgres/Redis test services could not be started)
- go test ./internal/core (blocked: local Postgres test dependency is not running at 127.0.0.1:5432)

### DIFF
--- a/integration_tests/version_migration/README.md
+++ b/integration_tests/version_migration/README.md
@@ -8,8 +8,8 @@ state, auth/config behavior, probes, and notifications.
 
 Scenario files:
 
-- API-key configure migration: tracked by #739.
-- API-key preconnect migration: tracked by #739.
+- API-key configure migration: `api_key_configure_change_test.md`.
+- API-key preconnect migration: `api_key_preconnect_change_test.md`.
 - OAuth required-scope reauth migration: tracked by #740.
 - OAuth required-scope rollback migration: tracked by #740.
 - No-auth defaults and probe migration: tracked by #741.
--- a/integration_tests/version_migration/README.md
+++ b/integration_tests/version_migration/README.md
@@ -8,8 +8,8 @@ state, auth/config behavior, probes, and notifications.
 
 Scenario files:
 
-- API-key configure migration: `api_key_configure_change_test.md`.
-- API-key preconnect migration: `api_key_preconnect_change_test.md`.
+- API-key configure migration: [api_key_configure_change_test.md](api_key_configure_change_test.md).
+- API-key preconnect migration: [api_key_preconnect_change_test.md](api_key_preconnect_change_test.md).
 - OAuth required-scope reauth migration: tracked by #740.
 - OAuth required-scope rollback migration: tracked by #740.
 - No-auth defaults and probe migration: tracked by #741.
--- a/internal/core/connection_setup_dispatch.go
+++ b/internal/core/connection_setup_dispatch.go
@@ -82,6 +82,9 @@ func (c *connection) completeFlow(ctx context.Context) (iface.ConnectionSetupRes
 			return nil, httperr.InternalServerError(httperr.WithInternalErrorf("failed to set connection state to configured: %w", err))
 		}
 	}
+	if err := c.resolveRequiredActionNotifications(ctx, database.NotificationKeySetupRequired); err != nil {
+		return nil, httperr.InternalServerError(httperr.WithInternalErrorf("failed to resolve setup-required notification: %w", err))
+	}
 	return &iface.ConnectionSetupComplete{
 		Id:   c.GetId(),
 		Type: iface.ConnectionSetupResponseTypeComplete,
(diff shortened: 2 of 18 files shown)


internal/core/connection_setup_dispatch.go
```diff
@@ -194,3 +197,29 @@ func ensureStepMatches(setupStep *cschema.SetupStep, reqStepId string) error {
...
+	}
+	return nil
+}
+
+func (c *connection) isAuthMethodSetupStep(stepID string) bool {
+	connector := c.cv.GetDefinition()
+	if connector == nil || connector.SetupFlow == nil {
+		return true
+	}
+	_, found := connector.SetupFlow.FindStepById(stepID)
```
rmorlok, line 234: Does `SetupFlow` contain both preconnect and configure steps? Am I understanding this correctly that if it's neither or these type of steps, it is an auth method step?

internal/core/connection_setup_dispatch.go
```diff
@@ -194,3 +197,29 @@ func ensureStepMatches(setupStep *cschema.SetupStep, reqStepId string) error {
...
+	if next != nil && next.Type() == iface.ManifestStepTypeVerify {
+		return nil
+	}
+	if err := c.resolveRequiredActionNotifications(ctx, database.NotificationKeyAuthRequired); err != nil {
+		return httperr.InternalServerError(httperr.WithInternalErrorf("failed to resolve auth-required notification: %w", err))
+	}
+	return nil
+}
+
+func (c *connection) isAuthMethodSetupStep(stepID string) bool {
```
rmorlok, line 229: Please add a comment describing what this method does and what constitutes an auth method step.

internal/core/connection_setup_dispatch.go
```diff
@@ -194,3 +197,29 @@ func ensureStepMatches(setupStep *cschema.SetupStep, reqStepId string) error {
 	}
 	return nil
 }
+
+func (c *connection) resolveAuthRequiredIfCredentialsEstablished(
```
rmorlok, line 207: Please add a comment on this method describing the scenarios in which it will apply changes. If this is specific to API key connections, that's not obvious to me from looking at it.

integration_tests/version_migration/README.md
```diff
@@ -8,8 +8,8 @@ state, auth/config behavior, probes, and notifications.
 
 Scenario files:
 
-- API-key configure migration: tracked by #739.
-- API-key preconnect migration: tracked by #739.
+- API-key configure migration: `api_key_configure_change_test.md`.
```
rmorlok: Can we make these relative links so they will be clickable in the markdown view in github?

internal/core/connection_setup_dispatch.go
rmorlok: Fixed in 009e31de. Added a method comment describing when auth_required is resolved: after a credential-establishing auth-method step is submitted, unless the flow is handing off to verify. It is not API-key-only, though the API-key form path is the scenario this PR exercises.

internal/core/connection_setup_dispatch.go
rmorlok: Fixed in 009e31de. Added a comment explaining what counts as an auth-method step and why those steps are identified by exclusion from the connector YAML setup_flow.

internal/core/connection_setup_dispatch.go
rmorlok: Fixed in 009e31de. Yes, that understanding is right: connector SetupFlow contains only the user-authored preconnect and configure steps; auth-method credential steps are injected into the runtime manifest flow between those phases. I added comments spelling that out where the helper makes this distinction.

PR #751, opened 2026-07-23
Move load-test commands into dedicated CLI

### LINKED ISSUE #749: Load test seed command should not be part of the core server CLI
See `cmd/server/main.go`, right now we have specific cli flags for load testing and seeding. That doesn't feel like it should be in the core CLI. Or at least it should be gated with a build flag so that only certain builds contain that functionality.

rmorlok:
## Summary
- move the seeder and background scenario runner out of the production server CLI
- add the standalone authproxy-loadtest executable with seed and background commands
- update load-test scripts and document LOADTEST_CLI_BIN

Closes #749

## Verification
- go test ./cmd/server ./cmd/loadtest
- go run ./cmd/server --help
- go run ./cmd/loadtest --help
- ./loadtest/scripts/seed smoke --verify-samples 1 --progress-every 100
- ./loadtest/scripts/background smoke scheduler-sync
- ./scripts/preflight.sh


cmd/server/main_test.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+func TestRootCommandExcludesLoadTestCommands(t *testing.T) {
+	rootCmd := newRootCommand()
+	commands := make(map[string]struct{})
+	for _, command := range rootCmd.Commands() {
+		commands[command.Name()] = struct{}{}
+	}
+
+	for _, commandName := range []string{"loadtest-seed", "loadtest-background", "seed", "background"} {
```
rmorlok: Don't add a test for this. Please remove.

rmorlok: Removed both command-surface tests and their test-only root-command factories in 06b9358f.

PR #777, opened 2026-07-27
Resource names: add shared validation and unversioned database storage

### LINKED ISSUE #771: Resource names: add shared validation and unversioned database storage
Parent: #770

## Goal

Establish the shared name contract and persistent storage for non-versioned
resources so later API and label work has one validated, provider-consistent
foundation.

## Scope

- Define a reusable resource-name type/validator in the appropriate schema
  package.
  - non-empty when stored or returned
  - compatible with implicit label values
  - accepts generated AuthProxy IDs, including `_`
  - has an explicit length limit and normalization/case policy
- Add `name` storage for actors, connections, keys, and rate limits in both
  SQLite and PostgreSQL.
- Backfill every existing row with its immutable ID and make new stored rows
  non-empty.
- Enforce uniqueness among live rows with a provider-equivalent constraint on
  `(namespace, name)` for each resource table.
- Make soft-deleted rows stop reserving the name.
- Treat namespace names as derived from the final path segment; do not add a
  second mutable namespace identity or implement path rename.
- Update database entities, `cols`/`fields`/`values`, interfaces, mocks, and
  seed/bootstrap paths, including the global key.

Connector storage is deliberately excluded; a connector is reconstructed from
version rows and needs a separate logical-resource change.

## Acceptance criteria

- Fresh and upgraded databases expose valid names for every affected row.
- Duplicate live names in one namespace/type are rejected under both providers.
- The same name is allowed in different namespaces, for different resource
  types, and after the prior owner is soft-deleted.
- Migration and database tests pass for SQLite and PostgreSQL.
- `go generate ./internal/database/...` is run when interfaces change.

rmorlok:
## Summary

- add a shared, case-sensitive resource name primitive compatible with implicit label values
- add SQLite and PostgreSQL storage, ID backfills, non-empty enforcement, and live namespace/name uniqueness for actors, connections, keys, and rate limits
- default omitted names to immutable IDs and update entity/fixture plumbing, including the global key
- cover rename conflicts, namespace/type isolation, soft-delete reuse, validation, and migration upgrades

## Testing

- `AUTH_PROXY_TEST_DATABASE_PROVIDER=sqlite go test ./internal/... -count=1`
- `AUTH_PROXY_TEST_DATABASE_PROVIDER=postgres go test ./internal/database/... -count=1`
- `./scripts/preflight.sh`

Closes #771

### DIFF
--- a/internal/database/actor.go
+++ b/internal/database/actor.go
@@ -16,6 +16,7 @@ import (
 	"github.com/rmorlok/authproxy/internal/apid"
 	"github.com/rmorlok/authproxy/internal/encfield"
 	aschema "github.com/rmorlok/authproxy/internal/schema/auth"
+	scommon "github.com/rmorlok/authproxy/internal/schema/common"
 	"github.com/rmorlok/authproxy/internal/schema/resources/namespace"
 	"github.com/rmorlok/authproxy/internal/util"
 	"github.com/rmorlok/authproxy/internal/util/pagination"
@@ -92,6 +93,7 @@ const ActorTable = "actors"
 // Actor is some entity taking action within the system.
 type Actor struct {
 	Id           apid.ID
+	Name         scommon.ResourceName
 	Namespace    string
 	ExternalId   string
 	Permissions  Permissions
@@ -133,6 +135,7 @@ func (a *Actor) cols() []string {
 		"updated_at",
 		"encrypted_at",
 		"deleted_at",
+		"name",
 	}
 }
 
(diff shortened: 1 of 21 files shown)


internal/database/actor.go
```diff
@@ -133,6 +135,7 @@ func (a *Actor) cols() []string {
 		"updated_at",
 		"encrypted_at",
 		"deleted_at",
+		"name",
```
rmorlok: Instead of as the last attribute in this list, can we position this as "id", "namespace", "name" for all resources?

rmorlok: Addressed in c5bce861. I reordered the struct fields and the matching cols/fields/values sequences to id, namespace, name for actors, connections, keys, and rate limits. The complete SQLite and PostgreSQL database suites and repository preflight pass with the new ordering.

PR #778, opened 2026-07-28
Resource names: persist one logical connector name across versions

### LINKED ISSUE #772: Resource names: persist one logical connector name across versions
Parent: #770

Depends on #771.

## Goal

Give each logical connector one persistent, mutable name shared by every
connector version.

## Scope

- Introduce connector-level storage in which the stable connector ID, namespace,
  and name have one source of truth independent of version rows.
- Backfill one connector record for every distinct live connector ID and use
  the connector ID as the initial name.
- Enforce live `(namespace, name)` uniqueness for logical connectors under
  SQLite and PostgreSQL.
- Wire connector creation, config/bootstrap loading, version creation, version
  lookup/list projections, archival, and soft deletion to the connector-level
  record.
- Ensure creating, editing, publishing, archiving, or migrating a version
  cannot fork or silently overwrite the logical connector name.
- Preserve the existing stable connector ID and version-history behavior.

## Acceptance criteria

- Every version projection for one connector reports the same name.
- Renaming a connector changes the logical resource without creating a new
  connector version or rewriting immutable version definitions.
- A conflicting connector name in the same namespace is rejected; the same
  name in another namespace or resource type remains valid.
- Existing databases and configured connectors backfill deterministically.
- Migration and connector lifecycle tests pass under SQLite and PostgreSQL.

rmorlok:
Closes #772

## Summary

- make `connectors` the single source of truth for connector ID, namespace, mutable name, labels, annotations, and lifecycle timestamps
- replace the physical `connector_versions` table with `connector_definition_versions`, containing a unique `cvd_` row ID, `connector_id`, logical version number, state, encrypted definition, and encryption timestamp
- align database models and files with the normalized schema:
  - `Connector` represents exactly one `connectors` row
  - `ConnectorDefinitionVersion` represents exactly one `connector_definition_versions` row
  - `ConnectorWithDefinition` is the joined projection used by definition-aware reads
- rename database methods, filters, state types, mocks, tests, and label-propagation tasks to use the connector-definition terminology
- enforce unique `(connector_id, version)` pairs and remove duplicated connector metadata, the legacy `type` column, and the persisted definition hash from version storage
- directly and deterministically migrate existing SQLite and PostgreSQL databases, with a rollback migration for local development
- project connector-owned metadata through every version read and update resource search, re-encryption, purge, config migration, metrics fixtures, and load-test seeding for the normalized schema
- remove connector `states` and `versions` summary fields from core interfaces, REST responses, Swagger, and the JS SDK; the admin detail view derives both from its version-list query, while metrics explicitly count definition versions
- propagate connector label changes by logical connector ID across every referencing connection
- retain definition history on connector soft deletion and remove it through the connector foreign-key cascade on hard purge

`encrypted_at` remains on definition rows because key rotation updates it together with the encrypted definition.

## UI verification

The connector detail view still presents deduplicated available states and the total number of versions, now derived from the version list rather than aggregate connector fields.

Admin connector detail with states and version count derived from definition versions

## Verification

- `env AUTH_PROXY_TEST_DATABASE_PROVIDER=sqlite GOCACHE=/tmp/authproxy3-go-cache go test ./internal/database/...`
- `env AUTH_PROXY_TEST_DATABASE_PROVIDER=sqlite go test ./internal/core ./internal/routes`
- PostgreSQL tests for `./internal/database/...`, `./internal/core`, `./internal/routes`, `./internal/app_metrics`, and `./internal/encrypt`
- `GOCACHE=/tmp/authproxy3-go-cache go test ./... -run '^$'`
- `yarn workspace @authproxy/api test --run`
- `yarn workspace @authproxy/admin typecheck`
- `yarn workspace @authproxy/admin test --run`
- `yarn workspace @authproxy/marketplace typecheck`
- `yarn workspace @authproxy/marketplace test --run`
- `./scripts/preflight.sh`


internal/app_metrics/task_resource_snapshot.go
```diff
@@ -195,7 +200,7 @@ func (h *ResourceSnapshotTaskHandler) snapshotConnectors(ctx context.Context, sa
 					Labels:            connector.Labels,
 					State:             connector.State,
 					ConnectorVersion:  connector.Version,
-					TotalVersions:     connector.TotalVersions,
+					TotalVersions:     totalVersions,
```
rmorlok: Let's remove total versions here to simplify the query for this job. Remove from the schema for the snapshot as well.

internal/core/connector.go
```diff
@@ -3,23 +3,12 @@ package core
...
 	"github.com/rmorlok/authproxy/internal/aplog"
 	"github.com/rmorlok/authproxy/internal/core/iface"
-	"github.com/rmorlok/authproxy/internal/database"
 )
 
-// Connector object is returned from queries for connectors, with one record per id. It aggregates some information
-// across all versions for a connector.
+// Connector represents a logical connector together with the definition
+// version selected for display.
 type Connector struct {
```
rmorlok, line 23: It doesn't seem like connector is adding anything on top of ConnectorVersion and that ConnectorVersion is really a ConnectorWithVersion (because it has the labels, etc).

For the core layer, let's simplify. ConnectorVersion becomes connector and this empty shell goes away.

internal/app_metrics/task_resource_snapshot.go
rmorlok: Addressed in 9be6e381. Removed `total_versions` from connector resource samples across the model, SQLite/PostgreSQL/ClickHouse schemas, persistence, and tests. The snapshot job no longer performs the per-connector definition-version count query.

internal/core/connector.go
rmorlok: Addressed in 9be6e381. The hydrated core model is now `Connector`: the empty wrapper and `ConnectorVersion` core interface/type are gone, related core files/builders/tests/mocks were renamed, and service/route signatures now consistently return `iface.Connector` while retaining version-specific operation names.

internal/core/iface/connection.go
```diff
@@ -36,7 +36,7 @@ type Connection interface {
 	 * Nested entities
 	 */
 
-	GetConnectorVersionEntity() ConnectorVersion
```
rmorlok, line 39: Let's rename this to `GetConnector()`

internal/core/connector_state.go
```diff
@@ -0,0 +1,17 @@
+package core
+
+import (
+	"context"
+
+	"github.com/rmorlok/authproxy/internal/database"
+)
+
+func (cv *Connector) SetState(ctx context.Context, state database.ConnectorDefinitionVersionState) error {
```
rmorlok: Instead of `cv` let's use `c` for all `Connector` instance methods.

internal/core/probe.go
```diff
@@ -80,7 +80,7 @@ func (p *probeBase) recordInvokeOutcome(
 	return outcome, err
 }
 
-func NewProbe(cfg *cschema.Probe, s *service, cv *ConnectorVersion, c *connection) iface.Probe {
+func NewProbe(cfg *cschema.Probe, s *service, cv *Connector, c *connection) iface.Probe {
```
rmorlok: Same note here, let's look for variables that are `Connector` and named `cv` and rename them `c`.

internal/core/service_connection_create.go
```diff
@@ -16,7 +16,7 @@ import (
 func (s *service) CreateConnection(
 	ctx context.Context,
 	namespace string,
-	cv iface.ConnectorVersion,
+	cv iface.Connector,
```
rmorlok: Same comment here about `cv` -> `c`

internal/core/iface/connection.go
rmorlok: Addressed in 3ffaa521. Renamed the nested connector accessor to `GetConnector()` across `iface.Connection`, the concrete and mock implementations, routes, OAuth consumers, and the HTTP factory adapter.

internal/core/connector_state.go
rmorlok: Addressed in 3ffaa521. All `Connector` instance methods now use `c` receivers. I also retained the multiline `SetState` formatting from the local worktree while updating the receiver and references.

internal/core/probe.go
rmorlok: Addressed in 3ffaa521. Hydrated `Connector` values no longer use `cv` names across core, routes, OAuth, or affected tests. Probe state now uses a `connector` field and `NewProbe` receives the connector as `c`.

internal/core/service_connection_create.go
rmorlok: Addressed in 3ffaa521. `CreateConnection` now accepts the connector as `c`, and the same naming cleanup was applied consistently to other hydrated `Connector` variables.

internal/database/reencrypt_registry.go
```diff
@@ -30,6 +30,10 @@ type EncryptedFieldRegistration struct {
 	JoinLocalCol     string // e.g. "connection_id" — FK column on this table
 	JoinRemoteCol    string // e.g. "id" — PK column on join table
 	JoinNamespaceCol string // e.g. "namespace" — namespace column on join table
+
+	// OmitBaseSoftDeleteFilter is used when the encrypted child table has no
+	// deleted_at column and lifecycle is owned by its joined parent.
+	OmitBaseSoftDeleteFilter bool
```
rmorlok: Instead of adding this filter, let's add a `deleted_at` column to the connector version definitions table and propagate soft deletes from the parent to the child.

rmorlok: Addressed in 97b06b40. Added `deleted_at` to `connector_definition_versions` in both SQLite and PostgreSQL (including migration backfill from the logical connector), and `DeleteConnector` now soft-deletes the connector and all definition-version rows with the same timestamp in one transaction. The re-encryption registry is back to its standard base-table soft-delete filter, and regression tests cover migration backfill, propagation, unrelated connectors, and exclusion from re-encryption.

internal/database/migrations/postgres/000016_create_connectors.up.sql
```diff
@@ -0,0 +1,108 @@
...
+where row_num = 1;
+
+create table connector_definition_versions
+(
+    id                   text primary key,
+    connector_id         text not null references connectors (id) on delete cascade,
+    version              bigint not null,
+    state                text not null,
+    encrypted_definition jsonb not null,
+    encrypted_at         timestamptz,
```
rmorlok, line 79: This table should also have created_at and updated_at the same as other tables. Please do that an make sure those fields are properly wired in the database methods.

rmorlok: Addressed in 0849fe11. Added non-null `created_at` and `updated_at` columns to `connector_definition_versions` for SQLite and PostgreSQL, preserving the original per-version timestamps during both migration directions. `ConnectorDefinitionVersion` and `ConnectorWithDefinition` now carry the per-definition timestamps separately from the logical connector timestamps; inserts, draft updates, state transitions, and soft deletes maintain them, and definition-version timestamp ordering now uses the definition rows. Tests cover migration values, create/update timestamps, ordering, and both providers.

PR #781, opened 2026-08-01
Add resource name API behavior

### LINKED ISSUE #773: Resource names: add create, rename, and exact-name API behavior
Parent: #770

Depends on #771 and #772.

## Goal

Expose resource names consistently through core interfaces and the Admin API,
including optional creation, rename-by-ID, and exact-name collection queries.

## Scope

- Add `name` to response projections for namespaces, actors, connectors,
  connector versions, connections, keys, and rate limits.
- Add optional `name` to applicable create requests. When omitted, assign the
  generated immutable ID after ID generation.
- Add optional `name` to update requests addressed by immutable ID.
- Treat connector name as connector-level metadata:
  - connector create/update may set it
  - connector version create/update cannot define a divergent name
  - every version response exposes the logical connector name
- Expose a namespace's final path segment as its read-only derived name.
  Namespace path/name rename remains out of scope.
- Add exact `name` filtering to database/core list builders and collection API
  query DTOs for every in-scope resource.
- Apply existing permission-constrained namespace matchers before returning
  name matches; a broad namespace query may legitimately return the same name
  from multiple namespaces.
- Map uniqueness violations to a stable HTTP 409 response and avoid
  provider-specific error leakage.
- Update centralized API schemas, OpenAPI adapters, generated mocks, and route
  tests.

Name-based item routes such as `/connections/by-name/...` are out of scope;
immutable IDs remain the only direct-addressing contract.

## Acceptance criteria

- Create-with-name, create-with-default, rename, conflict, get, list-by-name,
  pagination, and authorization behavior is covered for every resource type.
- Connector version APIs cannot create version-specific names.
- Namespace responses expose the correct path segment without introducing
  namespace rename behavior.
- Existing clients that omit `name` continue to work.

rmorlok:
## Summary

- expose names when creating, retrieving, updating, and listing actors, connections, connectors, keys, rate limits, and namespaces
- keep one logical connector name across versions and derive namespace names from the final path segment
- add exact name filtering with cursor pagination and namespace authorization constraints
- normalize SQLite and PostgreSQL uniqueness conflicts to stable HTTP 409 responses
- regenerate the public API documentation and add provider-level and route-level coverage

## Testing

- full Go test suite with SQLite
- database resource-name tests with PostgreSQL
- focused resource-name route tests with SQLite and PostgreSQL
- ./scripts/preflight.sh

Closes #773

### DIFF
--- a/internal/schema/resources/namespace/README.md
+++ b/internal/schema/resources/namespace/README.md
@@ -6,6 +6,7 @@ Use this package for namespace resource semantics. Auth permissions reference na
 
 API request and response DTOs for namespace routes live in `internal/schema/api`. Keep this package focused on reusable namespace primitives: paths, matchers, hierarchy helpers, and the JSON Schema definitions other packages can reference.
 
-A namespace's read-only resource name is the final segment of its path. For
-example, `root.prod.billing` has the name `billing`; renaming still requires a
-path change and is not a namespace API operation.
+A namespace's read-only resource name is set automatically to the final segment
+of its path when the namespace is created. For example, `root.prod.billing` has
+the name `billing`. Namespace paths are immutable, so namespace names cannot be
+changed after creation.
--- a/internal/schema/resources/namespace/README.md
+++ b/internal/schema/resources/namespace/README.md
@@ -5,3 +5,7 @@ This package owns namespace path and matcher validation plus constants for names
 Use this package for namespace resource semantics. Auth permissions reference namespace matchers, but namespace validation itself is not auth-specific.
 
 API request and response DTOs for namespace routes live in `internal/schema/api`. Keep this package focused on reusable namespace primitives: paths, matchers, hierarchy helpers, and the JSON Schema definitions other packages can reference.
+
+A namespace's read-only resource name is the final segment of its path. For
+example, `root.prod.billing` has the name `billing`; renaming still requires a
+path change and is not a namespace API operation.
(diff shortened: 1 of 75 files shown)


internal/schema/resources/namespace/README.md
```diff
@@ -5,3 +5,7 @@ This package owns namespace path and matcher validation plus constants for names
 Use this package for namespace resource semantics. Auth permissions reference namespace matchers, but namespace validation itself is not auth-specific.
 
 API request and response DTOs for namespace routes live in `internal/schema/api`. Keep this package focused on reusable namespace primitives: paths, matchers, hierarchy helpers, and the JSON Schema definitions other packages can reference.
+
+A namespace's read-only resource name is the final segment of its path. For
+example, `root.prod.billing` has the name `billing`; renaming still requires a
+path change and is not a namespace API operation.
```
rmorlok: Let's change this detail. You cannot rename a namespace because you cannot update the namespace path. The name must be the last value in the namespace and it is set automatically when the namespace is created.

Please update any implementation logic accordingly.

rmorlok: Addressed in 0880e3cc. Namespace documentation and generated OpenAPI now state that the name is automatically derived from the final path segment at creation and cannot be changed because the path is immutable. The update contract remains limited to labels and annotations, and route assertions now verify the derived name on both create and update responses.

PR #813, opened 2026-08-05
Serialize migrations before concurrent server startup

rmorlok:
## Summary
- complete auto-migrations once before `cmd/server` starts multiple service goroutines
- prevent fresh databases from being observed in golang-migrate's dirty state during all-service startup

## Validation
- `go test ./cmd/server`
- `./scripts/preflight.sh`


rmorlok:
Closing in favor of more explicit ticketed approach.

PR #820, opened 2026-08-12
Harden automatic migration locks

### LINKED ISSUE #816: Harden Redis-backed automatic migration locks
## Problem

Automatic startup migration currently relies on several Redis locks, but the locking contract is incomplete and one timeout helper is incorrect. This makes automatic migration unsuitable as the only coordinator for a production deployment and leaves a real concurrency gap for application-metrics stores.

During the disposable demo reset on 2026-08-05, the fresh PostgreSQL database was left at `schema_migrations` version `9` with `dirty=true`. The later container restart reported only `Dirty database version 9`; the original migration error was lost in the crash loop. This state proves an earlier attempt began and did not complete, but does **not** alone prove that two migrators owned the Redis lock at once.

## Current behavior and evidence

- `AutoMigrateDatabase` uses Redis key `db-migrate-lock` around the main relational and workflow migrations: [`dependency_manager.go`](https://github.com/rmorlok/authproxy/blob/main/internal/service/dependency_manager.go#L366-L396).
- PostgreSQL `golang-migrate` additionally uses a PostgreSQL advisory lock for its migrations. Redis is therefore an application-level guard, not the sole serializer.
- `AutoMigrateAppMetricsService` directly migrates configured SQL and ClickHouse stores with **no distributed lock**: [`dependency_manager.go`](https://github.com/rmorlok/authproxy/blob/main/internal/service/dependency_manager.go#L480-L488).
- Encryption-key, connector, and configured-actor reconciliation use their own Redis locks, while `AutoMigrateAll` invokes each operation in sequence: [`dependency_manager.go`](https://github.com/rmorlok/authproxy/blob/main/internal/service/dependency_manager.go#L788-L795).
- The primary migration lock has a fixed lease (default two minutes) and migration callers do not renew it. A migration exceeding the lease can allow another process to acquire the key before the first has finished.
- `MutexOptionRetryFor` has inverted deadline handling. With `context.Background()`—the normal migration caller—the `!ok` branch compares against the zero deadline and returns the original context, so the requested retry deadline is never applied: [`mutex.go`](https://github.com/rmorlok/authproxy/blob/main/internal/apredis/mutex.go#L97-L113).

## Suggested approach

1. Correct `MutexOptionRetryFor` so it creates a timeout unless the supplied context already has an earlier deadline. Add focused tests for no deadline, an earlier deadline, and a later deadline.
2. Define the automatic-migration locking contract explicitly:
   - retain backend-native schema locks where available (PostgreSQL advisory lock);
   - do not treat an expiring Redis lease as the only schema-migration guarantee;
   - if automatic migration retains a Redis lock, either renew its lease for the full operation or fail safely before a lease can expire;
   - surface lock acquisition, lease loss, and release failures in structured logs/metrics.
3. Add a distinct distributed lock for application-metrics migration, covering both configured stores and ClickHouse where present. Do not reuse `db-migrate-lock` for unrelated storage backends.
4. Audit the remaining automatic reconciliation operations (DEK generation/sync, connector loading, actor sync) and document which are protected, which need locking, and which are idempotent by design.
5. Add integration coverage with fresh PostgreSQL + Redis and multiple concurrent `AutoMigrateAll` callers. It should verify:
   - one migration owner per target;
   - a well-defined bounded wait on contention;
   - no dirty migration state after successful concurrent startup;
   - an actionable original error when a migration fails.

## Acceptance criteria

- The timeout helper has correct, tested deadline semantics.
- Every automatic migration target has a deliberate exclusivity strategy.
- App-metrics SQL and ClickHouse migration is protected from concurrent startup invocations.
- Locks cannot silently expire while an automatic migration is still running, or automatic migration fails before that can happen.
- A parallel-start integration test prevents regression.

## Relationship to production migration work

This issue hardens the development/automatic path. Production deployments should instead use the explicit migration command and Kubernetes Job tracked separately, so application workloads never race to migrate at startup.

rmorlok:
## Summary

- extend the shared `apredis.Mutex` support with a renewable `RunWithMutex` helper, cancellation on lease loss, reliable cleanup, and correct bounded-retry deadlines
- use the same Redis coordination contract for primary SQLite/PostgreSQL migrations, workflow migrations, and SQLite/PostgreSQL/ClickHouse app-metrics migrations
- move the other startup/task reconcilers onto the renewable helper and make golang-migrate respond to context cancellation at safe boundaries
- document the lock targets, lease behavior, provider coverage, and production rollout guidance

## Locking interface

Callers continue to construct an `apredis.Mutex` with the existing options, then run protected work through:

```go
err := apredis.RunWithMutex(ctx, mutex, leaseDuration, func(lockCtx context.Context) error {
    return migrate(lockCtx)
})
```

The helper acquires the lock, refreshes it every third of the lease, cancels `lockCtx` and returns `ErrMutexLeaseLost` if renewal fails, and always attempts release on success, error, or panic. This stays provider-independent because Redis is the common coordinator; PostgreSQL's migrate-driver advisory lock remains as defense in depth.

The primary and workflow schemas share `db-migrate-lock`. App metrics uses `app-metrics-migrate-lock`, allowing its separately configured database to migrate independently.

## Verification

- `go test -count=1 -timeout=20m ./...`
- SQLite affected-package suites
- PostgreSQL affected-package suites
- ClickHouse `internal/app_metrics` suite
- `yarn docs:build`
- `./scripts/preflight.sh`

Closes #816

### DIFF
--- a/internal/util/migrate.go
+++ b/internal/util/migrate.go
@@ -1,0 +1,28 @@
+package util
+
+import (
+	"context"
+
+	"github.com/golang-migrate/migrate/v4"
+)
+
+// RunMigrationsUp runs all pending migrations and asks golang-migrate to stop
+// at its next safe boundary if ctx is cancelled. golang-migrate does not accept
+// a context directly, so callers must bridge cancellation through GracefulStop.
+func RunMigrationsUp(ctx context.Context, migrator *migrate.Migrate) error {
+	done := make(chan struct{})
+	defer close(done)
+
+	go func() {
+		select {
+		case <-ctx.Done():
+			select {
+			case migrator.GracefulStop <- true:
+			case <-done:
+			}
+		case <-done:
+		}
+	}()
+
+	return migrator.Up()
+}
--- a/internal/workflows/runtime.go
+++ b/internal/workflows/runtime.go
@@ -126,7 +126,6 @@ type MigrateOption func(*migrateOptions)
 
 type migrateOptions struct {
 	postgresDB *sql.DB
-	ctx        context.Context
 }
 
 func WithPostgresMigrationDB(db *sql.DB) MigrateOption {
@@ -135,13 +134,7 @@ func WithPostgresMigrationDB(db *sql.DB) MigrateOption {
 	}
 }
 
-func WithMigrationContext(ctx context.Context) MigrateOption {
-	return func(o *migrateOptions) {
-		o.ctx = ctx
-	}
-}
-
-func Migrate(root *sconfig.Root, logger *slog.Logger, opts ...MigrateOption) error {
+func Migrate(ctx context.Context, root *sconfig.Root, logger *slog.Logger, opts ...MigrateOption) error {
 	if root == nil || root.Database == nil {
 		return fmt.Errorf("database configuration is required")
 	}
@@ -149,9 +142,6 @@ func Migrate(root *sconfig.Root, logger *slog.Logger, opts ...MigrateOption) err
 	migrateOpts := &migrateOptions{}
 	for _, opt := range opts {
 		opt(migrateOpts)
-	}
-	if migrateOpts.ctx == nil {
-		migrateOpts.ctx = context.Background()
 	}
 
 	switch cfg := root.Database.InnerVal.(type) {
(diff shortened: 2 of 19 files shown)


internal/util/migrate.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"context"
+
+	"github.com/golang-migrate/migrate/v4"
+)
+
+// RunMigrationsUp runs all pending migrations and asks golang-migrate to stop
+// at its next safe boundary if ctx is cancelled. golang-migrate does not accept
+// a context directly, so callers must bridge cancellation through GracefulStop.
+func RunMigrationsUp(ctx context.Context, migrator *migrate.Migrate) error {
```
rmorlok, line 12: Please add a test for this logic.

rmorlok: Added in e2b9b11d. The test uses the migrate package stubs to pause an in-flight migration, cancels the context after the next migration is discovered, verifies cancellation is bridged to GracefulStop, and confirms the current migration completes while the next one does not run. I also ran it 20 times plus the affected package suites and repository preflight.

internal/workflows/runtime.go
```diff
@@ -133,6 +135,12 @@ func WithPostgresMigrationDB(db *sql.DB) MigrateOption {
 	}
 }
 
+func WithMigrationContext(ctx context.Context) MigrateOption {
```
rmorlok: Please don't do this option-style for passing context. Modify the actual migrate function to take ctx as a required first parameter.

rmorlok: Addressed in 814ab4ac. `workflows.Migrate` now takes `context.Context` as its required first parameter; `WithMigrationContext` and its fallback were removed, and all service, unit-test, and integration-test call sites were updated. While validating this, I also fixed the failing CI renewal test by waiting for the fake-clock timer to be re-armed before advancing it again.

PR #822, opened 2026-08-15
Add explicit database migration workflow

### LINKED ISSUE #817: Make automatic startup migration a development-only CLI opt-in
## Goal

Move automatic migration out of persisted configuration and make it an explicit, development-only startup option. A process that is not explicitly allowed to migrate must still verify that every configured schema is at the expected version before it starts serving.

This separates two currently conflated concerns:

- **development convenience**: create/upgrade a local disposable environment automatically;
- **production safety**: refuse to serve when the deployed schema is not compatible with the binary.

Production schema changes are tracked by the explicit migration command and Kubernetes Job work.

## Current behavior

`autoMigrate` is currently persisted independently in the main database, connector, and app-metrics configuration. It is enabled in development, demo, and Helm defaults. [`DatabasePostgres`](https://github.com/rmorlok/authproxy/blob/main/internal/schema/config/database_postgres.go#L15-L48) and [`AppMetrics`](https://github.com/rmorlok/authproxy/blob/main/internal/schema/config/app_metrics.go#L15-L25) expose it as configuration.

Each service invokes `AutoMigrateAll()` when it starts:

- [`admin-api`](https://github.com/rmorlok/authproxy/blob/main/internal/service/admin_api/server.go#L290)
- [`api`](https://github.com/rmorlok/authproxy/blob/main/internal/service/api/server.go#L187)
- [`public`](https://github.com/rmorlok/authproxy/blob/main/internal/service/public/server.go#L282)
- [`worker`](https://github.com/rmorlok/authproxy/blob/main/internal/service/worker/server.go#L110)

When `authproxy serve ... all` runs those services in one process, the entire automatic migration chain is therefore attempted multiple times. [`AutoMigrateAll`](https://github.com/rmorlok/authproxy/blob/main/internal/service/dependency_manager.go#L788-L795) includes database/workflow schema, encryption bootstrap, app-metrics schema, connector reconciliation, and configured actor synchronization.

### CLI contract

Add a `--auto-migrate` flag to `authproxy serve`.

- It is intentionally absent from YAML/JSON configuration and Helm/Kustomize AuthProxy configuration.
- When specified, it logs a conspicuous warning before any work begins, for example: `WARNING: --auto-migrate is intended only for local/disposable development environments. Use 'authproxy migrate' before production deployment.`
- It must be documented as unsafe for production workloads.
- The flag is opt-in; its absence is the production-safe default.

### Centralized per-process migration

- Remove `AutoMigrateAll()` calls from the individual service startup paths.
- In `authproxy serve`, when `--auto-migrate` is enabled, create one migration coordinator and run the complete automatic-development migration/reconciliation plan exactly once before starting any requested services.
- Preserve correct behavior when a developer starts one service in one process; the migration plan still runs once for that process.
- Cross-process automatic migration must continue to use the hardened locks tracked in #816; per-process centralization is not a replacement for cross-process coordination.

### Mandatory compatibility verification

Whether or not `--auto-migrate` is set, startup must verify the version/status of every configured schema before opening listeners or workers:

- main relational database;
- workflow database schema;
- application-metrics SQL store(s), including ClickHouse when configured.

If any configured target is behind, ahead, dirty, unavailable, or otherwise incompatible, startup must fail with an actionable error describing the target and expected/current version. It must not start partially.

Verification must be read-only. It must not create migration tables, mutate schema state, seed encryption keys, synchronize connectors, or enqueue actor-sync tasks.

## Implementation scope

- Remove the `autoMigrate` fields, accessors, JSON Schema properties, validators, test fixtures, dev/demo configuration, Helm values/templates, and deployment documentation.
- Introduce a reusable migration status/compatibility API so `serve` and the explicit `migrate` command report the same target names and version semantics.
- Keep development reconciliation behavior (DEK setup/sync, configured connectors, actor sync) behind the explicit `--auto-migrate` flag; do not perform it as part of read-only production startup validation.
- Ensure all startup failures return/propagate cleanly rather than panicking from service goroutines.
- Update local development instructions to show `authproxy serve --auto-migrate ...` when a disposable database needs initialization.

## Acceptance criteria

- A configuration containing `autoMigrate` is rejected as an unknown field.
- `authproxy serve` without `--auto-migrate` performs only read-only compatibility checks and fails before serving if a target is not current.
- `authproxy serve --auto-migrate` logs the production warning and invokes the automatic plan once per process, before service fan-out.
- Starting `admin-api,api,public,worker` in one process does not multiply migration/reconciliation attempts.
- All configured relational, workflow, application-metrics SQL, and ClickHouse schema targets are validated.
- Unit/integration coverage proves both current-schema startup and rejected stale/dirty schema startup.

### LINKED ISSUE #818: Add an explicit migration command and Kubernetes migration Job
## Goal

Production-style deployments must perform migrations as an explicit, observable release step—not as a side effect of starting API, public, admin, or worker workloads. Add a first-class migration command and run it from a dedicated Kubernetes Job before deploying AuthProxy services.

This is the production counterpart to #817 (development-only `--auto-migrate`) and should use the hardened automatic-path locking contract from #816 where a backend-specific exclusivity mechanism is required.

## Why

There is no general migration command today. The server binary exposes `serve`, `routes`, and `reencrypt`; the only `--migrate` behavior is load-test seeding and covers only a subset of the required work. [`cmd/loadtest/seed.go`](https://github.com/rmorlok/authproxy/blob/main/cmd/loadtest/seed.go#L49-L64)

The current demo/Helm deployment configuration enables automatic migration in the application process. For example, the demo configuration enables main DB, connector, and app-metrics automatic migration: [`authproxy-config.yaml`](https://github.com/rmorlok/authproxy/blob/main/deploy/kustomize/authproxy-demo/overlays/demo/authproxy-config.yaml#L17-L42). A deployment rollout currently applies the application and data-store manifests before waiting for the AuthProxy deployment: [`deploy-demo.yml`](https://github.com/rmorlok/authproxy/blob/main/.github/workflows/deploy-demo.yml#L305-L382).

That makes migration ownership hard to observe, retry, diagnose, and sequence. A failed startup can obscure the original error behind a subsequent dirty-schema error.

## Proposed CLI

Add an explicit server-binary command:

```sh
authproxy migrate --config=/etc/authproxy/config.yaml
authproxy migrate status --config=/etc/authproxy/config.yaml
```

The command must not depend on or consult an `autoMigrate` configuration setting. It is the intentional operator/deploy-time action for production and disposable environments alike.

### `migrate`

- Validates and resolves the normal configuration/secrets.
- Runs the complete **schema** migration plan once, in a documented dependency order:
  1. primary relational database;
  2. workflow schema;
  3. application-metrics SQL store(s), including ClickHouse where configured.
- Performs necessary encryption-key bootstrap/synchronization required for a freshly initialized database to start safely.
- Does not start HTTP servers, workers, schedulers, or background jobs.
- Returns a non-zero error with target name and underlying cause; do not panic.
- Is idempotent when all targets are current.
- Does not automatically force or clear a dirty migration version. Any future repair/force operation must be an intentionally separate, explicitly versioned break-glass command.

Connector loading and configured-actor synchronization are reconciliation behavior rather than schema migration. Keep them out of the default production schema plan; if they need a deploy-time path, expose an explicit, separately named opt-in target and document its side effects.

### `migrate status`

- Is read-only.
- Reports current/expected version and dirty/incompatible state for every configured schema target.
- Uses the same status API as server startup validation in #817 so operators see the same naming and compatibility semantics.
- Exits non-zero if a target is not current or cannot be inspected.

## Locking and safety

- Use backend-native locks for schema stores where they exist (the PostgreSQL migration driver already uses an advisory lock).
- For targets without a suitable native lock, use a correctly bounded, renewable distributed lock as defined in #816.
- Treat the dedicated Job as the normal single migration owner, but remain safe if an operator accidentally starts two Jobs.
- Log the migration plan, each target start/finish, version/status, duration, and failure cause without leaking credentials.

## Kubernetes and deployment integration

Provide a dedicated migration Job for Kustomize and Helm deployments:

- uses the exact same AuthProxy image tag, ConfigMap, secret references, and mounted key material as the workload;
- runs `authproxy migrate --config=/etc/authproxy/config.yaml` with `restartPolicy: Never`, bounded retries, and retained failure diagnostics;
- is applied only after PostgreSQL/Redis and required dependencies are ready;
- is awaited to successful completion before applying/restarting AuthProxy application workloads;
- handles repeat deployments safely (a completed prior Job must not prevent the next image version from running; a failed Job must expose logs and stop the rollout).

Update the demo/development deployment workflows to sequence prerequisites → migration Job → AuthProxy rollout → seed/smoke test. Do not merely add the Job to the same unsequenced Kustomize apply as the application Deployment.

Update Helm values/templates and Kustomize overlays so production-style configurations set automatic migration off. Local developer instructions should instead use `authproxy serve --auto-migrate` only for disposable environments (#817).

## Acceptance criteria

- `authproxy migrate` succeeds against a fresh database and an already-current database, including configured ClickHouse app-metrics stores.
- `authproxy migrate status` accurately detects current, stale, and dirty targets without mutation.
- A migration error causes a clear non-zero command/Job failure and preserves the actionable root error.
- No AuthProxy service starts as part of the migration command.
- Kustomize and Helm render a migration Job using the release image/config/secrets.
- CI/deployment workflows wait for the Job before AuthProxy rollout and fail immediately when it fails.
- Production/demo values no longer use automatic migration; demo reset → migrate Job → deploy → seed → smoke test succeeds.
- Deployment and operational documentation explains normal migration, status inspection, failure diagnostics, and the deliberate no-automatic-force policy.

rmorlok:
## Summary

- add explicit migrate and migrate status commands for main-database, workflows, app-metrics, and sequential fail-fast all plans, including up/down and exact-version semantics
- make serve perform read-only compatibility verification by default and centralize development-only migration/reconciliation behind --auto-migrate
- remove persisted autoMigrate configuration and retain renewable Redis locking across database providers
- add production bootstrap for the root namespace and encryption keys without connector or actor reconciliation
- add Helm and Kustomize migration Jobs, sequence the demo deployment before the AuthProxy rollout, and update local/deployment documentation

## Validation

- go test ./...
- AUTH_PROXY_TEST_DATABASE_PROVIDER=postgres go test ./internal/database ./internal/workflows
- AUTH_PROXY_APP_METRICS_TEST_DATABASE_PROVIDER=postgres go test ./internal/app_metrics
- AUTH_PROXY_APP_METRICS_TEST_DATABASE_PROVIDER=clickhouse with the documented local ClickHouse credentials go test ./internal/app_metrics
- helm lint and Helm rendering, including the load-test release values
- kubectl kustomize for demo and dev overlays
- docker compose config --quiet
- workflow YAML parse validation
- yarn docs:build
- ./scripts/preflight.sh

Closes #817
Closes #818

### DIFF
--- a/internal/app_metrics/test_store.go
+++ b/internal/app_metrics/test_store.go
@@ -77,6 +77,34 @@ func newTestHarnessLogger() *slog.Logger {
 	return slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: slog.LevelError}))
 }
 
+// postgresAppMetricsMigrator configures pgtestdb's golang-migrate adapter to
+// create the same migrations table used by the production app-metrics path.
+type postgresAppMetricsMigrator struct {
+	*golangmigrator.GolangMigrator
+}
+
+func (m *postgresAppMetricsMigrator) Hash() (string, error) {
+	hash, err := m.GolangMigrator.Hash()
+	if err != nil {
+		return "", err
+	}
+	return hash + ":" + appMetricsMigrationsTable, nil
+}
+
+func (m *postgresAppMetricsMigrator) Migrate(
+	ctx context.Context,
+	db *sql.DB,
+	templateConfig pgtestdb.Config,
+) error {
+	options, err := url.ParseQuery(templateConfig.Options)
+	if err != nil {
+		return fmt.Errorf("parse postgres test database options: %w", err)
+	}
+	options.Set("x-migrations-table", appMetricsMigrationsTable)
+	templateConfig.Options = options.Encode()
+	return m.GolangMigrator.Migrate(ctx, db, templateConfig)
+}
+
 func mustNewBlankSqliteRequestEventsStore(t testing.TB) (RecordStore, RecordRetriever, *sql.DB) {
 	t.Helper()
 
@@ -117,10 +145,12 @@ func mustNewBlankPostgresRequestEventsStore(t testing.TB) (RecordStore, RecordRe
 		ForceTerminateConnections: true,
 	}
 
-	migrator := golangmigrator.New(
-		"migrations/postgres",
-		golangmigrator.WithFS(appMetricsMigrationsFs),
-	)
+	migrator := &postgresAppMetricsMigrator{
+		GolangMigrator: golangmigrator.New(
+			"migrations/postgres",
+			golangmigrator.WithFS(appMetricsMigrationsFs),
+		),
+	}
 
 	testDbConfig := pgtestdb.Custom(t, adminConfig, migrator)
 
@@ -156,21 +186,6 @@ func mustNewBlankPostgresRequestEventsStore(t testing.TB) (RecordStore, RecordRe
 		SSLMode:  scommon.NewStringValueDirectInline(sslMode),
 		Params:   params,
 	}}
-
-	// pgtestdb's golang-migrate adapter always uses schema_migrations. Production
-	// app-metrics migrations use their own table so they can share a database
-	// with the main schema; align the cloned test database with that layout.
-	setupDB, err := sql.Open(cfg.GetDriver(), cfg.GetDsn())
-	if err != nil {
-		t.Fatalf("failed to open postgres app_metrics test database: %v", err)
-	}
-	if _, err := setupDB.Exec(`ALTER TABLE schema_migrations RENAME TO app_metrics_schema_migrations`); err != nil {
-		_ = setupDB.Close()
-		t.Fatalf("failed to rename postgres app_metrics migration table: %v", err)
-	}
-	if err := setupDB.Close(); err != nil {
-		t.Fatalf("failed to close postgres app_metrics setup database: %v", err)
-	}
 
 	// pgtestdb already ran the migrator against the per-test database, so the
 	// SQL store can be built without re-migrating.
--- a/internal/app_metrics/test_store.go
+++ b/internal/app_metrics/test_store.go
@@ -17,6 +17,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/peterldowns/pgtestdb"
 	"github.com/peterldowns/pgtestdb/migrators/golangmigrator"
+	"github.com/rmorlok/authproxy/internal/migration"
 	scommon "github.com/rmorlok/authproxy/internal/schema/common"
 	sconfig "github.com/rmorlok/authproxy/internal/schema/config"
 	"github.com/rmorlok/authproxy/internal/util"
@@ -238,7 +239,7 @@ func mustNewBlankClickhouseRequestEventsStore(t testing.TB) (RecordStore, Record
 	logger := newTestHarnessLogger()
 	store := NewClickhouseRecordStore(cfg, logger).(*clickhouseRecordStore)
 	t.Cleanup(func() { _ = store.db.Close() })
-	if err := store.Migrate(context.Background()); err != nil {
+	if err := RunMigrations(context.Background(), cfg, logger, migration.DirectionUp, nil); err != nil {
 		t.Fatalf("failed to migrate clickhouse app_metrics test database: %v", err)
 	}
 
@@ -255,7 +256,7 @@ func mustBuildSqlStorePair(t testing.TB, cfg *sconfig.Database) (RecordStore, Re
 	t.Helper()
 
 	store, retriever, rawDb := buildSqlStorePairNoMigrate(t, cfg)
-	if err := store.(*sqlRecordStore).Migrate(context.Background()); err != nil {
+	if err := RunMigrations(context.Background(), cfg, newTestHarnessLogger(), migration.DirectionUp, nil); err != nil {
 		t.Fatalf("failed to migrate app_metrics test database: %v", err)
 	}
 	return store, retriever, rawDb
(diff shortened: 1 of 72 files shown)


internal/app_metrics/test_store.go
```diff
@@ -156,6 +157,21 @@ func mustNewBlankPostgresRequestEventsStore(t testing.TB) (RecordStore, RecordRe
 		Params:   params,
 	}}
 
+	// pgtestdb's golang-migrate adapter always uses schema_migrations. Production
+	// app-metrics migrations use their own table so they can share a database
+	// with the main schema; align the cloned test database with that layout.
+	setupDB, err := sql.Open(cfg.GetDriver(), cfg.GetDsn())
```
rmorlok: Is there no way to customize the migration table via an option so it is consistent across prod and the test environments when it's created and doesn't require this change to rename it?

rmorlok: Good call. The pgtestdb adapter does not expose a migration-table option directly, but its underlying golang-migrate PostgreSQL driver supports `x-migrations-table`. I added a small pgtestdb migrator wrapper that sets that option only on the migration connection, so the template creates `app_metrics_schema_migrations` directly. Its hash includes the table name to prevent reuse of an older template created with the default table, and the post-clone `ALTER TABLE` workaround is gone. Verified with the PostgreSQL app-metrics suite, `go test ./...`, and preflight in 6158268a.

PR #856, opened 2026-08-20
Isolate demo smoke resources by namespace

rmorlok:
## Summary

- configure external actors as independent namespace-scoped sync sources
- place smoke-user in root.smoke with only the permissions needed to list root.smoke connectors and create/get/proxy root.smoke.smoke-user connections
- create every smoke connector in root.smoke and every smoke connection in root.smoke.smoke-user
- keep root and smoke actor keys in separate Kubernetes Secrets and mounts
- exercise dynamic and seeded OAuth/API-key flows with isolated per-run connector definitions

## Gaps found and addressed

1. The remote smoke helper used one namespace for admin actors, user actors, connectors, and connections. Those concerns are now modeled independently and asserted through the live API.
2. External actor synchronization supported one key directory and one permission set, effectively coupling all configured actors to root. The configuration now requires namespace-keyed sources, each with its own keysPath and permissions. The legacy single-source shape is intentionally rejected, and repository configs plus Helm output have been migrated.
3. Development namespace migration did not create namespaces required by the configured actor sources. It now creates each source namespace and its parents before synchronization.
4. The demo deployments mounted all actor keys together. Root and root.smoke now use separate Secrets and mounts; workflows remove any stale smoke-user key from the root Secret.
5. OAuth state assumed the actor and connection shared a namespace. State now carries actorNamespace and connectionNamespace separately and still decodes legacy namespace state during rolling deployments.
6. Seeded smoke tests depended on root connector visibility. The admin now copies seeded definitions into uniquely labelled, per-run root.smoke connectors, so smoke-user remains restricted to root.smoke and archived copies cannot collide with later runs.
7. The OAuth integration helper mirrored the legacy state JSON shape and discarded the new namespace fields during PKCE state mutation. It now preserves both fields while retaining the legacy field for rolling-upgrade rejection tests.

## Verification

- go test ./internal/schema/config ./internal/config ./internal/apauth/tasks ./internal/core ./internal/service/...
- go test ./internal/auth_methods/oauth2
- go test -tags smoke ./smoke -run ^$ (from integration_tests)
- full live TestRemote suite against the PR environment: dynamic OAuth, seeded catalog, seeded OAuth, and seeded API key all passed
- helm lint and helm template for the AuthProxy chart
- kubectl kustomize for the dev and demo overlays
- workflow and overlay YAML parse checks
- yarn docs:build
- ./scripts/preflight.sh
- all GitHub checks green, including the integration suite, Helm install, and Deploy Dev smoke test

### DIFF
--- a/deploy/kustomize/authproxy-demo/overlays/dev/authproxy-config.yaml
+++ b/deploy/kustomize/authproxy-demo/overlays/dev/authproxy-config.yaml
@@ -42,6 +42,9 @@ systemAuth:
     path: /etc/authproxy/keys/encryption/global_aes.key
   actors:
     keysPath: /etc/authproxy/keys/actors
+    namespaceByExternalId:
+      smoke-user: root.smoke
+    syncCronSchedule: "* * * * *"
     permissions:
       - namespace: root.**
         resources:
--- a/deploy/kustomize/authproxy-demo/overlays/dev/authproxy-config.yaml
+++ b/deploy/kustomize/authproxy-demo/overlays/dev/authproxy-config.yaml
@@ -41,13 +41,27 @@ systemAuth:
   globalAesKey:
     path: /etc/authproxy/keys/encryption/global_aes.key
   actors:
-    keysPath: /etc/authproxy/keys/actors
-    namespaceByExternalId:
-      smoke-user: root.smoke
+    root:
+      keysPath: /etc/authproxy/keys/actors/root
+      permissions:
+        - namespace: root.**
+          resources:
+            - "*"
+          verbs:
+            - "*"
+    root.smoke:
+      keysPath: /etc/authproxy/keys/actors/smoke
+      permissions:
+        - namespace: root.smoke
+          resources:
+            - connectors
+          verbs:
+            - list
+        - namespace: root.smoke.{{external_id}}
+          resources:
+            - connections
+          verbs:
+            - create
+            - get
+            - proxy
     syncCronSchedule: "* * * * *"
-    permissions:
-      - namespace: root.**
-        resources:
-          - "*"
-        verbs:
-          - "*"
--- a/internal/auth_methods/oauth2/state.go
+++ b/internal/auth_methods/oauth2/state.go
@@ -22,12 +22,9 @@ import (
 )
 
 type state struct {
-	Id                  apid.ID `json:"id"`
-	ActorNamespace      string  `json:"actorNamespace,omitempty"`
-	ConnectionNamespace string  `json:"connectionNamespace,omitempty"`
-	// LegacyNamespace keeps OAuth flows created before actor and connection
-	// namespaces were stored independently valid across a rolling deployment.
-	LegacyNamespace        string    `json:"namespace,omitempty"`
+	Id                     apid.ID   `json:"id"`
+	ActorNamespace         string    `json:"actorNamespace,omitempty"`
+	ConnectionNamespace    string    `json:"connectionNamespace,omitempty"`
 	ActorId                apid.ID   `json:"actorId"`
 	ConnectorId            apid.ID   `json:"connectorId"`
 	ConnectorVersion       uint64    `json:"connectorVersion"`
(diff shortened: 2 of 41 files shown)


deploy/kustomize/authproxy-demo/overlays/dev/authproxy-config.yaml
```diff
@@ -42,6 +42,9 @@ systemAuth:
     path: /etc/authproxy/keys/encryption/global_aes.key
   actors:
     keysPath: /etc/authproxy/keys/actors
+    namespaceByExternalId:
```
rmorlok: Let's change this to have multiple syncs for actors.

```yaml
  actors:
    root:
      keysPath: /etc/authproxy/keys/actors/root
      permissions:
        - namespace: root.**
          resources:
            - "*"
          verbs:
            - "*"
    root.smoke:
      keysPath: /etc/authproxy/keys/actors/smoke
      permissions:
        - namespace: root.smoke.**
          resources:
            - "*"
          verbs:
            - "*"
    syncCronSchedule: "* * * * *"
```

For the smoke user please also use minimal permissions for the required actions. Something like the ability to list connectors in the `root.smoke` namespace and connect connections into the `root.smoke.<external_id>` namespace where it leverage the templated namespace feature for the permission.

rmorlok: Addressed in 539e4175. Actor directories are now keyed by namespace with a shared sync schedule; the demo mounts separate root and smoke actor Secrets, removes the stale smoke key from the root Secret, and grants smoke-user only connectors:list in root.smoke plus connections:create|get|proxy in root.smoke.{{external_id}}. I also added schema/runtime sync coverage and updated the configuration reference.

rmorlok: Follow-up after live verification: the restricted root.smoke actor exposed that the seeded smoke cases still looked for connectors in root. Commit 34149d3a fixes that without widening permissions: the admin copies the seeded definitions into uniquely labelled, per-run connectors in root.smoke, and the smoke-user lists/connects only there. Dynamic OAuth plus the seeded catalog, OAuth, and API-key cases all pass against the live PR deployment.

internal/schema/config/schema_test.go
rmorlok, line 839: Please simplify this. We only need to support the format:

```yaml
  actors:
    root:
      keysPath: /etc/authproxy/keys/actors/root
      permissions:
        - namespace: root.**
          resources:
            - "*"
          verbs:
            - "*"
    root.smoke:
      keysPath: /etc/authproxy/keys/actors/smoke
      permissions:
        - namespace: root.smoke.**
          resources:
            - "*"
          verbs:
            - "*"
    syncCronSchedule: "* * * * *"
```

we can drop support for the previous format:

```yaml
  actors:
    keysPath: /etc/authproxy/keys/actors/root
    permissions:
      - namespace: root.**
         resources:
           - "*"
         verbs:
           - "*"
```

rmorlok: Addressed in b764dee1. Actor mappings now decode only as namespace-keyed external sources; the legacy keysPath-at-actors-root branch, schema alternative, source-level cron field, and runtime type cases are removed. I migrated the default, Docker, integration, schema fixtures, docs, and Helm rendering to use a root source, and added YAML plus JSON-schema coverage proving the old shape is rejected.

internal/auth_methods/oauth2/state.go
```diff
@@ -22,8 +22,12 @@ import (
...
 
 type state struct {
-	Id                     apid.ID   `json:"id"`
-	Namespace              string    `json:"namespace"`
+	Id                  apid.ID `json:"id"`
+	ActorNamespace      string  `json:"actorNamespace,omitempty"`
+	ConnectionNamespace string  `json:"connectionNamespace,omitempty"`
+	// LegacyNamespace keeps OAuth flows created before actor and connection
+	// namespaces were stored independently valid across a rolling deployment.
+	LegacyNamespace        string    `json:"namespace,omitempty"`
```
rmorlok: Do no keep legacy namespace here. Make this a breaking change.

rmorlok: Addressed in 5eaa328e. The legacy namespace field and both fallback accessors are removed, so OAuth state now requires actorNamespace and connectionNamespace directly. Namespace comparisons and token-exchange telemetry use those fields without compatibility behavior, and the validity test now proves a namespace-only legacy payload is invalid.

rmorlok: Follow-up in 952cf9b5: the integration-test state mirror also no longer exposes the legacy namespace field. The actor- and connection-namespace mismatch fixtures now populate actorNamespace and connectionNamespace independently, preserving their intended rejection coverage under the breaking state shape.

PR #859, opened 2026-08-20
Enforce actor permission namespace boundaries

rmorlok:
## Summary
- reject actor creates and upserts when a permission resolves above or outside the actor namespace
- constrain effective permission namespaces to the actor namespace as a defense-in-depth fallback for legacy or corrupted data
- cover exact, wildcard, sibling, prefix, and actor-templated namespace cases, including rejected updates preserving stored permissions

## Tests
- env AUTH_PROXY_TEST_DATABASE_PROVIDER=sqlite go test ./...
- env AUTH_PROXY_TEST_DATABASE_PROVIDER=postgres go test ./internal/database
- ./scripts/preflight.sh


internal/apauth/core/permissions_test.go
rmorlok, line 606: Please add test cases here where a permission would have allowed access based on it's namespace/selector but the actor's namespace would have constrained it further to not be allowed.

rmorlok: Added direct allowsForActor coverage in commit 2f2eefe5. The new cases use exact and wildcard permission namespaces that independently match the requested target, then verify access is denied because the target is outside the actor namespace.

integration_tests/helpers/setup_oauth2.go
```diff
@@ -59,6 +60,16 @@ func WithActor(externalID, namespace string) ActorOption {
 	}
 }
 
+// AllPermissionsForNamespace grants broad test permissions at or below the
+// supplied namespace without allowing the actor to escape its namespace.
+func AllPermissionsForNamespace(namespace string) []aschema.Permission {
```
rmorlok: Let's move this to the `github.com/rmorlok/authproxy/internal/schema/resources/namespace` package to colocate with the existing `AllPermissions()` helper.

rmorlok: Addressed in e0a5e8e1. I moved the helper out of the integration package and colocated it with the existing AllPermissions() constructor in internal/schema/auth, since it returns []auth.Permission; it uses namespace.WildcardSuffix for the matcher. AllPermissions() now delegates to the namespace-scoped helper, all integration call sites use the shared API, and schema-level root/tenant coverage was added.

PR #871, opened 2026-08-21
Isolate demo users and connections under root.demo

rmorlok:
## Summary

- keep demo-admin as a synced, self-signing actor while signing demo-user and fresh-user JWTs with the configured system JWT key
- provision root.demo and the least-privilege demo-user through the deployment seed API; dynamically provision fresh users from full actor claims
- publish demo connectors in root.demo and isolate every Marketplace user connection under root.demo.<external_id>
- automatically run the seed job during demo deploys and remove stale public-key actors
- infer one exact permitted child connection namespace when clients omit intoNamespace, preserving the current Marketplace request shape
- update smoke cloning, local compose seeding, and documentation

## Migration and gaps

Connector namespaces are immutable. The seed job therefore creates new root.demo connector records and archives the legacy seeded root connectors. Existing historical records remain in root as archived resources.

The system JWT key is used without the internal systemSigned claim. That claim selects GlobalAES/HMAC and is reserved for AuthProxy internal token transfer; external demo JWTs use the configured asymmetric system JWT signing key.

## Validation

- ./scripts/preflight.sh
- AUTH_PROXY_TEST_DATABASE_PROVIDER=sqlite go test ./internal/core ./internal/routes
- go test ./demos/shell/backend ./demos/seed/backend
- go test -tags smoke ./smoke -run ^$ (integration_tests module)
- yarn docs:build
- kubectl kustomize for demo, demo/seed, dev, and dev/seed overlays
- docker compose config --quiet for demos/shell/compose

### DIFF
--- a/demos/shell/README.md
+++ b/demos/shell/README.md
@@ -24,16 +24,24 @@ before redirecting them into the appropriate AuthProxy UI.
                             └──────────────────────────────────────────────────┘
 ```
 
-The backend holds the demo actor private key for pre-provisioned identities and
-the host JWT private key for just-in-time provisioning. Selecting **Fresh user**
-creates a random `fresh-user-<uuid>` external ID, includes the complete actor in
-the JWT, and lets AuthProxy create that actor during session initiation.
+The backend holds the demo administrator's private key and the system JWT
+private key. The pre-provisioned `demo-admin` continues to self-sign. The
+API-provisioned `demo-user` receives a subject-only system-key-signed token, so
+its stored actor permissions remain authoritative. Selecting **Fresh user**
+creates a random `fresh-user-<uuid>` external ID and includes the complete,
+least-privilege actor in a system-key-signed JWT so AuthProxy can create it
+during session initiation.
+
+Demo Marketplace actors live in `root.demo`. They can list the demo connectors
+there and manage connections only in `root.demo.<external_id>`, isolating each
+user's connections from every other demo user.
 
 ## Running locally
 
 Pre-requisites:
 - AuthProxy server running locally (`go run ./cmd/server serve --config=./dev_config/default.yaml all`)
-- A configured admin actor in your AuthProxy config that the demo shell can use as its signing identity
+- A configured `demo-admin` actor in your AuthProxy config that the demo shell can use as its signing identity
+- A `root.demo` namespace, seeded demo connectors, and an API-created `demo-user` actor (the hosted deployment's seed job provisions these)
 - Node + Yarn pinned via Volta (see root `package.json`)
 
 ### 1. Generate the demo admin keypair (one-time, local)
@@ -49,7 +57,7 @@ Add the public key as an admin actor in `dev_config/default.yaml`:
 ```yaml
 systemAuth:
   actors:
-    - externalId: demo-shell
+    - externalId: demo-admin
       key:
         publicKey:
           path: ./demos/shell/dev_keys/demo-shell.pub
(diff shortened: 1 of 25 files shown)


deploy/kustomize/authproxy-demo/overlays/demo/seed/seed-config.yaml
```diff
@@ -4,27 +4,33 @@ metadata:
...
-        namespace: root
-        labels:
-          demo: "true"
-          role: user
-      - externalId: grafana
-        namespace: root
-        labels:
-          demo: "true"
-          role: datasource
+    legacyConnectorNamespaces:
```
rmorlok: Take out `legacyConnectorNamespaces`. I will manually clean up the data in demo.

demos/shell/README.md
```diff
@@ -24,16 +24,24 @@ before redirecting them into the appropriate AuthProxy UI.
                             └──────────────────────────────────────────────────┘
 ```
 
-The backend holds the demo actor private key for pre-provisioned identities and
-the host JWT private key for just-in-time provisioning. Selecting **Fresh user**
-creates a random `fresh-user-<uuid>` external ID, includes the complete actor in
-the JWT, and lets AuthProxy create that actor during session initiation.
+The backend holds the demo administrator's private key and the system JWT
+private key. The pre-provisioned `demo-admin` continues to self-sign. The
```
rmorlok: Don't say "continues" here. Only talk about the state of the world in READMEs and docs, not how the application evolves.

internal/core/service_connection_initiate.go
```diff
@@ -52,6 +52,18 @@ func (s *service) InitiateConnection(
 	targetNamespace := c.GetNamespace()
 	if req.HasIntoNamespace() {
 		targetNamespace = req.IntoNamespace
+	} else if val.ValidateNamespaceResourceId(targetNamespace, c.GetId().String()) != nil {
+		// Existing clients omit intoNamespace. When their permissions grant
+		// connection creation in one exact child namespace, use that unambiguous
+		// destination rather than forcing every client to duplicate tenancy
+		// routing logic.
```
rmorlok: Let's instead make this comment:

```go
		// If the `intoNamespace` option is not specified, attempt to infer where
		// the connection should be placed based on the actor's permissions.
        // Try to colocate the connection with the connector first (use the
        // connector's namespace). If that is not possible, see if the actor has
        // a single namespace they are authorized to create connections in that
        // is a child of the connector's namespace. If so, use that. If permissions 
        // would be ambiguous, reject and force the client to specify. This is to
        // cover the common case where actors can only connect into their own
        // namespace.
```

internal/core/service_connection_initiate.go
```diff
@@ -116,3 +128,18 @@ func (s *service) InitiateConnection(
 
 	return connection.advanceToStep(ctx, first, flow, req.ReturnToUrl)
 }
+
+func inferConnectionNamespace(connectorNamespace, actorNamespace string, allowedNamespaces []string) (string, bool) {
```
rmorlok, line 141: Add some variation of the explanation that I gave in the comment where this method is called (`inferConnectionNamespace`) as the comment on this method.

internal/core/service_connection_initiate.go
rmorlok: Replaced the call-site comment with the requested explanation of connector colocation, single permitted child-namespace inference, and ambiguous-permission rejection. Addressed in 040a49b4.

demos/shell/README.md
rmorlok: Updated the README to describe only the current behavior: demo-admin self-signs, without evolutionary wording. Addressed in 040a49b4.

deploy/kustomize/authproxy-demo/overlays/demo/seed/seed-config.yaml
rmorlok: Removed legacyConnectorNamespaces from both deployment configs and removed the seed backend legacy-connector archival support and tests. Existing demo data will now be left for manual cleanup. Addressed in 040a49b4.

internal/core/service_connection_initiate.go
rmorlok: Added a method comment explaining that inference requires one exact, unambiguous namespace that is a child of both the connector and actor namespaces; invalid, wildcard, unrelated, or ambiguous matches require explicit intoNamespace. Addressed in 040a49b4.

PR #891, opened 2026-08-30
Migrate Namespace API to v1alpha1 resources

### LINKED ISSUE #834: Migrate Namespace API and configuration to the v1alpha1 resource shape
Parent: #829

## Dependencies

- #831
- #833

## Scope

- Define NamespaceSpec/NamespaceStatus and map identity, name, labels, annotations, and parent namespace path into metadata.
- Update core/iface, database conversion boundaries, `/api/v1` routes, config, lists, references, validation, and tests.
- Derive canonical namespace paths from parent metadata plus name, including explicit root behavior.
- Update namespace authorization and immutable-field rules without moving persistence models into the public schema.

## Acceptance criteria

- [ ] Namespace CRUD/list responses use `authproxy.net/v1alpha1` and `kind: Namespace`/`NamespaceList`.
- [ ] Parent/path/defaulting behavior is tested for root and nested namespaces.
- [ ] No flat namespace API/config object remains.

rmorlok:
Closes #834

## Summary

- define the resource-owned `authproxy.net/v1alpha1` `Namespace`/`NamespacePatch` contracts and JSON Schema
- map the canonical namespace path to `metadata.id`, the final segment to `metadata.name`, and the parent path to `metadata.namespace` (omitted for `root`)
- expose encryption-key selection through `spec.encryptionKeyRef` and lifecycle state through server-owned `status.state`
- convert between the resource contract and the existing flat database model at the core boundary; no SQL schema or data-conversion job is added
- migrate namespace CRUD, list, and key-subresource requests/responses to resource envelopes, including `NamespaceList` pagination
- centralize metadata-patch validation/application so omitted maps remain unchanged while explicit empty maps clear labels or annotations
- regenerate API/admin OpenAPI artifacts and update direct core callers, including OAuth integration tests

## Example

```yaml
apiVersion: authproxy.net/v1alpha1
kind: Namespace
metadata:
  id: root.acme.billing
  name: billing
  namespace: root.acme
  labels:
    team: platform
spec:
  encryptionKeyRef:
    apiVersion: authproxy.net/v1alpha1
    kind: Key
    id: key_global
status:
  state: active
```

This is an intentional breaking change to the existing `/api/v1` contract. Downstream Terraform and JavaScript/frontend consumers remain assigned to their dependent migration issues on the feature branch.

## Verification

- `env AUTH_PROXY_TEST_DATABASE_PROVIDER=sqlite go test ./internal/... -count=1`
- PostgreSQL: affected `internal/core` and `internal/routes` namespace/rate-limit tests
- integration-test module compile for `integration_tests/oauth2`
- `./scripts/preflight.sh`


internal/routes/namespaces.go
```diff
@@ -423,7 +395,26 @@ func (r *NamespacesRoutes) update(gctx *gin.Context) {
 		}
 	}
 
-	apgin.APIJSON(gctx, http.StatusOK, NamespaceToJson(ns))
+	if ref := req.Spec.EncryptionKeyRef; ref != nil {
```
rmorlok: When specifying a reference to a key, I would like to be able to do it with either the id, or the namespace + name reference. For places we have references to other objects being specified in resources, this is the general pattern I'd like to be able to follow. Can we make sure that works here and set the general pattern for other places this will be used?

rmorlok: Addressed in 22016f4c. ObjectReference now establishes the shared identity pattern of either an immutable id or namespace + name, with reusable target-specific validation options. Namespace encryptionKeyRef accepts both forms, and the core resolves namespaced key references to their canonical ID before persistence. Added schema/unit/route coverage for both validation and create/update resolution.

internal/core/service_key.go
```diff
@@ -27,6 +29,46 @@ func (s *service) GetKey(ctx context.Context, id apid.ID) (iface.Key, error) {
 	return wrapKey(*ek, s), nil
 }
 
+func (s *service) ResolveKeyReference(
+	ctx context.Context,
+	reference meta.ObjectReference,
+) (iface.Key, error) {
```
rmorlok, line 35: Please add tests for this method and also include a test case both the id and namespace name is presented but they actually point to different resources. In that scenario it should error.

rmorlok: Addressed in 12a76dec. ResolveKeyReference now resolves and compares both identity forms when id plus namespace/name are supplied, returning ErrInvalidArgument if they identify different keys. Added direct method tests for invalid, ID-only, namespace/name-only, matching dual identity, mismatched dual identity, and not-found references, plus a route-level mismatch test that verifies a 400 response.

internal/routes/namespaces.go
```diff
@@ -18,50 +17,25 @@ import (
...
-		Name:        ns.GetName(),
-		State:       schemaapi.NamespaceState(ns.GetState()),
-		KeyId:       ekId,
-		Labels:      ns.GetLabels(),
-		Annotations: ns.GetAnnotations(),
-		CreatedAt:   ns.GetCreatedAt(),
-		UpdatedAt:   ns.GetUpdatedAt(),
-	}
-}
+var _ = schemaapiopenapi.ListNamespacesResponseJson{}
```
rmorlok, line 30: What is the purpose of this assignment?

internal/schema/resources/namespace/patch_serialization.go
```diff
@@ -0,0 +1,63 @@
...
+	if err := util.DecodeJSONStrict(data, &wire); err != nil {
+		return err
+	}
+
+	*p = NamespaceSpecPatch{}
+	if wire.EncryptionKeyRef == nil {
+		return nil
+	}
+	if bytes.Equal(bytes.TrimSpace(wire.EncryptionKeyRef), []byte("null")) {
+		return fmt.Errorf("encryptionKeyRef must not be null; omit it to leave the key unchanged")
```
rmorlok: Two questions:

1. Is this requiring all namespaces to have encryption keys? Because that is not correct
2. Even though there is a separate operation for clearing the encryption key, it's ok to support it through the normal patch flow. Setting an explicit nil can do that operation as long as all the correct business logic is carried out.

internal/routes/namespaces.go
rmorlok: It is a compile-time-only reference that keeps the OpenAPI adapter import valid. The Swagger annotation below references schemaapiopenapi.ListNamespacesResponseJson from a comment, but Go does not count comment references as import usage. I added a comment explaining that in 3a98c2e6; it has no runtime behavior.

internal/schema/resources/namespace/patch_serialization.go
rmorlok: Addressed in 3a98c2e6. Namespace encryption keys remain optional: an omitted encryptionKeyRef means no change, a reference sets the key, and an explicit null now clears it. NamespaceSpecPatch tracks field presence across strict JSON/YAML decoding and marshaling, ApplyTo preserves the distinction, and the PATCH route invokes the existing ClearNamespaceKey business operation for null. The JSON Schema and README now describe the same semantics, with unit/schema/route coverage including omission preserving an existing key and null clearing it.

internal/schema/resources/namespace/resource.go
```diff
@@ -0,0 +1,406 @@
...
+	EncryptionKeyRef *meta.ObjectReference `json:"encryptionKeyRef,omitempty" yaml:"encryptionKeyRef,omitempty"`
+}
+
+// NamespaceSpecPatch contains presence-aware desired-state updates. Keep this
+// separate from NamespaceSpec so future scalar spec fields cannot accidentally
+// lose omitted-versus-zero patch semantics.
+type NamespaceSpecPatch struct {
+	EncryptionKeyRef *meta.ObjectReference `json:"encryptionKeyRef,omitempty" yaml:"encryptionKeyRef,omitempty"`
+
+	encryptionKeyRefPresent bool
```
rmorlok, line 57: Why is this extra tracking variable needed?

rmorlok: The patch needs three states: omitted means leave the current key unchanged, a non-nil reference means set it, and explicit null means clear it. Go JSON/YAML decoding collapses both an omitted field and an explicit null into EncryptionKeyRef == nil, so the pointer alone cannot preserve the wire intent after decoding. encryptionKeyRefPresent records that third state; HasEncryptionKeyRef hides the implementation detail and also treats direct non-nil field assignment as present. Without the presence bit, the route would have to either clear on omission or ignore explicit null. If this pattern recurs across resource patches, we should encapsulate the same value/null/presence state in a shared optional-nullable helper.

Issue #838, opened 2026-08-16
Use the restricted v1alpha1 Actor resource in JWT claims


rmorlok:
Parent: #829

Target branch: `codex/k8s-resource-api`

The implementation PR must carry the `project:k8s-resource-api` label and target the feature branch, not `main`.

## Dependencies

- #837

## Scope

- Replace the flat JWT actor claim with the restricted Actor resource shape and remove legacy claim decoding.
- Require `sub == actor.spec.externalId` and agreement between the top-level namespace claim and `actor.metadata.namespace`.
- Forbid `metadata.id`, timestamps, status, and signing-key material inside JWT actors.
- Treat actor-spec permissions as the base and allow top-level JWT permissions only to narrow them.
- Update signing, verification, request authentication, configured actors, CLI/signing-proxy tooling, fixtures, examples, and tests.

## Acceptance criteria

- [ ] New claims round-trip through every signer/verifier and old flat claims are rejected.
- [ ] Subject/namespace/permission escalation mismatches fail with clear tests.
- [ ] JWT examples and error messages use `authproxy.net/v1alpha1` Actor objects.

PR #899, opened 2026-09-05
Migrate Actor resources to v1alpha1

### LINKED ISSUE #837: Migrate Actor core, API, and configuration to v1alpha1
Parent: #829

## Dependencies

- #831
- #833
- #834

## Scope

- Create the resource-owned Actor type with metadata, ActorSpec, and only useful server-owned status.
- Place external subject and base permissions in spec; place namespace and safe identity fields in metadata.
- Update actor core/iface, database conversions, API/config binding, authorization, references, lists, redaction, and tests.
- Keep JWT-specific restrictions out of the general resource type so the JWT migration can apply a dedicated validation policy.

## Acceptance criteria

- [ ] Actor API/config uses `kind: Actor` and no flat actor body remains outside explicitly temporary compile steps on the feature branch.
- [ ] External ID, permission, namespace, immutable metadata, and persistence conversion rules are tested.
- [ ] Other packages depend on resource/core interfaces rather than `internal/schema/api` actor DTOs.

rmorlok:
Closes #837

## Summary

- add the canonical `authproxy.net/v1alpha1` `Actor` resource, presence-aware patch schema, lifecycle validation, and resource package documentation
- add the core Actor abstraction and service CRUD/list layer, preserving encrypted signing keys as write-only input and exposing only `status.signingKeyConfigured`
- migrate Actor routes, static/external-source configuration, auth consumers, object reference resolution, SDKs, Admin UI, Terraform, Grafana, demos, API schemas, Swagger, and public documentation to the resource envelope
- preserve the existing JWT Actor claim shape for the dedicated JWT migration in #838

This is an intentionally breaking v1 contract change. It does not add a SQL migration or a background data rewrite; existing demo environments can be reset as planned for the feature branch.

## Screenshot

Admin Actor detail backed by the canonical resource payload (visual presentation is intentionally unchanged):

Canonical Actor detail

## Validation

- `./scripts/preflight.sh`
- `AUTH_PROXY_TEST_DATABASE_PROVIDER=sqlite go test ./internal/schema/... ./internal/core ./internal/apauth/service ./internal/apauth/tasks ./internal/routes ./internal/service/...`
- `yarn workspace @authproxy/api typecheck`
- `yarn workspace @authproxy/api test`
- `yarn workspace @authproxy/admin typecheck`
- `yarn workspace @authproxy/admin build`
- `yarn workspace @authproxy/admin build-storybook`
- `yarn docs:build`
- `go test ./internal/client ./internal/resources ./internal/datasources` from `terraform/provider`
- `go test ./pkg/plugin` from `plugins/grafana/authproxy-datasource`

Stacked on `codex/k8s-resource-api`.


internal/core/service_actor.go
```diff
@@ -0,0 +1,287 @@
...
+	if key == nil {
+		return nil, nil
+	}
+
+	data, err := json.Marshal(key)
+	if err != nil {
+		return nil, fmt.Errorf("marshal actor signing key: %w", err)
+	}
+
+	encrypted, err := s.encrypt.EncryptStringGlobal(ctx, string(data))
```
rmorlok: Why is this being encrypted with the global key? Shouldn't this select the key based on the namespace of the actor?

rmorlok: Good catch — yes, actor signing material should follow the actor namespace encryption hierarchy. Updated in 740dec4f: API create and signing-key replacement now call EncryptStringForNamespace with metadata.namespace, and configured-actor sync does the same. That selects the namespace key, inherits from ancestors, and only falls back to the global key when no namespace key is configured. I added focused tests for all three paths.

PR #900, opened 2026-09-05
Use v1alpha1 Actor resources in JWT claims

### LINKED ISSUE #838: Use the restricted v1alpha1 Actor resource in JWT claims

rmorlok:
## Summary

- represent embedded JWT actors as restricted authproxy.net/v1alpha1 Actor resources while preserving subject-only tokens
- reject legacy flat actors plus database identity, lifecycle status, timestamps, signing material, unknown fields, subject/namespace mismatches, and permission escalation
- validate the contract in the shared parser and progressive claims builder, then convert to the runtime actor at the auth-service boundary
- update JWT schemas, fixtures, CLI expectations, and integration/security documentation

## Validation

- go test ./...
- go test -run ^$ ./... (integration_tests module)
- yarn docs:build
- ./scripts/preflight.sh

Closes #838

### DIFF
--- a/internal/apauth/jwt/test_data/example.json
+++ b/internal/apauth/jwt/test_data/example.json
@@ -33,7 +33,7 @@
   },
   "permissions": [
     {
-      "namespace": "root.prod.payments",
+      "namespace": "root.prod.**",
       "resources": ["connections"],
       "verbs": ["list"]
     }
--- a/internal/apauth/jwt/test_data/example.json
+++ b/internal/apauth/jwt/test_data/example.json
@@ -8,17 +8,32 @@
   "exp": 1735689600,
   "nbf": 1735603200,
   "iat": 1735603200,
-  "jti": "550e8400-e29b-41d4-a716-446655440000",
+  "jti": "jwt_550e8400-e29b-41d4-a716-446655440000",
+  "namespace": "root.prod",
   "actor": {
-    "id": "user/12345",
-    "permissions": [
-      "root.prod.actor-1234#/connections/_initiate",
-      "root.prod.**#/connections/_list"
-    ]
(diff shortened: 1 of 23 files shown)


internal/apauth/jwt/test_data/example.json
```diff
@@ -8,17 +8,32 @@
...
+          "resources": ["connections"],
+          "verbs": ["get", "list"]
+        }
+      ]
+    }
   },
   "permissions": [
     {
-      "namespace": "root.prod.**",
+      "namespace": "root.prod.payments",
```
rmorlok: Why was this value changed? Wasn't the old value a valid scenario?

rmorlok: Yes—the original root.prod.** value is valid because a token restriction may equal the actor base permission; it just cannot be broader. I changed it only to make the example visibly narrower, which was unnecessary. Restored it in aec98fca.

PR #903, opened 2026-09-05
Migrate Connection resources and actions to v1alpha1

### LINKED ISSUE #840: Migrate Connection resources and setup actions to v1alpha1
Parent: #829

## Dependencies

- #837
- #839
- #833

## Scope

- Define ConnectionSpec with a typed generation-specific connector reference and connector-authored desired configuration. Define ConnectionStatus with lifecycle/setup/probe observations and grouped configuration status containing a server-derived schema-satisfaction flag and aggregate configuration schema.
- Update v1 CRUD/list routes, core/iface, database conversion, authorization, caching, telemetry labels, configuration projection, credential redaction, and tests.
- Convert typed setup/start/complete/cancel/probe/migration actions to explicit resource-shaped action contracts where appropriate.
- Leave OAuth redirects/callbacks, raw proxy streams, and third-party payloads unwrapped.

## Acceptance criteria

- [ ] Connection CRUD/list returns v1alpha1 resources with stable typed references and status.
- [ ] Connector-authored configuration is returned in cleartext under spec.configuration; status.configuration contains an aggregate schema for form-defined fields and a configured flag reporting whether the current configuration satisfies that schema.
- [ ] Auth-method credentials remain in dedicated encrypted storage and are not returned through Connection resources; setup-action form data remains write-only/redacted.
- [ ] Proxy streaming and OAuth protocol behavior is unchanged.

rmorlok:
Closes #840

## Summary

- add the canonical authproxy.net/v1alpha1 Connection resource schema with a generation-specific connector reference, connector-authored desired configuration, and lifecycle, health, setup, and configuration status; Connections remain namespace-scoped rather than actor-bound
- return decrypted connector-authored values under spec.configuration and group a schema-satisfaction flag with the aggregate form-field JSON Schema under status.configuration; required fields from unconditional setup steps are retained, while requirements gated by JavaScript predicates remain enforced by their setup steps; auth-method credentials remain in dedicated encrypted stores and are not exposed through the Connection resource
- migrate Connection CRUD, list, setup, lifecycle, force-state, disconnect, and connector-version migration bodies to resource-shaped typed actions while leaving OAuth and proxy protocol surfaces unwrapped
- keep setup-action form data irreversibly write-only, including for callers with secret-replay permission
- update core, database conversion, route, CLI compile surfaces, docs, schema fixtures, structured JSON/YAML serialization, and generated Swagger without changing the Connection SQL schema

## Verification

- ./scripts/preflight.sh
- go test ./internal/schema/common ./internal/schema/resources/connectors ./internal/schema/resources/connection ./internal/schema/api ./internal/core ./internal/routes -count=1
- yarn docs:build

### DIFF
--- a/docs/src/content/docs/reference/api.md
+++ b/docs/src/content/docs/reference/api.md
@@ -39,10 +39,10 @@ required namespace, resource, resource-ID, and verb scope. See
 ## Resource identity and names
 
 Namespace, actor, connector, connection, key, and rate-limit responses expose a
-human-readable `name` alongside their direct identity (`id`, or `path` for a
-namespace). Keep using the immutable ID in URLs, permissions, foreign-key
-fields, and stored references. Names are for display and discovery; renaming a
-resource does not change its URL.
+human-readable `metadata.name` alongside their direct identity (`metadata.id`,
+or the canonical path for a namespace). Keep using the immutable ID in URLs,
+permissions, foreign-key fields, and stored references. Names are for display
+and discovery; renaming a resource does not change its URL.
 
 Create requests for actors, connectors, connections, keys, and rate limits may
 include `name`. If it is omitted, AuthProxy generates the ID first and uses that
@@ -53,12 +53,28 @@ POST /api/v1/connections/_initiate
 Content-Type: application/json
 
 {
-  "connectorId": "cxr_01example",
-  "intoNamespace": "root.acme",
-  "name": "production-crm",
-  "returnToUrl": "https://app.example.com/integrations/complete"
+  "apiVersion": "authproxy.net/v1alpha1",
+  "kind": "ConnectionInitiate",
+  "metadata": {
+    "target": {
+      "apiVersion": "authproxy.net/v1alpha1",
+      "kind": "Connector",
+      "id": "cxr_01example",
+      "generation": 3
+    }
+  },
+  "spec": {
+    "intoNamespace": "root.acme",
+    "name": "production-crm",
+    "returnToUrl": "https://app.example.com/integrations/complete"
+  }
 }
 ```
+
+Omit the connector generation to select its primary generation. The setup
+result is a `ConnectionSetup` action whose `metadata.target` identifies the new
+connection and whose `status.type` is `redirect`, `form`, `verifying`,
+`complete`, or `error`.
 
 Rename through the immutable ID. The response keeps the same `id` and returns
 the new `name`:
@@ -68,7 +84,12 @@ PATCH /api/v1/connections/cxn_01example
 Content-Type: application/json
 
 {
-  "name": "production-salesforce"
+  "apiVersion": "authproxy.net/v1alpha1",
+  "kind": "Connection",
+  "metadata": {
+    "name": "production-salesforce"
+  },
+  "spec": {}
 }
 ```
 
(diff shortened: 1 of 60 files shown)


docs/src/content/docs/reference/api.md
```diff
@@ -94,7 +115,51 @@ GET /api/v1/connections?name=production-salesforce&namespace=root.acme
...
+    "namespace": "root.acme"
+  },
+  "spec": {
+    "connectorRef": {
+      "apiVersion": "authproxy.net/v1alpha1",
+      "kind": "Connector",
+      "id": "cxr_01example",
+      "generation": 3
+    },
+    "actorRef": {
```
rmorlok: What is `actorRef` in this context? Connections aren't bound to actors.

rmorlok: You are right—Connections are namespace-scoped and are not bound to an actor. I had interpreted the issue wording about typed actor references as creation provenance, but the initiating actor belongs on setup tokens, credential provenance, and audit/event records rather than in durable Connection spec. I removed actorRef, the core/database plumbing, and the new actor_id migrations in 49e14888, and updated the schema, docs, fixtures, tests, and generated OpenAPI accordingly.

internal/core/service_connection_initiate.go
```diff
@@ -27,19 +27,30 @@ func (s *service) InitiateConnection(ctx context.Context, req iface.InitiateConn
...
+		if errors.Is(err, database.ErrInvalidReference) {
+			return nil, httperr.BadRequest("invalid connector reference", httperr.WithInternalErr(err))
+		}
+		return nil, httperr.InternalServerError(httperr.WithInternalErr(err))
+	}
+
 	var c iface.Connector
-	if req.HasVersion() {
-		c, err = s.GetConnectorVersion(ctx, req.ConnectorId, req.ConnectorVersion)
+	if req.ConnectorRef.Generation != 0 {
```
rmorlok: Why do we do this second lookup here? Couldn't we use the connector reference returned to get what we need? Doesn't it have the same primary behavior?

rmorlok: The database ResolveConnectorReference call only resolves ID vs. namespace/name to the logical connector row; that row does not contain a versioned/decrypted definition, so initiation still needs a second lookup to obtain the iface.Connector passed through setup and connection creation. The generic core ResolveConnectorReference also does not currently have the same default: an explicit generation resolves exactly, but an omitted generation selects the newest definition version (which may be a draft). Connection initiation has the established and documented behavior that an omitted generation selects the primary version, so this path deliberately uses GetConnectorVersionForState(...Primary). Using the generic resolver here would silently change initiation semantics.

rmorlok: The shared resolver now has the required semantics after #905 merged: an omitted generation resolves the primary, while an explicit generation resolves exactly. I rebased this PR onto that change and updated connection initiation to call ResolveConnectorReference directly, removing the duplicated logical/version lookup in 6817d616.

internal/schema/resources/connection/README.md
```diff
@@ -0,0 +1,31 @@
...
+in `metadata`. `spec.connectorRef` pins the exact connector generation that
+interprets the connection.
+
+Connector-defined setup values appear under `spec.configuration`, flow through
+typed setup actions, and are encrypted at rest. API reads return those values
+in cleartext. Server-derived configuration state lives under
+`status.configuration`: `configured` reports whether encrypted configuration
+has been persisted, and `schema` contains an aggregate JSON Schema for the
+top-level fields from connector-authored preconnect and configure forms. The
+schema omits step-level `required` constraints so incomplete setup and skipped
```
rmorlok: Do we really need to omit required here? The resource has a configured setting:

```yaml
status:
  configuration:
    configured: true
    schema: ...
```

If `configured=true` doesn't that imply that all the setup steps are complete and the data in the configuration will match the schema? If it's `false` then that would mean that the data in the configuration is not sufficient for the schema.

rmorlok: Good catch. The prior configured value only reported whether an encrypted configuration blob existed, so it did not support that invariant. In e153c48f I changed configured to be computed by validating the current spec.configuration against the emitted aggregate schema, and retained required fields from unconditional form steps. Requirements from JavaScript-gated conditional steps remain omitted because their predicates cannot be represented faithfully in JSON Schema; otherwise a legitimately skipped step would make a complete configuration fail the aggregate schema. I also updated the package/public docs and added coverage for complete, incomplete, empty, invalid, and conditional configurations.

Issue #922, opened 2026-09-08
Implement sanitized apply history and three-way reconciliation


rmorlok:
Parent: #919

Depends on task 2. Define versioned authproxy.net/last-applied-configuration history, secret exclusions/presence tracking and annotation size checks. Compare previous desired/live/new desired; preserve unmanaged fields, remove formerly managed omitted fields where legal, treat lists atomically and generate resource-specific patches. Merge whole metadata maps with live state. Initial adoption warns and preserves unspecified fields. Implement --overwrite and --validate=strict|warn|ignore without accepting malformed identity or duplicate keys. Explicit write-only secrets are submitted, omitted secrets preserved unless an explicit supported clear is requested; never persist secret values or public hashes of low-entropy secrets in history. Reject redacted placeholders. Advance history only with successful resource writes where possible. Test omissions/null/empty/maps/arrays/adoption/drift/history failures and redaction.

Run focused tests and `./scripts/preflight.sh` before committing. Open a PR; do not merge.

Issue #919, opened 2026-09-08
Implement declarative ap apply command


rmorlok:
Implement client-side `ap apply` over AuthProxy REST APIs with Kubernetes-style resource manifests.

## Agreed behavior
- Identity must be an ID or namespace/name. Explicit metadata.namespace wins over --namespace; missing namespace with no ID is an error, never default to root. ID-only targets resolve their namespace from the server; supplied identity fields must agree. A missing explicit ID must not create another resource.
- Namespace metadata.namespace is the parent path; root is special.
- Create/update Namespace, Actor, Connector, Key, RateLimit. Connection support is limited to mutable metadata on existing connections.
- Use sanitized last-applied history and three-way reconciliation, preserving unrelated fields and handling omitted/null/empty values deliberately. Never store secrets in annotations; write-only fields have documented comparison limitations.
- Detect unchanged connector definitions before PATCH so repeated applies do not manufacture generations. Respect release intent and existing draft lifecycle.
- No server-side apply, field managers, force-conflicts, managed fields or subresource apply. Server dry-run and Kubernetes deletion flags remain unsupported until backend support exists.
- Batches are not transactional; prevalidate all input, order dependencies, skip failed dependents, report partial failures and exit nonzero.

## Initial command surface
`-f/--filename` (repeatable files, directories, URLs, stdin), `-R/--recursive`, `-n/--namespace`, `-l/--selector`, `--dry-run=none|client`, `--validate=strict|warn|ignore`, `--overwrite`, `-o/--output` (name/json/yaml), `--request-timeout`, and existing config/signing/endpoint flags. Client dry-run never contacts the cluster. Unknown fields are governed by validation mode; malformed documents, duplicate keys and invalid identity always fail. Structured output redacts secrets.

## Delivery
Implement the linked tasks in order. Each task includes its own focused tests and docs; the integration task completes cross-cutting coverage. The first PR exposes only client dry-run and rejects real apply until execution is implemented. Broader kubectl compatibility is tracked separately.

## Implementation sequence
- [ ] #920
- [ ] #921
- [ ] #922
- [ ] #923
- [ ] #924
- [ ] #925
- [ ] #926

PR #929, opened 2026-09-12
Add authenticated resource adapters for ap apply

### LINKED ISSUE #921: Add apply resource adapters and authenticated API client
Parent: #919

Depends on task 1. Add per-kind adapters for Namespace, Actor, Connector, Key, RateLimit and existing Connection metadata. Reuse canonical resource/patch types and CLI signing resolver, not Terraform internal packages. Resolve exact namespace/name with pagination or explicit ID; verify supplied identity consistency and fail missing explicit IDs. Detect aliases that resolve to one live ID. Use the admin endpoint for admin mode; reject Key batches without an admin endpoint before mutations. Add timeouts, response error handling, redaction-aware reads and create/update semantic validation. Test requests, permissions errors, pagination, ID/name mismatch, missing resources and service selection.

rmorlok:
Adds the authenticated REST layer for the next stages of `ap apply`. Documents can now be resolved by explicit ID or exact namespace/name across paginated results, checked against supplied identity, and validated using each resource's canonical create/patch contract. Batch resolution detects aliases targeting one live ID. Key resources are supported through both the API and admin API, with authorization enforced by the server. Explicit IDs and connector generations must exist; connections only support existing metadata updates.

The client provides single-resource create and patch methods, per-request timeouts, safe HTTP status errors, redaction metadata and write-only field information. Resource/patch factories, metadata access, ID validation, lifecycle validation, patch application and REST collection names are centralized in `internal/schema/registry`. The generic manifest scheme remains a decoder; the richer resource scheme exposes capability lookup and explicit patch decoding. Apply retains target resolution, endpoint selection and execution policy. The client reuses this registry and the CLI signing resolver, selects the API or admin API according to admin mode, refuses redirects, and does not retry mutations. Tests exercise all six resource adapters, signing/service selection, Key operations on both services and API permission denials, pagination and namespace filtering, conflicts/missing identities, nullable and immutable fields, redaction, deadlines and cancellation.

This is infrastructure for reconciliation and execution (#922/#923). The command still exposes offline client dry-run only; this PR does not enable cluster apply or promise concurrency protection. The package README documents that boundary.

Closes #921
Part of #919

## Validation
- `go test ./internal/schema/registry ./internal/schema/manifest ./internal/apply ./cmd/cli/config ./cmd/cli`
- `./scripts/preflight.sh`

### DIFF
--- a/internal/apply/adapters.go
+++ b/internal/apply/adapters.go
@@ -30,36 +30,61 @@ type resourceAdapter struct {
 func adapterFor(kind meta.Kind) (resourceAdapter, error) {
 	switch kind {
 	case actor.ActorKind:
-		return resourceAdapter{"actors", func() lifecycleResource { return actor.NewActorPatch() }, func(current any, p lifecycleResource) error {
-			_, err := p.(*actor.ActorPatch).ApplyTo(current.(*actor.Actor), nil)
-			return err
-		}}, nil
+		return resourceAdapter{
+			"actors",
+			func() lifecycleResource { return actor.NewActorPatch() },
+			func(current any, p lifecycleResource) error {
+				_, err := p.(*actor.ActorPatch).ApplyTo(current.(*actor.Actor), nil)
+				return err
+			},
+		}, nil
 	case namespace.NamespaceKind:
-		return resourceAdapter{"namespaces", func() lifecycleResource { return namespace.NewNamespacePatch() }, func(current any, p lifecycleResource) error {
-			_, err := p.(*namespace.NamespacePatch).ApplyTo(current.(*namespace.Namespace), nil)
-			return err
-		}}, nil
+		return resourceAdapter{
+			"namespaces",
+			func() lifecycleResource { return namespace.NewNamespacePatch() },
+			func(current any, p lifecycleResource) error {
+				_, err := p.(*namespace.NamespacePatch).ApplyTo(current.(*namespace.Namespace), nil)
+				return err
+			},
+		}, nil
 	case key.KeyKind:
-		return resourceAdapter{"keys", func() lifecycleResource { return key.NewKeyPatch() }, func(current any, p lifecycleResource) error {
-			_, err := p.(*key.KeyPatch).ApplyTo(current.(*key.Key), nil)
-			return err
-		}}, nil
+		return resourceAdapter{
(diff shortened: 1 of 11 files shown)


internal/apply/adapters.go
```diff
@@ -0,0 +1,145 @@
...
+
+// Adapters contain operation contracts, not resource registrations. Resource
+// decoding continues to use the shared schema registry.
+type resourceAdapter struct {
+	collection string
+	patch      func() lifecycleResource
+	merge      func(any, lifecycleResource) error
+}
+
+func adapterFor(kind meta.Kind) (resourceAdapter, error) {
```
rmorlok: All of this logic is per-resource type on on generic schema methods. Would it make more sense to put it in the internal/schema/registry package and extend the scheme to include these capabilities?

rmorlok: Addressed in f30e7407. The shared registry now owns resource/patch factories, metadata access, ID validation, lifecycle validation, patch application and REST collection names in one catalogue. ResourceScheme composes the generic manifest decoder with capability lookup; patch decoding is explicitly selected because it shares the resource GVK. Apply now delegates these capabilities and retains transport, target resolution, endpoint selection and execution policy. Registry and apply/CLI tests, including type-safety and metadata-isolation coverage, and preflight all passed.

internal/apply/client.go
```diff
@@ -0,0 +1,406 @@
...
+	Document Document
+	Current  *LiveResource
+}
+
+func (c *Client) checkKind(kind meta.Kind) error {
+	if _, err := resourceType(kind); err != nil {
+		return err
+	}
+	if kind == "Key" && !c.admin {
+		return fmt.Errorf("Key resources require admin mode and an admin API endpoint")
```
rmorlok: Please rebase this PR and update. This is not longer a restriction.

rmorlok: Rebased onto main, including #930, and removed the admin-only Key restriction in ddbe6403. Key lookup, creation, and updates now use whichever API service the caller selects; server authorization remains authoritative. Added regression coverage for Key operations on both services and HTTP 403 propagation from the main API, and updated the package documentation. Focused Go tests and ./scripts/preflight.sh pass.

Issue #923, opened 2026-09-08
Execute apply batches in dependency order with partial failure reporting


rmorlok:
Parent: #919

Depends on task 3. Build a stable dependency graph for namespace parents and explicit references; detect missing prerequisites and cycles before writes, including namespace/encryption-key cycles. Enable --dry-run=none with create/update/no-op execution and human or structured output. Skip dependents of failed resources, continue independent resources and return nonzero on partial failure. No transactional or rollback promise. Never retry ambiguous mutation outcomes blindly. Client dry-run remains offline and prints desired resources, not a claimed server execution plan. Tests cover dependency ordering, cycles, partial writes, cancellation, output errors and existing Connection restrictions.

Run focused tests and `./scripts/preflight.sh` before committing. Open a PR; do not merge.

PR #931, opened 2026-09-12
Add sanitized apply history and three-way reconciliation

### LINKED ISSUE #922: Implement sanitized apply history and three-way reconciliation

rmorlok:
Adds the reconciliation layer for `ap apply`: resolved resources now produce validated create, update, or unchanged plans using previous desired state, live state, and the new manifest. For example, removing an annotation previously managed by apply removes that annotation while preserving annotations added by other clients. Unchanged connector definitions stay out of PATCH requests.

The versioned `authproxy.net/last-applied-configuration` annotation excludes schema-tagged secrets and entire write-only key configurations. It records secret presence without values, masks, or hashes. History travels in the same resource write, with annotation size validation; malformed or mismatched history fails before mutation. Explicit secrets are submitted without comparison, omitted secrets are preserved, and object replacements that would replay masked credentials are rejected.

Maps reconcile by key and lists are atomic. Whole-object REST patches contain preserved unmanaged fields, and canonical resource contracts validate supported clears and immutable fields. `Overwrite: false` detects managed-field drift. The CLI now accepts `--validate=strict|warn|ignore` and `--overwrite`; validation modes filter unknown fields through embedded schemas referenced by the centralized registry, while retaining identity, duplicate-key, type, and placeholder checks.

The CLI still requires `--dry-run=client`; `--overwrite` has no effect in that mode. Batch execution is the next task (#923). Plans and patches may contain secrets and must not be logged. Explicit secrets cannot be classified as unchanged, and concurrent writes remain outside this stage's guarantees.

Validation:
- `go test -race ./internal/schema ./internal/apply ./internal/config` (concurrent compilation regression also reproduces races against the pre-fix code)
- `go test ./internal/apply ./internal/apserde ./internal/schema/registry ./internal/schema/manifest ./cmd/cli/config ./cmd/cli`
- `yarn docs:build`
- `./scripts/preflight.sh`

Closes #922
Part of #919


internal/apply/validation.go
```diff
@@ -0,0 +1,298 @@
...
+	if l.options.Validation == "" || l.options.Validation == ValidationStrict {
+		return nil
+	}
+
+	descriptor, err := resourceType(kind)
+	if err != nil {
+		return err
+	}
+
+	fieldSchemaMu.Lock()
```
rmorlok: Why do we need this mutex?

rmorlok: I added it to protect concurrent calls to `apply.Load` that reach schema compilation. `schema.CompileSchema` uses one shared `jsonschema.Compiler`, but its existing `compileMutex` only protects the outer cache lookup and insertion; it releases the lock before calling `schemaCompiler.Compile`. Compilation mutates the compiler's resource and compiled-schema state, so concurrent cache misses can race. Loading files within a single CLI invocation is sequential, so that path alone does not need this mutex.

However, this is the wrong place for the protection: `fieldSchemaMu` only coordinates apply callers and cannot protect against another caller, such as the config loader. The proper fix is inside `schema.CompileSchema`: keep the read-lock cache fast path, then acquire the write lock on a miss, recheck the cache, and hold that lock through compilation and insertion. That protects every caller and lets us remove this apply-specific mutex. I recommend making that change and adding a concurrent-compilation regression test run under `go test -race`.

rmorlok: Fixed in 40552b87. `schema.CompileSchema` now acquires its existing write lock on a cache miss, rechecks the cache, and holds the lock through compilation and cache insertion. The apply-specific mutex is removed. Added a cold-cache concurrent-compilation test covering repeated requests, different roots sharing references, cache reuse, and recovery after a compilation error. The test reports data races against the original implementation and passes with the fix. `go test -race ./internal/schema ./internal/apply ./internal/config` and `./scripts/preflight.sh` both pass.

internal/apply/history.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+func secretPaths(kind meta.Kind, resource any) [][]string {
+	paths := apserde.SecretPaths(resource)
+	switch kind {
+	case "Actor":
+		paths = append(paths, []string{"spec", "signingKey"})
+	case "Key":
+		paths = append(paths, []string{"spec", "keyData"})
+	}
```
rmorlok: Why do we need to encode paths to secrets here? That doesn't seem like a scalable solution as more resources are added.

rmorlok: Agreed—the resource-specific path list did not belong in apply. Fixed in cf79e523: canonical resource and patch fields now declare `apiwriteonly:"true"` alongside their existing JSON Schema `writeOnly` annotations. Shared `apserde.WriteOnlyPaths` / `SensitivePaths` discovery uses those declarations and existing `apiredact:"secret"` tags. Both history exclusions and live-resource write-only metadata now use that discovery; the Actor/Key switches are removed.

New resources participate through field declarations without adding apply-specific cases. The existing history presence markers remain, without storing secret values or hashes, and API masking/replay behavior is unchanged. Added tests for an unregistered resource shape, nil fields, containers/wrappers, and exclusion of non-secret write-only provider references. Focused apply, serialization, resource, registry and CLI tests plus preflight pass.

internal/apply/reconcile.go
```diff
@@ -0,0 +1,476 @@
...
+}
+
+func (m threeWay) merge(
+	old any,
+	op bool,
+	live any,
+	lp bool,
+	desired any,
+	dp bool,
+	path []string,
```
rmorlok, line 458: Please document what these parameters are. It's not obvious from looking at it.

rmorlok: Documented in 5629788e. The method comment now explains old/op, live/lp, desired/dp, the unescaped path segments, and the returned value/presence flag. It explicitly distinguishes (nil, true) for JSON null from (nil, false) for omission and describes conflict errors. No behavior changed. `go test ./internal/apply` and `./scripts/preflight.sh` pass.

Issue #924, opened 2026-09-08
Make apply concurrency and connector generation behavior reliable


rmorlok:
Parent: #919

Depends on task 4. Ensure unchanged definitions do not cause new connector generations; inspect selected and draft generations, respect desired release state and explicit generation addressing, and never force lifecycle state. Add ordinary API conditional-write support (object revision or ETag and atomic precondition enforcement) if providing strong conflict guarantees; no server-side apply or field managers. Cover all apply write paths and history updates, both database providers if persistence changes. Re-read/replan only safe conflicts, handle create races without creating duplicates, and document any remaining race guarantees. Test repeated applies, draft reuse, published changes, out-of-band edits and conditional-write failures. Update canonical permission table if checks change.

Run focused tests and `./scripts/preflight.sh` before committing. Open a PR; do not merge.

PR #932, opened 2026-09-13
Execute ap apply batches in dependency order

### LINKED ISSUE #923: Execute apply batches in dependency order with partial failure reporting

rmorlok:
Enables `ap apply` to create, update, and report unchanged cluster resources. The complete batch is resolved and validated before writes, then executed in stable dependency order. Failed resources skip their dependents while independent resources continue; partial failures return a nonzero exit status and preserve successful writes.

- Check namespace parents and typed resource references, including missing prerequisites and dependency cycles. Existing namespace updates do not block resources that only require the namespace to exist.
- Support human, name, JSON, and YAML results with sanitized resources, plus `--request-timeout` and the existing authentication/configuration flags. Client dry-run remains offline.
- Handle cancellation and output failures without retrying mutations, and preserve Connection creation/update restrictions.
- Update CLI and package documentation with execution behavior, permissions, and partial failure semantics.

Conditional writes, conflict replanning, and connector generation lifecycle remain in #924. This change does not promise rollback or retry ambiguous mutations.

Validation:
- `go test -race ./internal/apply ./internal/schema/registry ./cmd/cli/config ./cmd/cli`
- Additional external-reference and conflicting-alias regression test passed under the race detector.
- `yarn docs:build`
- `./scripts/preflight.sh`
- `git diff --check`

Closes #923.
Part of #919.


internal/apply/README.md
```diff
@@ -1,17 +1,57 @@
...
+  This includes the location where the document was taken from, including 
+  for the fact that there can be multiple resources in a single file. The 
+  document retains much of the raw object structure so it can be validated 
+  with things like JSON path. It explicitly extracts things like the 
+  resource kind and common metadata block 
+  (name/id/namesaces/labels/annotations).
+* **History** - object used to compute/track the version state that will be 
+  stored on the `authproxy.net/last-applied-configuration`.
+* **Plan** - A computed single field mutation that is applied to the live 
+  resource.
```
rmorlok: Please correct any mistakes I've made in the above sections but avoid making unnecessary rewrites.

Please add any other key terms/concepts.

internal/apply/README.md
```diff
@@ -1,17 +1,57 @@
...
+
+# Details
```
rmorlok: For this section and below feel free to make updates as you see fit (rewrite however much you want). Please review it and make sure it is presenting information in an order where a person with no knowledge of the system is walked through details about concepts in an order that builds on their knowledge. They will have the above overview to guide them, but avoid jumping between topics and concepts and try to progressively build the reader's understanding. Where necessary, introduce a concept with a simple, brief explanation before jumping into details.

internal/apply/README.md
rmorlok: Addressed in 16a6a8c0. Preserved the overview while correcting typos and clarifying three-way comparison, managed-field drift, and secret exclusion with presence markers. Corrected Plan to mean one resource operation (create/update/unchanged), which can change multiple fields. Added Target, Reconciliation, Managed field, Batch, and Result, and clarified that the history version describes its storage format.

internal/apply/README.md
rmorlok: Addressed in 16a6a8c0. Reordered Details to follow loading/validation, connection and target resolution, plan computation, history storage, dependency ordering, and execution/results. Added a small label example for field management and introduced prerequisites, edges, and cycles before their implementation details. Consolidated repeated failure/retry explanations and checked the descriptions against the code. Preflight and git diff --check passed; this change only touches the package README.

internal/apply/batch.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+	if kind == "Namespace" {
+		if path, err := namespace.PathFromMetadata(m); err == nil {
+			keys = append(keys, "Namespace/id/"+path)
+		}
+	}
+	return keys
+}
+
+func (b *Batch) dependency(
```
rmorlok, line 235: Please add documentation of what this method does and what the individual parameters mean.

rmorlok:
Addressed the pending comment on Batch.dependency in da8ac8fc. Added documentation for its behavior, all six parameters, namespace ordering exception, verification cache, generation checks, and errors. Preflight and git diff --check passed. GitHub rejected a direct thread reply because this account already has a pending review; leaving the update here without submitting or changing that review.

internal/schema/registry/references_test.go
```diff
@@ -0,0 +1,223 @@
...
+	"testing"
+
+	"github.com/rmorlok/authproxy/internal/schema/resources/connection"
+	"github.com/rmorlok/authproxy/internal/schema/resources/meta"
+	"github.com/rmorlok/authproxy/internal/schema/resources/namespace"
+	"github.com/rmorlok/authproxy/internal/schema/resources/rate_limit"
+	"github.com/stretchr/testify/require"
+)
+
+func TestReferencesDiscoverTypedResourceFields(t *testing.T) {
```
rmorlok, line 13: Please add additional test cases for this using the same style as below to make sure we exhaustively cover the References function and make sure it covers the future scenarios (e.g. polymorphic, json skipped, etc).

rmorlok: Addressed in ea4ebb32 using the existing nested-subtest style and scoped validator override. Added cases for polymorphic InnerVal wrappers, ordinary interfaces, JSON-excluded and unexported fields, embedded structs, zero and nil values, array/slice ordering, deterministic map traversal, mixed containers, untyped JSON lookalikes, pointer/map/slice cycles, and shared values reached through independent paths. Also removed a duplicate invocation in the nil-pointer test. The full registry suite passes under the race detector; References has 100% statement coverage. Preflight and git diff --check passed.

PR #933, opened 2026-09-19
Handle connector generations and refresh apply plans before execution

### LINKED ISSUE #924: Make apply concurrency and connector generation behavior reliable

rmorlok:
Apply now reconciles logical Connector updates against the generation the server will edit, and refreshes each resource plan immediately before execution. This prevents stale primary definitions from overwriting unrelated draft edits and preserves unrelated metadata changes visible at the final read.

- Expose optional generation lifecycle semantics through `registry.ResourceType.Generations`, backed by typed policy in the resource package. Apply performs generic transport, selection orchestration, and reconciliation without Connector casts or release-field/state knowledge. Manifest generation support also comes from the capability.
- Inspect selected, draft, and newest generations across validated pagination. Reuse drafts, preserve explicit publication intent, leave unrelated drafts alone when the primary already satisfies the declaration, and avoid unnecessary generations on repeated applies.
- Keep explicit generation targets on that generation and reject updates to published generations before writes. Publishing a matching draft includes its definition when required by the logical endpoint; masked credentials are still rejected.
- Refresh all batch resource kinds, including unchanged plans and history updates, using the original overwrite policy. Pin existing targets to their resolved IDs and reject newly appeared create targets or removed history rather than silently adopting them.
- Report create conflicts, HTTP 409/412, and ambiguous mutation failures without retrying. Successful writes remain applied.

Concurrency scope: refresh is not an atomic precondition. This PR does not add ETags, object revisions, or database changes and does not claim strong conflict guarantees. A writer between the final read and PATCH can still be overwritten, even with `--overwrite=false`; connector metadata/history and generation mutations can partially succeed within one API request. These limits and the additional `connectors:list/generations` read requirement are documented. Existing server permission checks are unchanged.

Validation:
- `go test -race ./internal/apply ./internal/schema/registry ./cmd/cli/config ./cmd/cli`
- After the lifecycle refactor: `go test -race ./internal/apply ./internal/schema/registry ./internal/schema/resources/connectors ./internal/schema/resources/meta ./cmd/cli`
- Added synthetic lifecycle orchestration with alternate collection/state/field names, registry type-safety checks, and policy input-immutability and explicit-generation tests.
- Additional masked-credential publication regression passed under the race detector.
- Tests cover repeated applies, draft reuse/publication, newest-generation cloning, explicit generation restrictions, pagination failures, concurrent metadata and managed-field edits, identity replacement, create races, removed history, and non-retried mutation failures.
- `yarn docs:build`
- `./scripts/preflight.sh`
- `git diff --check`

Closes #924.
Part of #919.


internal/apply/connector.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/rmorlok/authproxy/internal/schema/resources/connectors"
+	"github.com/rmorlok/authproxy/internal/schema/resources/meta"
+	"github.com/rmorlok/authproxy/internal/util"
+)
+
+// resolveApplyTarget chooses the generation that the logical connector update
+// endpoint will edit. Resolve itself retains ordinary GET semantics for references.
+func (c *Client) resolveApplyTarget(ctx context.Context, doc Document) (*LiveResource, error) {
+	selected, err := c.Resolve(ctx, doc)
+	if err != nil || selected == nil || doc.Kind != "Connector" || doc.Metadata.Generation != 0 {
```
rmorlok: This seems like we are getting logic specific to the entity type into the client. Is there a way we could generalize this and move a flag to the registry or something? In the future there may be other resources that support versioning and something like the released state.

rmorlok: Yes, this belongs behind a registry capability. The current `Kind == "Connector"` checks and concrete Connector casts in both target selection and reconciliation would otherwise need to be repeated for every versioned resource.

I would add an optional generation/lifecycle capability to `registry.ResourceType`, rather than just a `SupportsVersioning` boolean. A flag tells the client to list generations, but not how to identify an editable draft, interpret desired versus observed release state, identify generation-producing fields, or preserve publication intent. The capability should expose those semantics through typed accessors and pure policy functions, implemented alongside the resource contract. Resources without the capability retain the ordinary apply path.

The apply package would keep HTTP reads, pagination, three-way merging, and execution. It would use the registered collection and lifecycle capability to select the appropriate generation and finalize the patch, without importing Connector types or hard-coding draft/primary field paths. The logical endpoint's release-only publication shortcut also needs to be represented in that capability rather than assumed for all versioned resources.

That preserves the registry's existing boundary: shared resource semantics live there, while transport and apply orchestration stay in the client. I would retain the Connector regression tests and add a synthetic second lifecycle implementation to verify the generic path does not depend on Connector-specific names or states.

rmorlok: Please make the update.

rmorlok: Implemented in 80af4a09. ResourceType now exposes an optional Generations lifecycle capability, adapted from a typed meta.GenerationPolicy implemented in the Connector resource package. That policy owns observed-state classification, generation-producing fields, target selection, publication intent, and explicit-generation restrictions. Apply now uses generic generation orchestration with opaque policy context, keeping HTTP, pagination, three-way merging, and final patch validation in the client. The Connector imports, casts, and release-path/state checks are removed from apply runtime code; manifest generation support also consults the registry. Existing Connector regressions pass, with new synthetic lifecycle coverage using different states/fields/collection, registry type-safety tests, and policy immutability tests. Race tests and preflight passed.

internal/schema/resources/connectors/generations_test.go
```diff
@@ -0,0 +1,127 @@
...
+	// Create a draft connector
+	current := NewConnector()
+	current.Status = &ConnectorStatus{
+		Release: ConnectorReleaseStatus{
+			State: ConnectorReleaseStateDraft,
+		},
+	}
+	current.Spec.Definition.DisplayName = "desired draft"
+
+	// Create a primary version
```
rmorlok: Please enahance/add comments like these to walk through the steps the test is validating and why what it's doing is correct/incorrect.

rmorlok: Addressed in 7866fc24. Expanded comments across all three lifecycle-policy tests to explain the setup, boolean inputs, expected payload, state classifications, input immutability and aliasing checks, invalid context, and historical-generation rejection cases. Also clarified that finalization prepares a publication patch rather than changing the observed server state. Comments and spacing only; test behavior is unchanged. Preflight and git diff --check passed.

